PR #19, opened 2014-11-21
WIP: Backwards compatibility for Cartesian macros (ngenerate & nsplat)

timholy:
Just queuing this up, but the changes to master have to happen first.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -2,6 +2,10 @@ module Compat
 
 using Base.Meta
 
+include("ngenerate.jl")
+using .CompatCartesian
+export @ngenerate, @nsplat
+
 if VERSION < v"0.4.0-dev+1419"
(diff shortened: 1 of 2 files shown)


simonster:
This doesn't seem to work for functions that use both `@ngenerate` and Cartesian macros, e.g.:

```
julia> @ngenerate N Nothing function f(X::NTuple{N,Any}...)
       @ncall N println X
       end;

julia> f((1, 2))
ERROR: UndefVarError: X_1 not defined
 in anonymous at none:2
```

simonster:
I thought maybe just calling `macroexpand` from inside the stagedfunction might work, but that gives me an undefined symbol error. It seems like the macro expansion takes place in whatever module is being interpreted and not the module that calls `macroexpand`.

simonster:
#42 seems to fix this.

timholy:
Thanks, @simonster.

PR #29, opened 2014-12-17
Add Nullable{T}

shashi:
Just copy-pasted the most current `nullable.jl` from base. I want to start using this... Death to `Union(Nothing, T)`!

cc @johnmyleswhite

### DIFF
--- a/src/nullable.jl
+++ b/src/nullable.jl
@@ -1,5 +1,5 @@
 import Base: eltype, convert, get, isequal, ==, hash, show
-export Nullable, NullException, isnull, get
+export Nullable, NullException, isnull
 
 immutable Nullable{T}
     isnull::Bool
--- a/src/nullable.jl
+++ b/src/nullable.jl
@@ -13,16 +13,18 @@ immutable NullException <: Exception
 end
 
 Nullable{T}(value::T) = Nullable{T}(value)
+Nullable() = Nullable{Union()}()
 
 eltype{T}(::Type{Nullable{T}}) = T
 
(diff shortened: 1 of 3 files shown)


shashi:
Updated to match latest `nullable.jl`. is it a go for a merge?

src/nullable.jl
```diff
@@ -0,0 +1,63 @@
+import Base: eltype, convert, get, isequal, ==, hash, show
+export Nullable, NullException, isnull, get
```
johnmyleswhite: Exporting `get` seems weird.

src/nullable.jl
```diff
@@ -0,0 +1,63 @@
...
+    end
+end
+
+get(x::Nullable) = x.isnull ? throw(NullException()) : x.value
+
+get{T}(x::Nullable{T}, y) = x.isnull ? convert(T, y) : x.value
+
+isnull(x::Nullable) = x.isnull
+
+function isequal(x::Nullable, y::Nullable)
```
johnmyleswhite: Is this the definition in Base? This seems pretty clearly wrong, since it implies `Nullable{Int}()` is equal to `Nullable{UTF8String}()`.

simonster: `isequal(Int[], ASCIIString[])`, so if a Nullable is a single item collection, this seems right. It's also the correct definition for `hash` below.

johnmyleswhite: Ok. Thanks for clarification.

src/nullable.jl
shashi: right, updated.

PR #38, opened 2014-12-30
Fixed Base for IPv4 / IPv6

sbromberger:
This fixes the issue introduced in https://github.com/JuliaLang/Compat.jl/pull/36 - creating a new PR (and closing #36) due to the inclusion of squashed tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `sizehint` is now `sizehint!` [#9278](https://github.com/JuliaLang/julia/pull/9278)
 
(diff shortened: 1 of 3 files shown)


sbromberger:
cc @IainNZ @mlubin

mlubin:
Could you add a mention of these functions in the readme?

sbromberger:
What section should it go under, or would you prefer I create a new one? It's basically promoting the IPv4/IPv6 constructors to Base in the older versions.

mlubin:
You're defining new methods, so under "Renamed functions".

sbromberger:
Thanks. Let me know whether what I wrote is sufficient.

sbromberger:
Possible to tag? Thanks.

PR #39, opened 2015-01-06
beginswith has been renamed to startswith

mweastwood:
(I think I got the Julia version right..)


timholy:
Looks good to me, including the version number. Thanks!

PR #42, opened 2015-02-09
Define X_n variables on entry to function generated by nsplat/ngenerate

simonster:
This appears to fix DataArrays. cc @timholy

### DIFF
--- a/src/ngenerate.jl
+++ b/src/ngenerate.jl
@@ -51,7 +51,10 @@ function _nsplat(prototype, body, varname, T, itersym)
     bodyquot = Expr(:quote, body)
     newbody = quote
         $itersym = length($varsym)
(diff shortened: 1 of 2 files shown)


timholy:
LGTM. As you surely know, you can also use the `@nexprs N (d->X_d = ...)` construct, but given that you have to splice in the symbol your version is probably easier. Merge at will. (And then we should tag a new release.)

simonster:
Hmm, it seems like `@nexprs` might be a better idea if `X` is a heterogeneously typed tuple, since it looks like `X_1, X_2 = X` doesn't give us proper type information.

timholy:
Nice. I hadn't remembered that distinction, but it makes sense.

simonster:
Thanks for reviewing!

Issue #49, opened 2015-03-11
Compat for deprecated lowercase conversions?


jiahao:
With JuliaLang/julia#10452 merged, we could use some `Compat` support for `float64`, `int32`, etc.

tkelman:
You mean the reverse, right? Emulating call overloading `Float64(x)`, `Int32(x)`, etc via the `@compat` macro?

jiahao:
It does depend on how you read the preposition, but yes.

PR #50, opened 2015-03-12
Implement lower case conversion methods

jakebolewski:
Closes https://github.com/JuliaLang/Compat.jl/issues/49

cc @jiahao

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -178,6 +178,32 @@ if VERSION < v"0.4.0-dev+2485"
     export startswith
 end
 
+if VERSION < v"0.4.0-dev+3732"
+    const calltypes = Dict{Symbol,Symbol}([(:Integer,:integer),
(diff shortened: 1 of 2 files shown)


timholy:
Yay, thanks so much!

jiahao:
woot

aviks:
Yay!! Just saw a ton of deprecation messages in one my project, and came looking for this. Thanks!

mbauman:
Shouldn't these be getting lowered to `convert(T, ...)` instead of their lowercase equivalents? I think that would be a bit more consistent between the two versions.

jakebolewski:
Yes, I needed something working last night but this way enabled the `map(::Int, x)` -> `int(x)` change.  Should be fixed now.

mbauman:
Awesome.  Thank you!

PR #57, opened 2015-03-23
Add compat for unsafe_convert

simonster:
This is unexported from Base, but this way packages can `import Compat.unsafe_convert`


timholy:
Thanks!

PR #63, opened 2015-04-17
WIP: atsign-compat for Tuple{a, b} syntax

tkelman:
first attempt at making `@compat` handle the new tuple syntax from https://github.com/JuliaLang/julia/pull/10380 - putting this up immediately for feedback

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -261,6 +261,11 @@ function _compat(ex::Expr)
         elseif VERSION < v"0.4.0-dev+1419" && isexpr(f, :curly) && f.args[1] == :Ptr && length(ex.args) == 2 && ex.args[2] == 0
             ex = Expr(:call, :zero, f)
         end
+    elseif ex.head == :curly
+        f = ex.args[1]
+        if VERSION < v"0.4.0-dev+4319" && f == :Tuple
+            ex = Expr(:tuple, ex.args[2:end]...)
+        end
     end
     return Expr(ex.head, map(_compat, ex.args)...)
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -209,3 +209,11 @@ Compat.unsafe_convert(::Ptr{A}, x) = x
 
 # Test Ptr{T}(0)
 @test @compat(Ptr{Int}(0)) == C_NULL
+
+# Test Tuple{} syntax
+if VERSION < v"0.4.0-dev+4319"
+    @test @compat Tuple{1} == (1,)
+    @test @compat Tuple{:a, :b} == (:a, :b)
+    @test @compat Tuple{:a, Tuple{:b}} == (:a, (:b,))
+    @test @compat Tuple{:a, Tuple{:b, :c}} == (:a, (:b, :c))
+end


timholy:
LGTM

IainNZ:
When this is merged + tagged, and BinDeps is tagged with it, then I'll run pkgeval

simonster:
Would be nice to handle varargs as well, although I guess the syntax is not yet fully decided.

rened:
yes, `@compat Tuple{Varargs{SomeType}}` does work

test/runtests.jl
```diff
@@ -209,3 +209,11 @@ Compat.unsafe_convert(::Ptr{A}, x) = x
 
 # Test Ptr{T}(0)
 @test @compat(Ptr{Int}(0)) == C_NULL
+
+# Test Tuple{} syntax
+if VERSION < v"0.4.0-dev+4319"
+    @test @compat Tuple{1} == (1,)
+    @test @compat Tuple{:a, :b} == (:a, :b)
+    @test @compat Tuple{:a, Tuple{:b}} == (:a, (:b,))
+    @test @compat Tuple{:a, Tuple{:b, :c}} == (:a, (:b, :c))
```
stevengj: Shouldn't there be a test specifically for types?  i.e.
```
@test @compat Tuple{Int, Float64} == (Int, Float64)
@test @compat Tuple{Int, Vararg{Float64}} == (Int, Float64...)
```
?  Tuples of types, not symbols, is the whole point of the new `Tuple` syntax as I understand it.

src/Compat.jl
```diff
@@ -261,6 +261,11 @@ function _compat(ex::Expr)
         elseif VERSION < v"0.4.0-dev+1419" && isexpr(f, :curly) && f.args[1] == :Ptr && length(ex.args) == 2 && ex.args[2] == 0
             ex = Expr(:call, :zero, f)
         end
+    elseif ex.head == :curly
+        f = ex.args[1]
+        if VERSION < v"0.4.0-dev+4319" && f == :Tuple
+            ex = Expr(:tuple, ex.args[2:end]...)
```
stevengj: We also need to turn `Vararg{T}` into `T...`

stevengj:
Hmm, looks like I merged this too soon.  I tried it with types in Julia 0.3 and got:
```
julia> @compat Tuple{Int, Float64}
ERROR: type: instantiate_type: expected TypeConstructor, got Type{(Any...,)}
```

stevengj:
Tagged as Compat v0.4.0

test/runtests.jl
tkelman: Yeah that would be a good idea (I should also add a line about this change to the readme), but since the two syntaxes are the same in 0.3, I figured it shouldn't be Compat's job to try to tell them apart.

tkelman:
Could you do `git push --tags` please? Nice to have the tags show up in the github ui here.

stevengj:
I hate git's defaults for so many commands.

tkelman:
Yeah, definitely. Looks like it was double-tagged? Now on 0.4 we at least have a nice and fast `Pkg.publish` which is convenient as long as ssh keys are all working properly

PR #67, opened 2015-04-18
Add a linear eachindex definition

### LINKED ISSUE #64: eachindex
Following the discussion at JuliaLang/julia#10858, I realized that `eachindex` is not implemented in Compat for julia 0.3.  Might it make sense to implement it?

mbauman:
This is my first Compat addition - so it'd be good if someone could double-check my version number and such.

Fixes #64.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -217,3 +217,19 @@ if VERSION < v"0.4.0-dev+4319"
     @test @compat Tuple{:a, Tuple{:b}} == (:a, (:b,))
     @test @compat Tuple{:a, Tuple{:b, :c}} == (:a, (:b, :c))
 end
+
+# Ensure eachindex iterates over the whole array
+let A, B, s
+    A = reshape(1:20,4,5)
+    s = 0
+    for i in eachindex(A)
+        s += A[i]
+    end
+    @test s == 210
+    B = sparse(A)
+    s = 0
+    for i in eachindex(B)
+        s += B[i]
+    end
+    @test s == 210
+end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -8,6 +8,11 @@ if VERSION >= v"0.4.0-dev+3184"
     export @ngenerate, @nsplat
 end
 
+if VERSION < v"0.4.0-dev+1624"
+    eachindex(A::AbstractArray) = 1:length(A)
+    export eachindex
+end
+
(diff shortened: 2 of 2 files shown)


test/runtests.jl
```diff
@@ -217,3 +217,17 @@ if VERSION < v"0.4.0-dev+4319"
     @test @compat Tuple{:a, Tuple{:b}} == (:a, (:b,))
     @test @compat Tuple{:a, Tuple{:b, :c}} == (:a, (:b, :c))
 end
+
+# Ensure eachindex iterates over the whole array
+A = reshape(1:20,4,5)
```
kmsquire: Conflicts with https://github.com/JuliaLang/Compat.jl/blob/d49e234ddb563b558d890a5cfbd87dc6f641c56f/test/runtests.jl#L204

test/runtests.jl
```diff
@@ -217,3 +217,19 @@ if VERSION < v"0.4.0-dev+4319"
     @test @compat Tuple{:a, Tuple{:b}} == (:a, (:b,))
     @test @compat Tuple{:a, Tuple{:b, :c}} == (:a, (:b, :c))
 end
+
+# Ensure eachindex iterates over the whole array
+let
+    AA = reshape(1:20,4,5)
```
kmsquire: Note that using let blocks in this way prevents `AA` (and other vars) from leaking out, but doesn't prevent inward leakage (e.g., try defining `type AA end` and running this block).

mbauman: Yeah, that's why it's now `AA` instead of `A`.  I just figured it'd be best practice in any case.

kmsquire: Okay.  Having been bit by predefined variables before (in `Base`'s tests), I'm wondering if it isn't better to explicitly make everything a `let` variable:

```julia
let AA, s, B
    AA = reshape(1:20, 4, 5)
    ...
```

mbauman: Ah, interesting.  That would prevent the collision with a `type AA`, wouldn't it?

mbauman: Very cool.  Thanks for the review and the tip! :+1:

timholy:
Version number looks right. Thanks!

tkelman:
probably worth a line in the readme

timholy:
Done.

mbauman:
Thanks!

PR #68, opened 2015-04-21
bin/version.sh: check code for generating version numbers into bin

StefanKarpinski:
This could probably use a better name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,13 +95,13 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## Developer tips
 
-If you're adding additional compatibility code to this package, the following shell script is useful for extracting the version number from a git commit SHA:
+If you're adding additional compatibility code to this package, the [`bin/version.sh` script is useful for extracting the version number from a git commit SHA. For example, from the git repository of `julia`, run something like this:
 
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Leaving this here for discussion of whether putting this in a file is a good idea and/or if we should make the script fancier and/or rewrite it in Julia. It seems like almost everyone who wants to add anything to this package will want this script saved somewhere.

tkelman:
You spelled the file name wrong

I'll need to dig up where I wrote out an explanation of why, but I think we should add a note here that you want to run this script on the **merge commit** where a change gets added to master. Otherwise the build-number counting can be off.

StefanKarpinski:
Jeez. Thanks for catching that, @tkelman.

StefanKarpinski:
Fixed the file naming issue.

stevengj:
I would add an `echo -n "0.4.0-dev+"` to the script so that it outputs the whole version string.

StefanKarpinski:
See also: https://github.com/JuliaLang/julia/blob/f2bd73132765cf850ffe83996e8d6c5473dfda8f/contrib/commit-name.sh

PR #75, opened 2015-04-28
added fma and muladd

stevengj:
This adds the `fma` and `muladd` functions from JuliaLang/julia#8112 and JuliaLang/julia#9840, respectively.


jakebolewski:
committed manually in 2583c3827f93

PR #91, opened 2015-05-24
add karg remove_destination to mv()

rened:
This keyword argument was originally added in master in https://github.com/JuliaLang/julia/commit/e0a318b4b9e53767e14c472eeb16a3a5c5217188#diff-561359fb30664fed00e236179e3cead5L118

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -400,4 +400,19 @@ if VERSION < v"0.4.0-dev+2861"
     export muladd
 end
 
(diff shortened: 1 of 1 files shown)


tkelman:
Please see the discussion in the PR, this is horribly unsafe if `src` and `dst` point to the same location

rened:
@tkelman thanks, very good point. I see the code on master is now different from what I copy & pasted here - I'll update this PR.

PR #95, opened 2015-06-02
Add `@compat chol(A, Val{:U})`

timholy:
Ref https://github.com/JuliaLang/julia/issues/11546


tkelman:
looks ok, but needs a test

timholy:
Put test on the wrong branch, but this will fix it once this gets merged. (Too many branches in too many packages going at once for my feeble brain to keep track of, grr....)

tkelman:
> Too many branches in too many packages going at once

I can certainly reproduce that one locally.

PR #96, opened 2015-06-03
Convert generated function -> stagedfunction for older julia 0.4

timholy:
This is mildly silly, perhaps, and I won't object if people don't think this should be merged. But I have one machine running an older version of 0.4, and adding support for `@generated function` (translating it back into `stagedfunction`) was one of the only missing items.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -293,6 +293,14 @@ function _compat(ex::Expr)
                 Expr(:tuple, args...)
             end
         end
+    elseif ex.head == :macrocall
+        f = ex.args[1]
+        if f == symbol("@generated") && VERSION < v"0.4.0-dev+4387"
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Not silly at all – I've added this kind of Compat support myself since I've been in the situation to keep things running on older versions of a current branch myself.

timholy:
I have to say, Compat has been sheer genius (and IIRC you're one of the original people behind it). Really awesome leveraging of Julia's metaprogramming abilities. I can't imagine we'd be anywhere close to where we are now without it---it's kept both 0.3 and 0.4 humming along.

IainNZ:
The degree to which packages have continued to support 0.3 is proof of this from my PkgEval perspective - has made PkgEval not even needed to the same degree IMO.

PR #97, opened 2015-06-04
Jq/finalize

quinnj:


jakebolewski:
@quinnj, I would argue that this does not belong in Compat.  If we can't preserve semantics between the versions it is not Compat'able.  Calling `finalize(x)` on object x causes a side effect on 0.4 so the program would be in two different states depending on what Julia version you are using.

PR #100, opened 2015-06-04
Add using Dates compatibility

quinnj:


tkelman:
What's the intended usage here? Putting `using Dates` inside the macro's transformation function doesn't do much good.

quinnj:
If I have a package that depends on `Dates`, on 0.3, I just need to do `using Dates` and have `Dates` in the REQUIRES file. For 0.4, I need to do `using Base.Dates`. It hurts my head to think about right now, so I'm going to try and think about the best way to do this a little later.

tkelman:
I'm not sure whether `using` statements can be metaprogrammed via `@compat using Foo`, but for consistency with the rest of the way Compat works it would probably have to translate the 0.4 syntax, `@compat using Base.Dates`, to `using Dates` on 0.3. If that can work.

jakebolewski:
You could so the symbolic argument rewrite, and splice the new expression into a top level node. 

```julia
Expr(:toplevel, $(rewritten using Expr))
```

hayd:
This would be good for Markdown also. You have to put Dates in the REQUIRE (which is kindof annoying).

PR #101, opened 2015-06-05
Revert "Jq/finalize"

quinnj:
Reverts JuliaLang/Compat.jl#97


quinnj:
You're right @jakebolewski; I was going off a comment by Tim Holy, but this is better for my local use (where the state shouldn't really matter), but this isn't a more general solution.

PR #103, opened 2015-06-14
RFC: Added compat bindings for new Timer code in v0.4

kmsquire:
* Ref: https://github.com/JuliaLang/julia/pull/11669

This is actually a backport of the relevant functionality.  The only differences are

* the class is named Timer2
* `Base._uv_hook_asynccb` is used for the callback, instead of `Base._uv_hook_timercb`, which doesn't exist prior to https://github.com/JuliaLang/julia/pull/11669.  The signature is specialized to `Timer2` types, so I don't think this will cause any problems

Also adds some basic tests (which should also be added to Base).

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@compat chol(A, Val{:U})` - Julia 0.4 type-stable cholesky factorizations (will not be type-stable on 0.3)
 
+* `@compat f(t::Timer)` - mimic the Julia 0.4 Timer class
+
 ## Type Aliases
 
 * `typealias AbstractString String` - `String` has been renamed to `AbstractString` [#8872](https://github.com/JuliaLang/julia/pull/8872)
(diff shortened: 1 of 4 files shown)


kmsquire:
Cc: @JeffBezanson 

Something like this is needed for https://github.com/JuliaLang/Tk.jl/issues/101

yuyichao:
Do you know if this will help https://github.com/stevengj/PyPlot.jl/issues/136 as well?

kmsquire:
From https://github.com/stevengj/PyPlot.jl/issues/136#issuecomment-111628269, the problem with `PyPlot` was caused by the same commit, and `PyCall` uses `start_timer`, which is the old timer API. So yes, this should fix things after `PyCall` is updated to use the new `Timer` API (and references `@compat` for older versions of Julia).

Cc: @stevengj

kmsquire:
Its interesting that, on master, it's possible that the timer still triggers after it's closed.

kmsquire:
This should be good to go.

I worked around a race condition on master, where it's possible that a timer still can triggers after it's closed (see https://travis-ci.org/JuliaLang/Compat.jl/jobs/66771495#L149-L159).  It's hard to force it to happen, so I inserted a `sleep` [here](https://github.com/JuliaLang/Compat.jl/commit/5247fce2cead073cae998a5d2788a1c36874cac8#diff-fce720c43af3c52c862fd7451c7374b8R355).  Should probably be addressed in master, but for now, this seems to give the right behavior.

@JeffBezanson, it would be great if you could take a look, since you were the one who rewrote the `Timer` code.  

I'll merge in the next day or two if there are no comments.

PR #113, opened 2015-07-08
@compat for documenting macros after definition

yuyichao:
`@compat` for the yet-to-be-merged https://github.com/JuliaLang/julia/pull/12000

This will need at least one rebase before merging since the version comparason is just a place holder for now.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -306,6 +306,16 @@ function _compat(ex::Expr)
             if isexpr(f, :function)
                 ex = Expr(:stagedfunction, f.args...)
             end
+        elseif f == symbol("@doc") && VERSION < v"0.4.0-dev+5873"
+            # Replace the quote with a direct macrocall for the special cases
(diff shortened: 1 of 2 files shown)


yuyichao:
The version number points to the merge commit of JuliaLang/julia#12000 now so this is ready to be merged if we think it's necessary. (and I'd wait for a PkgEval run to tell how much breakage there are.)

MichaelHatherly:
Hopefully it won't cause much breakage at all. Docile's not got this syntax anyway, so we shouldn't have to worry about it's impact on 0.3 packages.

yuyichao:
> Docile's not got this syntax anyway,

I didn't know about that. Should be fine then.

yuyichao:
Given @MichaelHatherly 's comment and that there isn't a complaint form @IainNZ , I guess this isn't really an issue and I'm closing this PR.

Will keep the branch and the commit SHA fc89b07f9eba3ba99ee48f7ead9e719ae29e955d for longer just in case.

PR #115, opened 2015-07-15
Added Compat for DataStructures.OrderedDict

kmsquire:
* OrderedDict construction has been changed to match Dict construction
  in Base.  This change allows `@compat OrderedDict{...}(...)` to be
  used on v0.3 (and early versions of v0.4)

`OrderedDict` is in DataStructures.jl, which is part of the JuliaLang organization, so it still seems appropriate to me to add this to Compat.jl.  But I'm willing to hear arguments otherwise.

The alternative would be to create a separate macro in DataStructures.jl (see https://github.com/JuliaLang/DataStructures.jl/pull/105) and ask users to use that macro instead of `@compat`, which is awkward, but doable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ wherever you want to use syntax that differs in the latest Julia
 
 Currently, the `@compat` macro supports the following syntaxes:
 
-* `@compat Dict(foo => bar, baz => qux)` - type-inferred `Dict` construction.
+* `@compat Dict(foo => bar, baz => qux)` - type-inferred `Dict` construction. (Also works for DataStructures.OrderedDict)
 
-* `@compat Dict{Foo,Bar}(foo => bar, baz => qux)` - type-declared `Dict` construction.
+* `@compat Dict{Foo,Bar}(foo => bar, baz => qux)` - type-declared `Dict` construction. (Also works for DataStructures.OrderedDict)
 
 * `@compat split(str, splitter; keywords...)` - the Julia 0.4-style keyword-based `split` function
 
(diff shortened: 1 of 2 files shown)


yuyichao:
Is the new constructor compatible with Julia 0.3?

kmsquire:
Except for the `Pair` syntax (`'a'=>1`), yes.

```julia
julia> versioninfo()
Julia Version 0.3.11-pre+13
Commit daa39e4 (2015-07-03 14:31 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin14.3.0)
  CPU: Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas
  LIBM: libopenlibm
  LLVM: libLLVM-3.3

julia> using DataStructures

julia> x = OrderedDict{String, Int}()
OrderedDict{String,Int64} with 0 entries

julia> using Compat

julia> y = @compat OrderedDict{String, Int}("a"=>1, "b"=>2)
OrderedDict{String,Int64} with 2 entries:
  "a" => 1
  "b" => 2

julia> z = @compat OrderedDict('c'=>3, 'd'=>4)
OrderedDict{Char,Int64} with 2 entries:
  'c' => 3
  'd' => 4
```

kmsquire:
Will merge today if there aren't any additional comments/concerns.

PR #117, opened 2015-07-21
compat for MemoryError -> OutOfMemoryError change

stevengj:
This adds compatibility for JuliaLang/julia#10503 (I just noticed this in PyCall; there is no deprecation warning because it is a type rename).


yuyichao:
.... Just noticed that you already have a PR for `MemoryError` I guess I should have checked.....

stevengj:
No problem, I totally forgot about this.

PR #120, opened 2015-07-28
Add support for set_zero_subnormals()

davidavdav:
As requested in [this thread](https://groups.google.com/d/msg/julia-dev/XiZoKWed7B8/EVOjrGulwMgJ).


stevengj:
Needs a rebase. 

Also, it seems like you should also provide get_zero_subnormals too.

yuyichao:
> Also, it seems like you should also provide `get_zero_subnormals` too.

I don't think the c runtime for 0.3 provides that.

davidavdav:
Yes, that was my assessment at the time, so I didn't include that.

stevengj:
Ah, okay.  Just a rebase, then, and this should be good to merge I think?

PR #125, opened 2015-08-04
Add require(::Symbol)

timholy:


timholy:
Or should we just force people to use
```jl
ex = Expr(:import, pkg)
@eval $ex
```
which works on both 0.3 and 0.4?

timholy:
I think I prefer the idea of transitioning away from `require`, so closing.

PR #126, opened 2015-08-05
Add doc_str and doc macros, to support Base.Docs/Docile syntax.

### LINKED ISSUE #56: make `@doc` a no-op if Docile can't be loaded?
I've had some trouble getting Docile installed on some older versions of Julia. Since Docile provides no necessary functionality for most packages that "depend" on it, it would be nice to make this an optional dependency instead: if the Docile package cannot be loaded then just ignore `@doc` macros.

hayd:
fix #56.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -45,6 +45,17 @@ if VERSION < v"0.4.0-dev+412"
     eval(Base, :(const IPAddr = IpAddr))
 end
 
+if VERSION < v"0.4.0-dev+1501"
+    try
+        import Docile: @doc_str, @doc_mstr, @doc
+    catch
+        macro doc_str(x) x end
+        macro doc_mstr(x) x end
+        macro doc(ex) esc(ex.args[end]) end
+    end
+    export @doc_str, @doc_mstr, @doc
+end
+
 if VERSION < v"0.4.0-dev+2197"
     Base.IPv4(ipstr::AbstractString) = Base.parseipv4(ipstr)
     Base.IPv6(ipstr::AbstractString) = Base.parseipv6(ipstr)
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@inline` and `@noinline` have been added. On 0.3, these are "no-ops," meaning they don't actually do anything.
 
+* `@doc`, `@doc_str` and `@doc_mstr` have been added. On 0.3 if Docile is available it will use its definitions, else these are no-ops.
+
 ## Other changes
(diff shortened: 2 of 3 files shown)


src/Compat.jl
```diff
@@ -38,6 +38,16 @@ if VERSION < v"0.4.0-dev+412"
     eval(Base, :(const IPAddr = IpAddr))
 end
 
+if VERSION < v"0.4.0-dev"
+    try
+        import Docile: @doc_str, @docmstr, @doc
```
MichaelHatherly: ``@docmstr`` -> ``@doc_mstr``

MichaelHatherly:
The version condition should probably be a bit more specific since ``@doc`` didn't get added until later during 0.4.

src/Compat.jl
hayd: fixed, thanks!

hayd:
Hmmm, it's been a long time > 6 months I think? Happy to tweak this if it's easy to work out.

MichaelHatherly:
``0.4.0-dev+1501`` I think.

hayd:
okey dokey, let's go with that. :)

stevengj:
Needs a test.

hayd:
Added test. locally this works on 0.3 both with and without Docile.

hayd:
@MichaelHatherly actually when I look at this, help doesn't seem to work for me on 0.3 (before this PR). I was under the impression that

    help> func # would not show doc
    julia> using Docile
    help> func  # would show doc

Maybe there is something funky about my 0.3 install.

Should this work?

MichaelHatherly:
> Should this work?

To get help in 0.3 for packages you need to import Lexicon first. Docile only provides the docstrings while Lexicon hooks into the base help system.

hayd:
I think it would be nice to merge this soon...

One potential issue is this *may* cause some warnings where libraries have already tried to import Docile ? The trick being, remove the using/import of Docile and let Compat do the work. ?
(Perhaps this could be solved by moving the export into the catch block?)

hayd:
I'm going to pull the trigger on this in the next few days unless there's any more comments.

MichaelHatherly:
@hayd, this is causing failures for packages on 0.3 that import Docile *after* Compat, https://github.com/JuliaGraphics/ColorTypes.jl/blob/a8dd9a3834b76e48db138141ab586325cd4efaa6/src/ColorTypes.jl#L5-L8.

This probably needs to be reverted unless there's a way to avoid the imports clashing.

First see here https://groups.google.com/forum/#!topic/julia-users/rsM4hxdkAxg.

PR #130, opened 2015-08-10
compat for JuliaLang/julia#12472

### LINKED ISSUE #121: Take care of the transition to AbstractFloat
Need to provide the facility to help package developers to deal with the `FloatingPoint` to `AbstractFloat` transition (https://github.com/JuliaLang/julia/pull/12162).

stevengj:
This adds the new `isapprox` array functions and binary operators from JuliaLang/julia#12472.

It also added the new `real(T)` methods from JuliaLang/julia#12201, which were needed for `rtoldefault`.

I also needed to fix #121 by adding `AbstractFloat`

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `typealias AbstractString String` - `String` has been renamed to `AbstractString` [#8872](https://github.com/JuliaLang/julia/pull/8872)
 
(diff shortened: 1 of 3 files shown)


stevengj:
Failure on 0.3 seems related to JuliaLang/julia#12552

stevengj:
Actually, failure on 0.3 is because of 12552 combined with `norm([0,Inf]) == NaN` on 0.3 (`== Inf` on 0.4).  I think I should just remove this test case and accept that `isapprox` will behave badly for `Inf` in 0.3.

PR #131, opened 2015-08-14
Add macro for conditionally choosing code based on Julia version

simonster:
Some things, like uses of the old FFT API, are difficult/never exported/not used widely enough to define in Compat. However, when it's necessary to select different code based on the Julia version within a function, comparing `VERSION` directly is a bad idea because it incurs a runtime cost and also potentially screws with type inference. This PR defines a `@julia_geq` macro that works like:

```julia
function f()
    x = @julia_geq v"0.4.0-dev+6632" ? 1 : 2.0
    x + 1
end
```

It can also be used with blocks:

```julia
function f()
    @julia_geq v"0.4.0-dev+6632" ? begin
        ...
    end : begin
        ...
     end
end
```

I'm open to bikeshedding both the name and the syntax, but since I'm already using something like this in DSP, I figured it might be useful enough to include here.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -566,4 +566,14 @@ if VERSION < v"0.4.0-dev+6578"
     export ≈, ≉
 end
 
+macro julia_geq(ex)
+    (Base.Meta.isexpr(ex, :if) && length(ex.args) == 3) ||
+        throw(ArgumentError("invalid syntax"))
+    version = ex.args[1]
+    (Base.Meta.isexpr(version, :macrocall) && length(version.args) == 2 && version.args[1] === symbol("@v_str")) ||
+        throw(ArgumentError("invalid syntax"))
+    VERSION >= convert(VersionNumber, version.args[2]) ? esc(ex.args[2]) : esc(ex.args[3])
+end
+export @julia_geq
+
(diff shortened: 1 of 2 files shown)


simonster:
Also worth referring to JuliaLang/julia#5892 and JuliaLang/julia#7449. In particular, @one-more-minute's macro from JuliaLang/julia#5892 is another possible approach.

stevengj:
For syntax, how about `@VERSION >= v"...."`?

stevengj:
Of course, it would be better if the compiler could statically eval `VersionNumber` comparisons, but that seems like another case for #414.

simonster:
Unfortunately you can't start a macro argument with a comparison operator because they aren't unary operators and so that won't parse, e.g.:

```julia
julia> :(>= v"0.4" ? 1 : 2)
ERROR: syntax: ">=" is not a unary operator
```

TotalVerb:
This is obsoleted by https://github.com/JuliaLang/julia/pull/16219, which was just merged. Now
```julia
@static if VERSION [ ... ] end
```
can be used.

The `@static` macro can be provided (and exported) from Compat itself on 0.3 and 0.4.

PR #139, opened 2015-09-16
Add compat for Void

yuyichao:
Also fix deprecation warnings in the test.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -410,6 +410,9 @@ function _compat(ex::Expr)
             end
         elseif VERSION < v"0.4.0-dev" && f == :Union
             ex = Expr(:call,:Union,ex.args[2:end]...)
+        elseif ex == :(Ptr{Void})
+            # Do no change Ptr{Void} to Ptr{Nothing}: 0.4.0-dev+768
+            return ex
         end
     elseif ex.head == :macrocall
         f = ex.args[1]
@@ -426,6 +429,12 @@ function _compat(ex::Expr)
         end
     end
     return Expr(ex.head, map(_compat, ex.args)...)
+end
+function _compat(ex::Symbol)
+    if VERSION < v"0.4.0-dev+768" && ex == :Void
+        return :Nothing
+    end
+    return ex
 end
 _compat(ex) = ex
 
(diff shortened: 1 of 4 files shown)


stevengj:
In 0.3, `Void` is already defined,  but as an alias for `None` rather than `Nothing`.   It seems like this is something that may need to be handled via the `@compat` macro?

yuyichao:
You are right. (Was checking on juliabox after submitting the PR).

yuyichao:
done

stevengj:
Make sure the macro leaves Ptr{Void} unchanged.

yuyichao:
The timer test is not happy about `Void` being translated to `Nothing`. I've just tested on JuliaBox 0.3 and it seems that `Ptr{Union()}` and `Ptr{Nothing}` both works for ccall.

yuyichao:
> Make sure the macro leaves Ptr{Void} unchanged.

I'm wondering if there's a case that you want the type of `Ptr{Void}` and `Void` to be consistent?

yuyichao:
...... Finally passes all tests .....

yuyichao:
Any more comment on this?

test/runtests.jl
```diff
@@ -510,3 +510,9 @@ for T = (Float32, Float64)
     @test [-3u:u:3u;] == [Compat.linspace(-3u,3u,7);] == [-3:3;].*u
     @test [3u:-u:-3u;] == [Compat.linspace(3u,-3u,7);] == [3:-1:-3;].*u
 end
+
+if VERSION < v"0.4.0-dev+768"
+    @test @compat(Void) === Nothing
+else
+    @test @compat(Void) === Void
+end
```
stevengj: Maybe `@test Ptr{Void} == @compat(Ptr{Void})`

yuyichao: Sure and done.

simonster:
Needs a mention in README.md.

yuyichao:
... done...

src/Compat.jl
```diff
@@ -410,6 +410,12 @@ function _compat(ex::Expr)
             end
         elseif VERSION < v"0.4.0-dev" && f == :Union
             ex = Expr(:call,:Union,ex.args[2:end]...)
+        elseif f == :Ptr
+            # Do no touch Ptr{Void}
+            if (VERSION < v"0.4.0-dev+768" && length(ex.args) == 2 &&
```
stevengj: You don't need the version check here.  Seems cleaner to just do:
```
elseif ex == :(Ptr{Void}) # don't change Ptr{Void} to Ptr{Nothing}
     return ex
end

yuyichao: I keep the version check mainly to help keeping track of when it can be removed. I can certainly remove it if it is preferred (when I'm back to my computer)

yuyichao: Back and done...

stevengj:
LGTM.

PR #141, opened 2015-09-16
Add missing type aliases

yuyichao:
Needed for PyCall.jl


yuyichao:
Adding README entry now.....

yuyichao:
Test and README entry added.

yuyichao:
Hold on. The version check is a little bit wrong and I think it's better to add all the missing ones together

yuyichao:
Done. The only deprecated binding that is not in Compat is `UnionType` but that one is very different.

simonster:
LGTM

yuyichao:
Should we tag a new release after this? Or is there sth else that others want to be included (I guess we can always tag another one....).

stevengj:
Tags are cheap...

PR #143, opened 2015-09-20
Ignore quote Exprs in @ compat

simonster:
Skip applying the Void -> Nothing transformation for cases such as:

```julia
julia> using Compat

julia> @compat Dict(:Void => :Nothing)
Dict{Symbol,Symbol} with 1 entry:
  :Nothing => :Nothing
```

which is what broke the JLD/JLDArchives tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@compat Void` - `Nothing` on 0.3 (`Ptr{Void}` is not changed).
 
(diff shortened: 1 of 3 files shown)


yuyichao:
Can this be limitted to simple quote? IIRC at least one of the PR I made for `Union{}` is relying on this.

simonster:
Updated to only pass through quoted symbols

yuyichao:
Maybe a test to make sure `@compat(:(Union{}))` is `:(Union())` on 0.3?

yuyichao:
LGTM otherwise. THX.

PR #144, opened 2015-09-20
Revert "Add doc_str and doc macros, to support Base.Docs/Docile syntax."

MichaelHatherly:
This reverts commit 0c57f272f6e124ae6af211bc9f24b0023694f893.

See https://github.com/JuliaLang/Compat.jl/pull/126#issuecomment-141771087 and mailing list post https://groups.google.com/forum/#!topic/julia-users/rsM4hxdkAxg.

I doubt that any packages on 0.3 are relying on Compat providing `@doc` rather than directly importing it from Docile, so I don't think reverting this should be a problem.

cc @hayd


MichaelHatherly:
Thanks.

PR #145, opened 2015-09-22
brackets around union in README.md

rened:
Recommend using
```jl
@compat(Union{args...})
```
instead of
```jl
(@compat Union{args...})
```
Using the latter in this context:
```jl
typealias MeasureOrNumber (@compat Union{Measure, Number})
```
yields the following error on 0.4 because of the Tuple-looking brackets:
```jl
ERROR: LoadError: LoadError: TypeError: Tuple: in parameter, expected Type{T}, got Tuple{Union}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@compat Void` - `Nothing` on 0.3 (`Ptr{Void}` is not changed).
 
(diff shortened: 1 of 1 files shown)


yuyichao:
This is noting specific about `Union` and putting `@compat` before `typealias` works.

rened:
Putting in the `@compat` before the typealias is good point, thanks!

PR #148, opened 2015-09-30
Unbreak 0.3

andreasnoack:
by moving the import statement outside the if block


tkelman:
does it work if inside the if but not the for?

these entries also need tests and docs

PR #149, opened 2015-09-30
Add some documentation and tests to remoteX compats

andreasnoack:


tkelman:
Thanks! No test for remote_do?

andreasnoack:
`remote_do` doesn't return anything but I have added a call without a `@test` now. If it fails we will probably get some output from that.

tkelman:
And I guess it isn't exported from either Base or Compat. Qualify as `Base.remote_do` to test it?

PR #150, opened 2015-10-05
add TCPSocket -> TcpSocket alias

rened:


yuyichao:
LGTM. I think I missed this in https://github.com/JuliaLang/Compat.jl/pull/141 because it isn't using `@deprecate_binding` ....

PR #153, opened 2015-10-13
Add a default value for the input to __precompile__

tkelman:
I think this is causing https://groups.google.com/forum/#!topic/julia-users/0abTv_UK4UY


tkelman:
Assuming no objections here?

PR #154, opened 2015-10-18
Compat for SparseArrays module

yuyichao:
Ref JuliaLang/julia#13440

Looks like the document needs some update for JuliaLang/julia#13440 too.

@mbauman

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -701,4 +701,9 @@ if VERSION < v"0.5.0-dev+431"
     end
 end
 
+if VERSION < v"0.5.0-dev+763"
+    export SparseArrays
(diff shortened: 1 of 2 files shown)


mbauman:
LGTM.  Thanks.

I'm not surprised I missed some documentation.  Could you point me to the spots that still need updating?

yuyichao:
I just did a search of SparseMatrix and saw some matches. I suppose most of them needs update.

yuyichao:
Actually none of them are actually in the sparse matrix doc but should still be updated.

yuyichao:
Just realized that I should probably add doc for this ....

PR #156, opened 2015-12-07
modify path for tests that shell out on windows

tkelman:
without this, 0.5 nightlies (though new nightlies haven't been built for about a week, so not yet)
and 0.4.2 will call the windows native versions of these rather than coreutils,
giving either different line endings or more substantially different behavior


tkelman:
Oh, I thought compat implemented `withenv`, apparently it doesn't yet. It should though.

Issue #160, opened 2016-01-18
Compatibility with 0.3 for qr


martinholters:
I have recently added

```Julia
if VERSION < v"0.4.0-dev+2840"
    qr(A, ::Type{Val{true}}; thin::Bool=true) =
        Base.qr(A, pivot=true, thin=thin)
    qr(A, ::Type{Val{false}}; thin::Bool=true) =
        Base.qr(A, pivot=false, thin=thin)
end
```

to a project. This looks like a reasonable addition for Compat, but I'm not sure whether this is better handled using expression rewriting with `@compat` like what is done for `chol`.

Anyway, having compatibility for `qr` in Compat would be nice.

andreasnoack:
I think this is fine. It would be great if you could open a PR.

yuyichao:
Closed by #161

PR #161, opened 2016-01-19
Compat for qr

martinholters:
This change applies the patch suggested in #160 and adds a test case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `super` is now `supertype` [#14338](https://github.com/JuliaLang/julia/pull/14338)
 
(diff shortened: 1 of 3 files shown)


andreasnoack:
Thanks. It would be great if you could add definitions for `qrfact` and `qrfact!` as well.

martinholters:
I would also add a few words to the README.md, I'm just not sure where. Is there a better place than the catch-all "Other changes" section? The fix is more similar to a renamed function, except that it's not a change of name but rather a change of type signature. Opinions?

martinholters:
Ok, I have added an entry in the "Renamed functions" section of the README. Anything else this should see before a merge?

andreasnoack:
@martinholters Sorry for the delay here.

PR #162, opened 2016-01-21
readbytes and readall, replaced by read and readstring per #14660

samoconnor:
Hi, @kmsquire, @nalimilan, re https://github.com/JuliaLang/GZip.jl/pull/45#issuecomment-173524469: I've made an attempt. I haven't looked at Compat.jl before, so I'm not sure if I'm doing it right.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -750,4 +750,26 @@ if VERSION < v"0.5.0-dev+1946"
     export supertype
 end
 
+
+if VERSION < v"0.5.0-dev+2228"
+    const readstring = readall
+    export readstring
+
+    read(s::IO, nb::Integer=typemax(Int)) = readbytes(s, nb)
+
+    write(filename::AbstractString, args...) = open(io->write(io, args...), filename, "w")
+
+    read(filename::AbstractString, args...) = open(io->read(io, args...), filename)
+    read!(filename::AbstractString, a) = open(io->read!(io, a), filename)
+    readuntil(filename::AbstractString, args...) = open(io->readuntil(io, args...), filename)
+    readline(filename::AbstractString) = open(readline, filename)
+    readlines(filename::AbstractString) = open(readlines, filename)
+
+    function write(to::IO, from::IO)
+        while !eof(from)
+            write(to, readavailable(from))
+        end
+    end
+end
+
 end # module
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -755,21 +755,30 @@ if VERSION < v"0.5.0-dev+2228"
     const readstring = readall
     export readstring
 
-    read(s::IO, nb::Integer=typemax(Int)) = readbytes(s, nb)
+    Base.read(s::IO) = readbytes(s)
+    Base.read(s::IO, nb) = readbytes(s, nb)
 
-    write(filename::AbstractString, args...) = open(io->write(io, args...), filename, "w")
+    Base.write(filename::AbstractString, args...) = open(io->write(io, args...), filename, "w")
 
-    read(filename::AbstractString, args...) = open(io->read(io, args...), filename)
-    read!(filename::AbstractString, a) = open(io->read!(io, a), filename)
-    readuntil(filename::AbstractString, args...) = open(io->readuntil(io, args...), filename)
-    readline(filename::AbstractString) = open(readline, filename)
-    readlines(filename::AbstractString) = open(readlines, filename)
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -717,4 +717,27 @@ if VERSION < v"0.5.0-dev+1946"
     export supertype
 end
 
+
+if VERSION < v"0.5.0-dev+2228"
+    const readall = readstring
+    const readbytes = read
+    export readall
+    export readbytes
```
yuyichao: Looks like it's at least reversed. Also I believe only `readstring` is a new name. Other ones shouldn't be overwirtten.

yuyichao:
And please use a more descriptive commit and PR title.

src/Compat.jl
samoconnor: thanks for the tips @yuyichao!

jrevels:
@samoconnor Could you rebase against the current master and add some tests? Otherwise, LGTM. This will fix a lot of annoying deprecation warnings for web-related stuff.

samoconnor:
@jrevels, done

tkelman:
`ERROR: `mktempdir` has no method matching mktempdir(::Function)` on 0.3 (that could be added as a new feature here in a separate PR though), and `ERROR: LoadError: MethodError: Cannot `convert` an object of type Array{Any,1} to an object of type UTF8String` on nightly.

samoconnor:
Hi @tkelman,
I don't have a nightly build to test this against, but it works with the 0.5 build I do have (Version 0.5.0-dev+2236 (2016-01-21 23:04 UTC)).
Is this possibly a type-inference regression in nightly?

This is the code that is failing:
```julia
        for text in [
            old_text,
            UTF8String(['A' + i % 52 for i in 1:(div(Base.SZ_UNBUFFERED_IO,2))]),
            UTF8String(['A' + i % 52 for i in 1:(    Base.SZ_UNBUFFERED_IO -1)]),
            UTF8String(['A' + i % 52 for i in 1:(    Base.SZ_UNBUFFERED_IO   )]),
            UTF8String(['A' + i % 52 for i in 1:(    Base.SZ_UNBUFFERED_IO +1)])
        ]
```

It looks to me that the compiler ought to be able to tell that those are `Char` arrays:
`Char` + `Int` % `Int` = `Char`

tkelman:
A type inference issue seems likely. Adding the `Char[...]` element type might work around the issue for now, but this could be a bug worth opening on base as well.

samoconnor:
See https://github.com/JuliaLang/julia/commit/639621859863609c5f3abbc2ed75c675695b3693#diff-435745fdfa02a0dc4f3b747ba977af11L180

test/runtests.jl
```diff
@@ -560,3 +560,199 @@ Base.remote_do(() -> true, 1) # Doesn't return anything so cannot be `@test`ed b
 @test "1234" == @compat withenv("_TESTVAR" => 1234) do
     return ENV["_TESTVAR"]
 end
+
+# https://github.com/JuliaLang/julia/pull/14660
+mktempdir() do dir
```
stevengj: This relies on JuliaLang/julia#9017, which was introduced in Julia 0.4.

Probably we should have a Compat shim for that `mktempdir` feature on 0.3 as well.

stevengj: Oh, I see @tkelman commented on this above.

malmaud:
How's this PR doing? Can I help? Would love to see Requests.jl green again.

tkelman:
A separate addition of `mktempdir` should probably be the only thing this needs. We could do that in a separate PR, once that's good to merge we can just restart the CI builds here and they should pass.

malmaud:
OK, I'll do that today.

samoconnor:
Hi @malmaud, that would me much appreciated.
I'm afraid I'm not at all familiar with `Compat.jl`, so this PR is just my best attempt.
I do all my work against v0.4.
I occasionally build a v0.5 environment if I'm trying to submit a bug fix, but I don't have any setup for testing against v0.3.
If you find anything else is broken in this PR please feel free to hijack my branch into your own fork and fix it up before merging with master.

tkelman:
You can download and extract a tarball binary to test against 0.3. It couldn't be any easier.

malmaud:
@samoconnor OK, I'll take you up on that

tkelman:
thank @malmaud, do open a new pr so appveyor runs

PR #163, opened 2016-01-25
Fix test failing on Windows due to carriage return

jrevels:
This should fix the error that's causing AppVeyor build failure on master (there might be other Windows errors, but I don't own Windows so I can only find out through CI).


tkelman:
see #156 for a solution that will be more consistent between 0.3 and 0.4. Adding `withenv` to Compat would make my test for that actually work correctly.

jrevels:
Ah, my bad, I missed that when I skimmed the open PRs. I'll close this, then, and look into adding `withenv`.

PR #167, opened 2016-02-03
Remove duplicated definitions

MichaelHatherly:
`gen_splatinfo` and `isfuncexpr` appeared twice in `ngenerate.jl`. This removes the extra definitions from the end of the file.


tkelman:
good catch!

MichaelHatherly:
Thanks. Luckily most of these duplicate warnings that 0.5 is now printing seem to be harmless. https://github.com/JuliaWeb/MbedTLS.jl/pull/27 surfaced a not so harmless one though.

PR #172, opened 2016-02-11
Add displaysize(), fix version script

nalimilan:
Two not-so-related fixes.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -820,7 +820,11 @@ if VERSION < v"0.5.0-dev+2228"
         s = open(filename)
         EachLine(s, ()->close(s))
     end
+end
 
+if VERSION < v"0.5.0-dev+2023"
+    const displaysize = Base.tty_size
+    export displaysize
 end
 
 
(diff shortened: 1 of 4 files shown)


nalimilan:
Regarding the first commit, it would be much more useful if we defined ```displaysize(io::IO) = tty_size()`, which would allow people to use the correct API on 0.5, and fallback to the REPL-limited one on 0.4. But it might also be confusing to ignore completely `io`. Opinions?

src/Compat.jl
```diff
@@ -761,7 +761,7 @@ if VERSION < v"0.5.0-dev+1946"
     export supertype
 end
 
-if VERSION < v"0.4.0-dev+1653"
+if VERSION < v"0.5.0-dev+1653"
```
tkelman: which PR was this in base?

nalimilan: Ah, nevermind, this was apparently correct. I looked for `rm_recursive` in 0.4 to check whether it was already present, but I should have checked `mktemp` and `mktempdir` instead.

tkelman:
test, readme

tkelman:
Is it possible to test `displaysize`? Needs a rebase.

nalimilan:
It doesn't look like a proper test can be done easily (I don't see one in Base), but at least I've added a test that the function returns a tuple of two ints as expected. Anyway, the correctness of the result itself should be checked by base; the only failure which could happen in Compat is that the function wouldn't be defined correctly.

tkelman:
Makes sense.

PR #175, opened 2016-02-24
Add compatibility function Compat.LinAlg.checksquare

### LINKED ISSUE #168: chksquare is now called checksquare
https://github.com/JuliaLang/julia/pull/14601

andreasnoack:
Fixes #168

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -859,4 +859,12 @@ else  # if VERSION < v"0.5.0-dev+1182"
 
 end
 
+module LinAlg
+    if VERSION < v"0.5.0-dev+2022"
+        const checksquare = Base.LinAlg.chksquare
+    else
+        import Base.LinAlg.checksquare
+    end
+end
+
 end # module
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 *  `Base.tty_size` (which was not exported) is now `displaysize` in Julia 0.5.
 
+* `Compat.LinAlg.checksquare` [#14601](https://github.com/JuliaLang/julia/pull/14601)
+
 ## New macros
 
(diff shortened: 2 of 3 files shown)


yuyichao:
Shouldn't this be better handled by a unexported `Compat.checksquare` or `Compat.LinAlg.checksquare` function?

andreasnoack:
Maybe. What are the trade-offs?

yuyichao:
An actual name is much more reliable than rewriting the ast with a macro and should be the preferred way.

yuyichao:
The down side is of course that you can't just remove the `@compat` when dropping support for old julia version and need to replace `Compat` or `Compat.LinAlg` by `Base.LinAlg` instead.

andreasnoack:
I can see that, but it is compelling that you can just use latest syntax + `@compat`. Do you see any specific vulnerabilities with the rewrite function I've written?

yuyichao:
The fact that you need to patch `using` (and missing `import` and `importall`)? In general, I think we only use `@compat` to bridge syntax changes and while there are changes that are tricky to classify, a renaming like this is certainly not one of those.

src/Compat.jl
```diff
@@ -859,4 +859,14 @@ else  # if VERSION < v"0.5.0-dev+1182"
 
 end
 
+module LinAlg
+    if VERSION < v"0.5.0-dev+2022"
+        checksquare(x) = Base.LinAlg.chksquare(x)
```
yuyichao: These can be `const checksquare = Base.LinAlg.chksquare` and `const checksquare = Base.LinAlg.checksquare` (or `import Base.LinAlg.checksquare`).

andreasnoack:
@yuyichao Thanks for the comments. I've updated the PR.

yuyichao:
LGTM. =)

PR #179, opened 2016-03-24
RFC: APL indexing (fixes #158)

### LINKED ISSUE #158: Arraypocalypse slicing compatibility
Not a lot has happened over at https://github.com/JuliaLang/julia/issues/13157 yet, but the breakage from APL-style slicing (https://github.com/JuliaLang/julia/pull/13612) is already apparent.  I hope there's something that can be done in Compat to ease the slicing transition.  Here's one possibility:
1. For 0.4, `@compat A[:,1]` rewrites the `:ref` AST node to include a placeholder type which can be dispatched on - something like `A[SliceAPL, :, 1]`.
2. We then implement special versions of `getindex` and `setindex!` taking a `SliceAPL` as the first index, which inspect the types of the incoming slice expression and squeeze out any singleton dimensions as required.

martinholters:
For now, a proof-of-concept for one of the solutions discussed in #158. I'm particularly unhappy about the way the dimensions to squeeze are determined; suggestions for improvements are welcome!

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -521,6 +521,10 @@ function _compat(ex::Expr)
             callee.args[1] = :(Base.call)
             insert!(callexpr.args, 2, obj)
         end
+    elseif ex.head === :ref && (length(ex.args) > 2 || (length(ex.args) == 2 && isa(ex.args[2], Expr) && ex.args[2].head === :...))
+        if VERSION < v"0.5.0-dev+1195"
+            ex.args[1] = :(Compat.ArrayAPL($(ex.args[1])))
+        end
     end
     return Expr(ex.head, map(_compat, ex.args)...)
 end
@@ -777,6 +781,45 @@ end
 
 if !isdefined(symbol("@generated"))
     macro generated(x)
+    end
+end
+
+if VERSION < v"0.5.0-dev+1195"
+    import Base: getindex, setindex!, ndims, size
+    immutable ArrayAPL{T}
+        A::T
+    end
+
+    setindex!(A::ArrayAPL, args...) = setindex!(A.A, args...)
+    ndims(A::ArrayAPL) = ndims(A.A)
+    size(A::ArrayAPL, d::Int) = size(A.A, d)
+    getindex(A::ArrayAPL, idx...) = A.A[idx...]
+
+    if VERSION < v"0.4.0"
+        _int_dims(offset::Int) = ()
+        _int_dims(offset::Int, ::Integer, is...) = tuple(offset, _int_dims(offset+1, is...)...)
+        _int_dims(offset::Int, ::Any, is...) = _int_dims(offset+1, is...)
+        _need_squeeze() = Val{false}
+        _need_squeeze(::Any, is...) = _need_squeeze(is...)
+        _need_squeeze(::Integer, is...) = _any_non_int(is...)
+        _any_non_int() = Val{false}
+        _any_non_int(::Any, is...) = Val{true}
+        _any_non_int(::Integer, is...) = _any_non_int(is...)
+        _maybe_squeeze(::Type{Val{false}}, A, idx) = A
+        _maybe_squeeze(::Type{Val{true}}, A, idx) = squeeze(A, _int_dims(1, idx...))
+        getindex{T<:AbstractArray}(A::ArrayAPL{T}, idx::Integer...) = A.A[idx...]
+        getindex{T<:AbstractArray}(A::ArrayAPL{T}, idx...) = _maybe_squeeze(_need_squeeze(idx...), A.A[idx...], idx)
+    else
+        @generated function getindex{T<:AbstractArray}(A::ArrayAPL{T}, idx...)
+            last_nonint_dim = findlast(t -> !(t<:Integer), idx)
+            dims = tuple(find(t -> t<:Integer, collect(idx[1:last_nonint_dim]))...)
+            if isempty(dims)
+                # no need to squeeze
+                return :(A.A[idx...])
+            else
+                return :(squeeze(A.A[idx...], $dims))
+            end
+        end
     end
 end
 
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -775,6 +775,11 @@ else
     const linspace = Base.linspace
 end
 
+if !isdefined(symbol("@generated"))
+    macro generated(x)
+    end
+end
+
 if VERSION < v"0.4.0-dev+436"
     export UDPSocket
     const UDPSocket = UdpSocket
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -858,6 +858,20 @@ let
     @test a == [(2,10),(4,20),(6,30)]
 end
 
+# APL-style indexing
+let X = [100x+10y+z for x in 1:3, y in 1:3, z in 1:3]
+    y = @compat X[1, 1, 1]
+    @test y == 111
+    y = @compat X[:, :, :]
+    @test y == X
+    y = @compat X[1, 1, :]
+    @test size(y) == (3,)
+    @test y == [111, 112, 113]
+    y = @compat X[:, 1, :]
+    @test size(y) == (3, 3)
+    @test y == [111 112 113; 211 212 213; 311 312 313]
+end
+
 module CallTest
 
 using Base.Test, Compat


c42f:
Cool, thanks for giving it a go! Note that `ArrayAPL` will also need to work with `setindex!()` - may be just a delegation to the underlying `setindex!`, I'm not sure whether the APL indexing rules have any implications there.

Because the dimension squeezing is purely dependent on the type of the input, you should be able to create an `@generated` function to make the extra stuff in `getindex` free at runtime in many cases.  In particular, you'd be able to elide the call to squeeze at the syntax level if integer indexing occurs only for trailing dimensions.  As a special case, this would also cover pure scalar indexing.

martinholters:
Yes, `setindex!` was missing - added.

But if I didn't miss anything, `@generated` is not in 0.3, so I cooked up a solution with a helper that recurses over its parameter list. It's type-stable and in the simple cases I checked, Julia was able to inline and evaluate it completely, so `squeeze` was in fact called with a constant second argument. 

Now let's see how to get rid of the `squeeze` if all integer indexes are for trailing dimensions...

martinholters:
Hm, to only squeeze when necessary without resorting to `@generated`, I've only come up with the following monstrosity:
```Julia
    _need_squeeze() = Val{false}
    _need_squeeze(::ANY, is...) = _need_squeeze(is...)
    _need_squeeze(::Integer, is...) = _any_non_int(is...)
    _any_non_int() = Val{false}
    _any_non_int(::ANY, is...) = Val{true}
    _any_non_int(::Integer, is...) = _any_non_int(is...)
    _maybe_squeeze(::Type{Val{false}}, A, d) = A
    _maybe_squeeze(::Type{Val{true}}, A, d) = squeeze(A, d)
    getindex{T<:AbstractArray}(A::ArrayAPL{T}, idx...) = _maybe_squeeze(_need_squeeze(idx...), A.A[idx...], _int_dims(1, idx...))
```
Does anyone see a simpler way?

c42f:
Right, `@generated` is definitely not in 0.3, and if you can do without it that's usually a good idea anyway.  `Val` itself wasn't in 0.3 but it seems to be in `Compat`...

martinholters:
I did some benchmarks with the above options. With Julia 0.3, the first approach (`_maybe_squeeze` and friends) works nicely up to six dimensions, nearly no performance penalty if all integer indexes are trailing dimensions. With more than six dimensions things get slightly worse, but usually only to around 20% additional runtime. With the `_squeeze_int_dims` approach, the point where a performance penalty becomes visible is already at five dimensions, so I stayed with the first approach. (And pushed that.)

I also tried benchmarking with Julia 0.4, but the results were a bit dubious in that everything - including the plain indexing for reference - was slower than with 0.3. Furthermore, just wrapping in `ArrayAPL` and unwrapping had a major impact (much more so than with 0.3). Looks like some more research is necessary...

martinholters:
Hm, I've run the benchmarking code found at https://gist.github.com/martinholters/79e8bc8ea2faef66055775def4a4c58b. Detailed results are at https://gist.github.com/martinholters/ab9313f3fa06863510eca495e799a0bb. I've used Julia 0.3.2 as shipped with Debian and self-built Julia 0.4.5.

TL, DR: With Julia 0.3.2 everything is fine, with 0.4.5 it is not.

For the former, things look rather nice as already observed previously: Up to an array dimension of 6, run-time is within a few percent with one notable exception: If all indexes are integers, the plain array indexing is so fast that creation of the `ArrayAPL` instance may add up to 45 percent. For higher array dimension, run-time increases by around 20 percent in general and 50 percent for the all-integer-index case.

Julia 0.4.5 is significantly slower for many access to the plain array already for some reason. Indexing into a pre-fabricated `ArrayAPL` adds between 10 and 50 percent, (almost) independent of the array dimension. However, creation of the `ArrayAPL` that it quadruples the runtime in the worst case.

As using `@generated` functions for 0.4 won't save on `ArrayAPL` creation, I don't think this is an attractive road to follow for now. What could be done is better analysis of the AST before wrapping the indexed object into an `ArrayAPL`: If there is only one indexed dimension, nothing needs to be done anyway; or more generally, if all but the last dimension are certainly not integer-indexed as in `A[:,:,i]`. Before implementing this, I'd appreciate some feedback whether the performance impact that still seems to be unavoidable with the `ArrayAPL` approach would be acceptable.

andreasnoack:
@mbauman Are you able to share any insights here?

mbauman:
Oh, wow, this is extremely impressive.  Sorry I don't typically follow Compat.jl, so I had missed this.  We need to get you working on arrays more, @martinholters! :)

The regressions from 0.3 -> 0.4 are surprising and frustrating.  I'll have to look at them closer, but very nice benchmarking here.  CC @jrevels.  I'm short on time, so I can't guarantee getting to this anytime soon.

An alternative here is to define a dropped scalar type.  I put together a quick sketch before realizing that we'd *also* have to change the core indexing algorithm because of this change: https://github.com/JuliaLang/julia/pull/13612/files#diff-89e97f67f8058d32eac8d917db5bbdbfR613. But maybe this will give you some inspiration? 

```jl
immutable DroppedInteger{T<:Integer} <: Integer
    x::T
end
Base.convert(::Type{DroppedInteger}, x::Integer) = DroppedInteger(x)
Base.convert{T<:Integer}(::Type{DroppedInteger{T}}, x::Integer) = DroppedInteger{T}(convert(T, x))

# And act like the integer it wraps
Base.promote_rule{T, R<:Real}(::Type{DroppedInteger{T}}, ::Type{R}) = promote_type(T, R)
# It'll need a slew of definitions to prevent ambiguities
Base.convert{N<:Number}(::Type{N}, d::DroppedInteger) = convert(N, d.x)

# Hack into Base's `index_shape` definition to only drop these specific integers
Base.index_shape_dim(A, dim, i::DroppedInteger, I::Real...) = Base.index_shape_dim(A, dim+1, I...)
Base.index_shape_dim(A, dim, i::DroppedInteger, I...) = Base.index_shape_dim(A, dim+1, I...)
# And make sure it doesn't get converted back to an Int by to_index
Base.to_index(x::DroppedInteger) = x

drop_scalar(x::Integer) = DroppedInteger(x)
drop_scalar(x) = x

drop_scalars() = ()
drop_scalars(x, xs...) = (drop_scalar(x), drop_scalars(xs...)...)

# Base.checksize is frustrating. It basically re-encodes Base.index_shape in a different place.
# I don't like it, and it gets gutted in my full APL indexing PR.  But we have to deal with it in 0.4:
# it assumes that each index corresponds to a dimension. I'll punt for now.
Base.checksize(A::AbstractArray, I...) = true
```

martinholters:
Before embarking on the endeavor of following @mbauman's suggestion, I thought I might try a `@generated` version, and it is significantly faster on 0.4. (Benchmark results to follow...)

But what's the best way to have both versions in Compat and choose based on Julia version?
```Julia
if VERSION < v"0.4.0"
    # code without @generated
else
    @generated function getindex{T<:AbstractArray}(A::ArrayAPL{T}, idx...)
        # magic happens here
    end
end
```
does not work, because Julia 0.3 will still complain about `@generated` not being defined. Is there a standard way around this?

mbauman:
You can hide it behind an `include` to a separate file — that way the 0.3 parser never sees the `@generated` macro.

My worry with an array wrapper is that it'll change the behavior of the result… even if you make `ArrayAPL <: AbstractArray`, it still won't work like the wrapped array in dispatch (e.g., AbstractArrays don't dispatch to BLAS functions).

martinholters:
OK, will try the include trick.

Concerning your worries because of the result type: Note that the result of the indexing is not an `ArrayAPL` but the underlying type. The`ArrayAPL` is only used as the first argument to `getindex` and `setindex!` (and whatever those call in turn).

mbauman:
Ah, of course, I see now.  Thanks for the correction.

c42f:
I assume the issue of the parsing of `@generated` is going to be an ongoing issue in Compat.  An alternative to hiding behind an include would be to define a trivial non-exported `@generated` inside Compat, so expressions using it at least parse correctly.

nalimilan:
Defining a no-op `@generated` if it's not yet defined indeed sounds like a good solution.

martinholters:
Ok, updated benchmark results for the `@generated` version are at https://gist.github.com/martinholters/ab9313f3fa06863510eca495e799a0bb/74b71a7d6f5cca2d5f9fd8b8dae75b625b8af49c. Interestingly, the `@generated` `getindex` often is _faster_ than plain indexing (why???). Still, the overhead of instantiating `ArrayAPL` dominates for small dimensionality. At least for 1d-indexing, that should be combated by not doing rewriting in `@compat` in the first place. But note that this is not reflected in the benchmark results as the dimensionality of `A[idx...]` cannot be differentiated on macro level.

c42f:
Great effort with the benchmarking!  It seems like something fishy is going on with those latest results.  Some possible issues which strike me:

* The for loops don't actually do anything, so it's possible that some parts will be optimized out.  I'd be tempted to do something cheap like a reduction instead, just to ensure the return value of `getindex` is being used.
* The index splatting is neat but can be very expensive in certain situations - probably not here if I had to guess, but it's worth checking.

martinholters:
@c42f I tried the following (timing code removed for clarity):
```Julia
r = 0
for i=1:N; r += length(a[idx...]); end
for i=1:N; r += length(b[idx...]); end
for i=1:N; r += length(@compat a[idx...]); end
@assert r > 0
```
Results are as fishy as before.

And there is something else, just as fishy. To avoid instantiating `ArrayAPL` without need, I rewrote `a[indexes]` to `apl(a,indexes)[indexes]` and made `apl` a `@generated` function that returned `a` or `ArrayAPL(a)`, depending on the type of `indexes`. This sounded quite smart to me. But then I did
```Julia
f(a) = a[:,1]
g(a) = apl(a, :, 1)[:, 1]
```
and timed both of them for `a=rand(3,3)`. Disappointment: `g` was about 50% slower. Trying to find out why - `apl` should be trivially inlined - I discovered the _really_ fishy bit: `@code_llvm` gives the same code for both of them!

martinholters:
Scratch the last part. I had `a` global which obviously interfered with inference/dispatch; doing `let a=rand(3,3); @time ...; end` instead, both `f` and `g` are equally fast.

martinholters:
...and sure enough, in the benchmarking I have reported, the type of the array being indexed was unstable as well. I'm re-running the whole benchmark suite now with the updated code ([see here](https://gist.github.com/martinholters/79e8bc8ea2faef66055775def4a4c58b)). From the first results, the additional cost of `ArrayAPL` instantiation is gone - but the curious speed-up remains. Oh man, is there a benchmarking expert in the house?

martinholters:
Ok, new benchmark results are now [available](https://gist.github.com/martinholters/ab9313f3fa06863510eca495e799a0bb/fb1da799211bc6ad86a96754ba489e94406082dc). So for indexing an array where the type is known at compile-time (and type-stability should be very high priority for performance-critical code anyway), only when indexing arrays with more than six dimensions in Julia 0.3, there is a non-negligible performance impact. (With Julia 0.4, there often is an inexplicable performance gain.)

However, this is only for cases where `a[idx...] == @compat a[idx...]`. Seeing how bad the `squeeze` is if it is invoked comes next...

martinholters:
Obsoleted by requiring julia 0.5.

PR #181, opened 2016-03-29
Implement at-compat for new style call overloading

yuyichao:
The first commit replaces some `==` with `===` with it is used on symbols.

The check is version based. There's probably a way to do "feature" test and I'm not sure what is the preferred way now... (only version? only feature? both?)

I'm curious why no one asked for it yet.....

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ wherever you want to use syntax that differs in the latest Julia
 
 Currently, the `@compat` macro supports the following syntaxes:
 
+* `@compat (a::B{T}){T}(c) = d` - the Julia 0.5-style call overload.
+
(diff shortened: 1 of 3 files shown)


nalimilan:
I would have asked for it at some point if you hadn't made this PR. :-)

For now, the common practice is to check for a given commit. See https://github.com/JuliaLang/Compat.jl#developer-tips In the present case, checking the feature doesn't sound easy anyway.

yuyichao:
> For now, the common practice is to check for a given commit.

I'm mainly wondering due to https://github.com/JuliaLang/julia/issues/15636 . It won't be particularly hard to do a feature detection since the function internals changed a lot and it should at most be a matter of finding the right field name to check on some type.

yuyichao:
I'll keep the version check for now. If we decide to go with the other way, the simpliest test I can think of is `!isleaftype(Function)`.

PR #183, opened 2016-04-14
add istext -> istextmime

stevengj:
Compatibility for JuliaLang/julia#15708


stevengj:
I originally had `if VERSION < v"0.5.0-dev+10878"`, which the `version.sh` script tells me is the correct version for the merge commit https://github.com/JuliaLang/julia/commit/839f8c9374286a1b44b0e1ae0225948820f83b84 on JuliaLang/julia#15708, but the nightly build on Travis and AppVeyor still gave me a failure, claiming I was exporting `istextmime` when it was already defined by Base.    So, I just switched it to `if !isdefined(Base, :istextmime)`

tkelman:
10000 sounds way too high, should be more like 3-4000. Something went wrong there.

martinholters:
I get 0.5.0-dev+3460 for that commit which sounds reasonable. But the `!defined` check works too, obviously.

tkelman:
@vtjnash mentioned somewhere having experienced "resistance" to basing things on `isdefined` instead of version numbers. How long ago was this and in what context, if you can remember? I think feature-checks are generally preferable to using version numbers.

stevengj:
I think the advantage of version checks is that it is obvious when they can be removed.  e.g. if I see a `VERSION < v"0.2-xxx"` in my code, but I require Julia 0.4, then I can obviously remove the check.  If I have `isdefined(Base, :frob)`, figuring out whether it is safe to remove may involve digging through the Julia logs.

stevengj:
I just tried on another computer, and I get 0.5.0-dev+10878 from the Compat.jl `version.sh` script.

stevengj:
Oh, I see the problem.  I need to do `git pull --tags`.   I hate git sometimes...why doesn't it pull all the tags on the current branch by default?

PR #184, opened 2016-04-18
Add @functorize

martinholters:
The new macro `@functorize(f)` turns a function `f` into a functor object if
there is one in Base corresponding to `f`. E.g. `@functorize(+)` yields
`Base.AddFun()`; `@functorize(<)` yields `<` in Julia 0.3, but `Base.LessFun()`
in 0.4.

While at present this would more naturally be achieved by providing `LessFun` etc. in Compat, this PR is in anticipation of JuliaLang/julia#15804, which will deprecate all functors.

Note that the version check in this PR can only be finalized once JuliaLang/julia#15804 has been merged.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -980,4 +980,79 @@ if !isdefined(Base, :istextmime)
     istextmime(m::@compat(Union{MIME,AbstractString})) = istext(m)
 end
 
+export @functorize
+macro functorize(f)
+    if VERSION >= v"0.5.0-dev+3701"
+        f === :scalarmax       ? :(Base.scalarmax) :
+        f === :scalarmin       ? :(Base.scalarmin) :
+        f
+    else
+        f = f === :identity        ? :(Base.IdFun()) :
+            f === :abs             ? :(Base.AbsFun()) :
+            f === :abs2            ? :(Base.Abs2Fun()) :
+            f === :exp             ? :(Base.ExpFun()) :
+            f === :log             ? :(Base.LogFun()) :
+            f === :&               ? :(Base.AndFun()) :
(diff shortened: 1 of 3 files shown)


timholy:
This looks great. Very elegant and easy to read. :+1: 

Am I right in understand that I shouldn't merge this until https://github.com/JuliaLang/julia/pull/15804 is finished?

martinholters:
@timholy Thanks. And yes, I'd like to adjust the version check before this is merged. I just wanted this to be ready and having received thumbs-up before JuliaLang/julia#15804 is merged so that affected packages don't have to wait too long for a strategy to handle the deprecations.

Oh, and I probably should add something for `CentralizedAbs2Fun`. And don't forget a short description in the README. Looks like this really still is WIP, I'm just glad the general approach didn't raise any objections.

martinholters:
Rebased, added README entry, added a few more cases (including `CentralizedAbs2Fun`).

Once JuliaLang/julia#15804 has been merged and the version check fixed, this should be done.

src/Compat.jl
```diff
@@ -980,4 +980,82 @@ if !isdefined(Base, :istextmime)
     istextmime(m::@compat(Union{MIME,AbstractString})) = istext(m)
 end
 
+export @functorize
+macro functorize(f)
+    # TODO replace with proper condition on version once Julia#15804 is merged
+    if VERSION >= v"0.5.0-" && !isdefined(Base, :BitFunctorUnary)
```
tkelman: ready?

martinholters:
Updated the version check, so as long as the CI checks don't find some last-minute-errors I might have introduced, this should be good to merge.

tkelman:
Very nice. Thanks much!

PR #185, opened 2016-04-20
Add support for RemoteChannel/RemoteRef

timholy:
See also https://github.com/JuliaLang/julia/pull/15953 for further discussion.

CC @amitmurthy. This doesn't obviate the need for trying to come up with a more complete deprecation strategy, but at least with this, adding `@compat` will allow support for both 0.4 and 0.5.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -528,6 +528,9 @@ function _compat(ex::Symbol)
     if VERSION < v"0.4.0-dev+768" && ex === :Void
         return :Nothing
     end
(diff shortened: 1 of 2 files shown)


amitmurthy:
Looks good.

hayd:
Looks like this is failing on master after https://github.com/JuliaLang/julia/pull/15976.

amitmurthy:
ref : https://github.com/JuliaLang/Compat.jl/pull/191

PR #186, opened 2016-04-23
RFC: Add compat for at-threads

tkelman:
no op macro for older versions of Julia that don't have a Threads module in Base

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1057,7 +1057,7 @@ macro functorize(f)
     end
 end
 
-if !isdefined(Base, :threads)
+if !isdefined(Base, :Threads)
     @eval module Threads
         macro threads(expr)
             return esc(expr)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1057,4 +1057,14 @@ macro functorize(f)
     end
 end
 
+if !isdefined(Base, :threads)
+    @eval module Threads
+        macro threads(expr)
+            return esc(expr)
+        end
+        export @threads
+    end
+    export Threads
+end
+
 end # module
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `cov` and `cor` don't allow keyword arguments anymore. Loading Compat defines compatibility methods for the new API. [#13465](https://github.com/JuliaLang/julia/pull/13465)
 
+* On versions of Julia that do not contain a Base.Threads module, Compat defines a Threads module containing a no-op `@threads` macro.
+
 ## New types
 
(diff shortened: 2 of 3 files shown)


src/Compat.jl
```diff
@@ -1057,4 +1057,14 @@ macro functorize(f)
...
 
+if !isdefined(Base, :threads)
+    @eval module Threads
+        macro threads(expr)
+            return esc(expr)
+        end
+        export @threads
+    end
+    export Threads
+end
```
yuyichao: See the `LinAlg` case above. I believe we usually define thing in a way that doesn't require any condition after importing/using Compat. In this case, which is identical to the `LinAlg` one, we should unconditionally define a non-exported `Threads` module with a exported `@threads`. (Or importing but not exporting `Threads` from base as `Compat.Threads` would also work.)

tkelman: I think this is different than LinAlg because Base.Threads does not exist on 0.4

tkelman: This is more like the Libc or Libdl case.

yuyichao: Well, the point is `Threads` is not exported from `Base` and neither is `LinAlg`. On 0.5, `@threads` is used as `Base.Threads.@threads` or `using Base.Threads; @threads` so the usage through `Compat` should be directly replacing `Base` with `Compat` in the above code.

tkelman: right that's how I'm using it in the tests
```
if isdefined(Base, :Threads)
    using Base.Threads
else
    using Compat.Threads
end
```
though I guess we should be able to get away without exporting Threads from Compat.

yuyichao: `Libc` and `Libdl` are exported, which is why they are exported from `Compat`. `Threads` and `LinAlg` are not exported from `Base` and they shouldn't be exported from `Compat` either. Other than this, the difference between the two impl in `Compat` is basically whether importing the module (used by `Libc` and `Libdl`) or defining the module and import the name (used by `LinAlg`) on the new version. These two are the methods I mentioned above and it doesn't really matter which one to use.

yuyichao: The test should be

```jl
using Compat.Threads
```

One of the point of `Compat` is to remove explicit version/feature check and while having a no-op `@thread` does save the user some code I believe we usually try to remove the explicit check as long as it's possible to do.

tkelman: Threads is exported from Base.

yuyichao: Sorry I thought it wasn't, (I guess I was confused with `using Threads` not working, which has nothing to do with whether `Threads` is exported....)

LGTM then.

tkelman: Yeah that is a bit confusing. And why does `Compat.Threads` work on nightly where `Base.Threads` exists, if I don't import `Base.Threads` into `Compat` first?

PR #189, opened 2016-04-30
Add advanced examples for @functorize

nalimilan:
I needed these today for NullableArrays, and they are not totally obvious.

Cc: @martinholters

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@inline` and `@noinline` have been added. On 0.3, these are "no-ops," meaning they don't actually do anything.
 
-* `@functorize` (not present in any Julia version) takes a function (or operator) and turns it into a functor object if one is available in the used Julia version. E.g. something like `mapreduce(Base.AbsFun(), Base.MulFun(), x)` can now be written as `mapreduce(@functorize(abs), @functorize(*), x)` to work accross different Julia versions.
+* `@functorize` (not present in any Julia version) takes a function (or operator) and turns it into a functor object if one is available in the used Julia version. E.g. something like `mapreduce(Base.AbsFun(), Base.MulFun(), x)` can now be written as `mapreduce(@functorize(abs), @functorize(*), x)`, and `f(::Base.AbsFun())` as `f(::typeof(@functorize(abs)))`, to work across different Julia versions. `Func{1}` can be written as `supertype(typeof(@functorize(abs)))` (and so on for `Func{2}`), which will fall back to `Function` on Julia 0.5.
 
 ## Other changes
 


martinholters:
Adding `f(::typeof(@functorize(abs)))` is without a doubt a good idea.

For the `Func{N}` case, I have to admit I hadn't even thought about it. But is it really needed? For dispatch, it might make more sense to allow `Any` - after all, in fact anything might be callable.

So while I agree that `supertype(typeof(@functorize(abs)))` is nothing that is immediately obvious, I am doubtful whether it's actually a good idea to suggest people to use it. Or am I missing an important use case here?

nalimilan:
Currently, `sum`, `prod`, etc., only accepts `Function` as first argument in their two-argument form because of the ambiguities that would arise if any type was accepted (this will probably be fixed using a `Callable` trait at some point). This is a very specific use case, but since I encountered it in NullableArrays I figured I would mention it. I don't expect it to be so appealing that people would overuse it (especially given how complex is the replacement syntax)...

README.md
```diff
@@ -160,7 +160,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@inline` and `@noinline` have been added. On 0.3, these are "no-ops," meaning they don't actually do anything.
 
-* `@functorize` (not present in any Julia version) takes a function (or operator) and turns it into a functor object if one is available in the used Julia version. E.g. something like `mapreduce(Base.AbsFun(), Base.MulFun(), x)` can now be written as `mapreduce(@functorize(abs), @functorize(*), x)` to work accross different Julia versions.
+* `@functorize` (not present in any Julia version) takes a function (or operator) and turns it into a functor object if one is available in the used Julia version. E.g. something like `mapreduce(Base.AbsFun(), Base.MulFun(), x)` can now be written as `mapreduce(@functorize(abs), @functorize(*), x)`, and `f(::Base.AbsFun())` as `f(::typeof(@functorize(abs)))`, to work accross different Julia versions. `Func{1}` can be written as `supertype(typeof(@functorize(abs)))` (and so on for `Func{2}`), which will fall back to `Function` on Julia 0.5.
```
stevengj: s/accross/across/

nalimilan: Good catch. That's one of my favorite mistakes, but it was actually present in the original. I've updated the PR.

martinholters: Yes, my bad.

martinholters:
I don't see the potential ambiguity in allowing `Any` for the function. There is `sum(func, iter)` and `sum(array, region)`. Even if the former was `sum(::Any, ::Any)`, the latter would be more specific as `array` would have to be `<:AbstractArray`, no? On the other hand, a `Callable` trait would not change the situation much, as conceivable, one could make an array callable. So yes, `sum(callable_array, array)` would be ambiguous/potentially confusing for the user, but unambiguously resolved to `sum(array, region)`.

This discussion aside, you're probably right that the chances of `supertype(typeof(@functorize(abs)))` being overused are rather slim. So :+1: from my side.

nalimilan:
@martinholters Since scalars will no longer be iterable soon, looks like your plan would make sense (before, I think it would have been confusing).

PR #190, opened 2016-05-03
add Symbol constructor

bicycle1885:
This defines `Symbol(args...)` constructor because `symbol(args...)` is deprecated in https://github.com/JuliaLang/julia/commit/5c4058e115dec44b5ed470b054fc41053c17aadf.

CC: @hayd

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `istext` is now `istextmime` [#15708](https://github.com/JuliaLang/julia/pull/15708)
 
-* `symbol` is now `Symbol` [#16154](https://github.com/JuliaLang/julia/pull/16154)
+* `symbol` is now `Symbol` [#16154](https://github.com/JuliaLang/julia/pull/16154); use `@compat Symbol(...)` if you need Julia 0.3 compatibility.
 
 ## New macros
 
--- a/README.md
+++ b/README.md
@@ -156,6 +156,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `istext` is now `istextmime` [#15708](https://github.com/JuliaLang/julia/pull/15708)
 
+* `symbol` is now `Symbol` [#16154](https://github.com/JuliaLang/julia/pull/16154)
+
 ## New macros
 
 * `@inline` and `@noinline` have been added. On 0.3, these are "no-ops," meaning they don't actually do anything.
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1086,7 +1086,7 @@ if !isdefined(Base, :Threads)
 end
 
 if v"0.4.0-dev+3732" ≤ VERSION < v"0.5.0-dev+3831"
-    Symbol(args...) = symbol(args...)
+    Base.Symbol(args...) = symbol(args...)
 end
 
 end # module
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -418,7 +418,8 @@ if VERSION < v"0.4.0-dev+3732"
                    (:Complex64,:complex64),
                    (:Complex128,:complex128),
                    (:Char,:char),
-                   (:Bool,:bool)]
+                   (:Bool,:bool),
+                   (:Symbol,:symbol)]
         calltypes[k] = v
     end
 end
@@ -457,6 +458,8 @@ function _compat(ex::Expr)
             T = ex.args[1]
             if T in (:Complex32, :Complex64, :Complex128)
                 ex = Expr(:call, calltypes[T], ex.args[2:end]...)
+            elseif T == :Symbol
+                ex = Expr(:call, calltypes[T], Expr(:call, :string, ex.args[2:end]...))
             else
                 ex = Expr(:(::), Expr(:call, :convert, T, ex.args[2:end]...), T)
             end
@@ -498,7 +501,7 @@ function _compat(ex::Expr)
         end
     elseif ex.head === :macrocall
         f = ex.args[1]
-        if f === Symbol("@generated") && VERSION < v"0.4.0-dev+4387"
+        if VERSION < v"0.4.0-dev+4387" && f === symbol("@generated")
             f = ex.args[2]
             if isexpr(f, :function)
                 ex = Expr(:stagedfunction, f.args...)
@@ -710,7 +713,7 @@ end
 
 if VERSION < v"0.4.0-dev+5688"
     typealias Irrational MathConst
-    @eval const $(Symbol("@irrational")) = getfield(Base, Symbol("@math_const"))
+    @eval const $(symbol("@irrational")) = getfield(Base, symbol("@math_const"))
     export Irrational
 else
     import Base.@irrational
@@ -1082,9 +1085,7 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
-if VERSION < v"0.4.0-dev+2678"
-    Symbol(args...) = symbol(string(args...))
-elseif VERSION < v"0.5.0-dev+3831"
+if v"0.4.0-dev+3732" ≤ VERSION < v"0.5.0-dev+3831"
     Symbol(args...) = symbol(args...)
 end
 
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1081,7 +1081,9 @@ if !isdefined(Base, :Threads)
     end
     export Threads
 
-if VERSION < v"0.5.0-dev+3831"
+if VERSION < v"0.4.0-dev+2678"
+    Symbol(args...) = symbol(string(args...))
+elseif VERSION < v"0.5.0-dev+3831"
     Symbol(args...) = symbol(args...)
 end
 
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -422,6 +422,8 @@ if VERSION < v"0.4.0-dev+3732"
                    (:Symbol,:symbol)]
         calltypes[k] = v
     end
+elseif VERSION < v"0.5.0-dev+3831"
+    Base.Symbol(args...) = symbol(args...)
 end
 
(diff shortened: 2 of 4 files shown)


src/Compat.jl
```diff
@@ -1049,12 +1049,16 @@ macro functorize(f)
...
+            f = f === Symbol(".÷") ? :(Base.DotIDivFun()) :
                 f === :.%          ? :(Base.DotRemFun()) :
                 f
         end
         f
     end
 end
 
+if VERSION < v"0.5.0-dev+3831"
+    Symbol(args...) = symbol(args...)
```
stevengj: This won't work in Julia 0.3.    I think we have to have `@compat Symbol(args...)` for Julia 0.3.

bicycle1885: Ah, I didn't know that. In that case, doesn't `Symbol(args...) = symbol(string(args...))` work?

bicycle1885:
I forked it from a bad commit, now I'm fixing it.

src/Compat.jl
yuyichao: The issue is the `Symbol(args...)` part not the `symbol(args...)` part.

test/runtests.jl
```diff
@@ -1101,3 +1101,6 @@ using Compat.Threads
 @threads for i=1:10
     @test true
 end
+
+@test Symbol("foo") === :foo
+@test Symbol("foo", "bar") === :foobar
```
hayd: do you also mind testing Symbol(1), Symbol("a_", 2), Symbol('c') ?

src/Compat.jl
stevengj: The problem is that in Julia 0.3, it is not possible to define new `Symbol(...)` methods.

stevengj: For people who want to support Julia 0.3, I think we will have to tell them to use `@compat Symbol(args...)`.   For the (many) packages that have dropped 0.3 support, we can tell them to just use `Symbol(args...)`.

bicycle1885: I understood. Now `@compat Symbol("foo", "bar")` and `Symbol("foo", "bar")` work for v0.3 and v0.4, respectively.

test/runtests.jl
bicycle1885: I added `Symbol("a_", 2)` and `Sybmol('c')` to the tests, but I have no idea on `Symbol(1)` since I couldn't write a symbol literal corresponding to it. Do you have any idea?

hayd: Symbol("1") or convert(Symbol, "1") ? Maybe it's not that important, thanks for addressing the other two!

bicycle1885:
The tests passed on v0.3 and v0.4, but not v0.5, though the error seems to be unrelated to this pull request:

```
ERROR: LoadError: MethodError: no method matching RemoteChannel{Channel{Any}}(::Int64, ::Int64, ::Int64)
Closest candidates are:
  RemoteChannel{Channel{Any}}{T<:AbstractChannel}(::Int64, ::Base.RRID)
  RemoteChannel{Channel{Any}}{T}(::Any)
 in include(::ASCIIString) at ./boot.jl:233
 in include_from_node1(::UTF8String) at ./loading.jl:426
 in process_options(::Base.JLOptions) at ./client.jl:263
 in _start() at ./client.jl:319
while loading /home/travis/.julia/v0.5/Compat/test/runtests.jl, in expression starting on line 1009
```

hayd:
Looks like the RemoteChannel API changed in https://github.com/JuliaLang/julia/pull/15976 cc @amitmurthy.

src/Compat.jl
```diff
@@ -1077,4 +1080,8 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
+if v"0.4.0-dev+3732" ≤ VERSION < v"0.5.0-dev+3831"
+    Symbol(args...) = symbol(args...)
```
stevengj: Base.Symbol(...) = ...

Or is Symbol imported somewhere?

yuyichao: `Symbol` is not a generic function.

stevengj:
Needs README docs.

src/Compat.jl
yuyichao: nvm, I thought I've seen different behavior bevore.

bicycle1885: Exactly, thank you.

bicycle1885:
Done.

src/Compat.jl
```diff
@@ -1077,4 +1080,8 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
+if v"0.4.0-dev+3732" ≤ VERSION < v"0.5.0-dev+3831"
```
tkelman: What does this lower bound correspond to? A comment with a link would be useful for reference.

src/ngenerate.jl
```diff
@@ -3,7 +3,7 @@ module CompatCartesian
 export @ngenerate, @nsplat
 
 macro ngenerate(itersym, returntypeexpr, funcexpr)
-    if isa(funcexpr, Expr) && funcexpr.head == :macrocall && funcexpr.args[1] == symbol("@inline")
+    if isa(funcexpr, Expr) && funcexpr.head == :macrocall && funcexpr.args[1] == Symbol("@inline")
```
tkelman: do these need to be `@compat(Symbol(...))` in this file ?

src/Compat.jl
bicycle1885: This is coupled with https://github.com/JuliaLang/Compat.jl/blob/937f7d0fa2ada9cb8e43d28e4d50ab398a18f9a7/src/Compat.jl#L392, where I guess constructor syntax like `Int8(1)` was introduced.

src/ngenerate.jl
yuyichao: This should be 0.4+ only

bicycle1885: Yes. Exactly speaking, this file will be included in `VERSION >= v"0.4.0-dev+3184"`, and `Base.Symbol(args...)` method is defined when `v"0.4.0-dev+3732" ≤ VERSION`. So, there is actually a term that needs `@compat` but I believe nobody uses such an old development version.

tkelman: ah right I wasn't seeing the second commit https://github.com/JuliaLang/Compat.jl/commit/a47e03742f63fc0c2b904e83f292507bf4394325 - that's a small enough window that yeah it's probably not worth worrying about

src/Compat.jl
bicycle1885: I moved the definition to make it clear.

test/runtests.jl
```diff
@@ -1101,3 +1101,15 @@ using Compat.Threads
 @threads for i=1:10
     @test true
 end
+
+if VERSION < v"0.4.0-dev+3732"
+    @test @compat(Symbol("foo")) === :foo
```
stevengj: These first four tests should be tested unconditionally.  (`@compat Symbol(...)` needs to work in 0.4 too).

test/runtests.jl
```diff
@@ -1101,3 +1101,15 @@ using Compat.Threads
...
+    @test @compat(Symbol("foo", "bar")) === :foobar
+    @test @compat(Symbol("a_", 2)) === :a_2
+    @test @compat(Symbol('c')) === :c
+else
```
stevengj: Only these last four tests should be guarded by `if VERSION ≥ v"0.4.0-dev+3732"`

test/runtests.jl
bicycle1885: Agreed. Thank you.

test/runtests.jl
bicycle1885: Done.

README.md
```diff
@@ -156,6 +156,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `istext` is now `istextmime` [#15708](https://github.com/JuliaLang/julia/pull/15708)
 
+* `symbol` is now `Symbol` [#16154](https://github.com/JuliaLang/julia/pull/16154)
```
stevengj: Should also say: "Use `@compat(Symbol(...))` if you need Julia 0.3 compatibility.

bicycle1885: Okay, done.

stevengj:
LGTM once tests are green.

hayd:
Can you give this a rebase, hopefully now (after #191) the tests will go green!

bicycle1885:
Sure & Done. Do I need to squash all commits? I think GitHub can do it instead of merging all small commits.

test/runtests.jl
nalimilan: `symbol(1)` was equivalent to `Symbol("1")` in 0.4, so we should test that `Symbol(1)` also gives the same result when using Compat.

bicycle1885: Okay, I added this test.

hayd:
Well, 0.3 is unhappy also:
```
ERROR: test error during IPv4("1.2.3.4") == @ip_str "1.2.3.4"
`IPv4` has no method matching IPv4(::ASCIIString)
```

cc @sbromberger  #38.

I don't get why this would be broken since #191, which was off the same commit. !!

sbromberger:
@hayd - IIRC (it's been a while), 0.3 used `Ipv4` (note the lower-case 'p').

hayd:
@sbromberger it worked yesterday/presumably it's been working ok since 2014... makes *no* sense.

bicycle1885:
It's really mysterious. Some recent commits passed tests (https://github.com/JuliaLang/Compat.jl/commit/c861452b57158865d9cc2d98b9c3e23034b9b869) but the master branch doesn't.

hayd:
@bicycle1885 please give this one more rebase :)

bicycle1885:
@hayd Done.

hayd:
Thanks! (Apologies all for not squashing on merge.)

PR #191, opened 2016-05-04
fix RemoteChannel tests

amitmurthy:
@timholy , please have a look.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1003,11 +1003,12 @@ rm(dirwalk, recursive=true)
 r = remotecall(sin, 1, pi/3)
 @compat foo(r::Future) = 7
 @test foo(r) == 7
-if VERSION < v"0.4.0"
-    @compat rc = RemoteChannel(1,2,3)
-else
-    @compat rc = RemoteChannel{Channel{Any}}(1,2,3)
-end
+
+@compat rc = RemoteChannel()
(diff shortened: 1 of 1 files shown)


timholy:
I'm an infrequent diver into the world of parallel, so I haven't followed exactly what these are and what these settings should be. I trust you implicitly in all such matters :smile:, and thanks for fixing!

amitmurthy:
😄 . Likewise for me and `Compat`. Therefore, just wanted a quick review before merging.

test/runtests.jl
```diff
@@ -1003,10 +1003,12 @@ rm(dirwalk, recursive=true)
 r = remotecall(sin, 1, pi/3)
 @compat foo(r::Future) = 7
 @test foo(r) == 7
-if VERSION < v"0.4.0"
-    @compat rc = RemoteChannel(1,2,3)
-else
-    @compat rc = RemoteChannel{Channel{Any}}(1,2,3)
+if VERSION < v"0.5.0-dev+1343"
```
martinholters: Why are these tests not executed for current Julia?

amitmurthy: Yes, they should be - will push an update.

amitmurthy:
Good to merge? The error seems unrelated.

hayd:
It does seem unrelated, and the tests certainly fail without this.

PR #192, opened 2016-05-04
String: alias for new String or old UTF8String

### LINKED ISSUE #193: String is back
As of JuliaLang/julia#16058, @StefanKarpinski has restored the `String` type, replacing `ASCIIString` and `UTF8String`.   This will cause a zillion deprecation warnings everywhere, so Compat will need some glue ASAP.

`String` in Julia 0.3 was what we now call `AbstractString`, but hopefully almost everyone has upgraded their old 0.3 packages.

StefanKarpinski:
Closes #193.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1077,4 +1077,12 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
+if isdefined(Core, :String) && isdefined(Core, :AbstractString)
+    typealias String Core.String
+    export String
+else
+    typealias String Base.UTF8String
+    export String
+end
+
 end # module


src/Compat.jl
```diff
@@ -1077,4 +1077,12 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
+if isdefined(Core, :String)
```
stevengj: Will this work on Julia 0.3?

stevengj: Maybe it's better to check `VERSION >= 0.5.0-dev+3876`

src/Compat.jl
```diff
@@ -1077,4 +1077,12 @@ if !isdefined(Base, :Threads)
...
+    typealias String Core.String
+    export String
```
stevengj: I'm confused, why isn't `String` already exported in Julia 0.5?

StefanKarpinski:
Updated to support 0.3 as well.

src/Compat.jl
StefanKarpinski: Unfortunately, for this to work `using Compat` needs to be followed by `import Compat: String`. Kind of a bummer, but I couldn't figure out any other way to deal with this change.

src/Compat.jl
```diff
@@ -1077,4 +1077,12 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
+if isdefined(Core, :String) && isdefined(Core, :AbstractString)
+    typealias String Core.String
+    export String
```
stevengj: I still don't understand why this is needed.  I just tried running the latest 0.5, and `String` already seems to be exported.

StefanKarpinski: I guess we don't actually need it – it seems that you can import a binding from `Compat` that was inherited from `Base`.

stevengj: We don't need to import it from Compat in the first place, because you already get the symbol from Base anyway.

StefanKarpinski: It's the wrong binding on 0.3 and 0.4 – you want to use a thing called `String` that is actually `Base.UTF8String`.

stevengj: On 0.3 and 0.4, you `export` the typealias.  That part is fine.   What I'm saying is that on 0.5, Compat can just do nothing (no exports, no typealiases).

People should be doing `using Compat` to get all of the exports (if any), not `import Compat: String` etcetera.

stevengj: (The exceptions are things like the `Libc` and `Libdl` modules, which indeed have to be explicitly imported from Compat.)

stevengj: See e.g. how [AbstractString was handled](https://github.com/JuliaLang/Compat.jl/blob/4d8b348f10ab573f3b60dd455d41bb4aa2c0a11c/src/Compat.jl#L39-L42).

StefanKarpinski: That won't work in this case since `String` is already defined and exported on both 0.3 and 0.4 – but it has a different meaning than the one we're exporting here. On 0.3, not doing the import works since at that point we just let the last export win. On 0.4, however, this will not work since we now do this:

```jl
julia> using Compat; String
WARNING: both Compat and Base export "String"; uses of it in module Main must be qualified
ERROR: UndefVarError: String not defined
 in eval(::Module, ::Any) at ./boot.jl:236
```

So the only way to write code that uses the name `String` to mean something UTF8String-like on all of 0.3, 0.4 and 0.5 is to do `using Compat; import Compat.String` and then use the name `String`. If you can think of some way around this, I'm all ears.

stevengj: Oh, I see.

tkelman: can't `@compat(String(...))` do a rewriting without messing with imports?

stevengj: Yes, but `String` types are so common that it is probably easier to do `import` than to sprinkle `@Compat` all over the place.

yuyichao:
What's the status of this?

tkelman:
still needs tests, still needs docs, and I think Stefan is traveling at the moment

timholy:
Bump. Would be nice to be able to start fixing packages.

StefanKarpinski:
All the code required to start fixing packages is merged already: all you need to do is `using Compat; import Compat.String` and then replace all uses of `ASCIIString`, `UTF8String` and `ByteString` with `String`. I'm speaking at conferences in Krakow this week and won't have any time until Thursday at the earliest if someone else wants to write some tests and docs.

andreasnoack:
I tried to go ahead and fix some packages following the advice in https://github.com/JuliaLang/Compat.jl/pull/192#issuecomment-218041673 but it doesn't work. Right now on 0.4 I get
```jl
julia> using Compat

julia> import Compat.String

julia> foo(x::String) = 1;

julia> foo("Hello")
ERROR: MethodError: `foo` has no method matching foo(::ASCIIString)

julia> methods(foo)
# 1 method for generic function "foo":
foo(x::UTF8String) at none:1
```

yuyichao:
Right. you can't just replace `UTF8String` `ASCIIString` and `ByteString` all with `UTF8String` on 0.4 without changing the package API.

andreasnoack:
But then I need to think. I tried to avoid that.

stevengj:
Maybe we need `@compat foo::String -> foo::ByteString` on 0.4?

timholy:
Probably superseded by #197.

PR #196, opened 2016-05-10
@compat for f.(args...) and foo.:bar

stevengj:
This adds compatibility for JuliaLang/julia#15032.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -542,18 +542,18 @@ function _compat(ex::Expr)
             callexpr.args[3] = obj
         end
     elseif VERSION < v"0.5.0-dev+4002" && ex.head == :. && length(ex.args) == 2 # 15032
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -541,6 +541,22 @@ function _compat(ex::Expr)
             callexpr.args[2] = params
             callexpr.args[3] = obj
         end
+    elseif VERSION < v"0.5.0-dev+4002" && ex.head == :. && length(ex.args) == 2 # 15032
+        if isexpr(ex.args[2], :quote) && typeof(ex.args[2].args[1]) == QuoteNode
```
timholy: `isa(x, QuoteNode)`

timholy:
LGTM. The `isa` is basically stylistic, so feel free to go ahead and merge if you prefer.

stevengj:
Changed to `isa`.  Merge when green.

timholy:
Thanks!

PR #197, opened 2016-05-10
Aliases for new String

timholy:
Slightly modified from #192

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -4,10 +4,12 @@ module Compat
 
 using Base.Meta
 
-if !isdefined(Core, :String) || !isdefined(Core, :AbstractString)
-    typealias String Base.UTF8String
-    export String
+if isdefined(Core, :String) && isdefined(Core, :AbstractString)
+    typealias String Core.String
+else
+    typealias String Base.ByteString
 end
+export String
 
 if VERSION < v"0.4.0-dev+2340"
     const base64encode = base64
--- a/README.md
+++ b/README.md
@@ -23,7 +23,14 @@ it is important to maintain cross-version compatibility.
 
 To use Compat in your Julia package, add a line `Compat` to the
 `REQUIRE` file in your package directory.  Then, in your package,
-after a `using Compat` statement to load Compat, simply use:
+shortly after the `module` statement include lines like these:
+
+```
+using Compat
+import Compat.String
+```
+
+and then as needed add
 
 ```
 @compat ...Julia master syntax...
@@ -63,6 +70,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `@compat withenv(f, "a" => a, "b" => b...)` on 0.3.
 
 ## Type Aliases
+
+* `String` has undergone multiple changes: in julia 0.3 it was an abstract type and then got renamed to `AbstractString`; later, `ASCIIString` and `UTF8String` got merged into a concrete type, `String`.
 
 * `typealias AbstractString String` - `String` has been renamed to `AbstractString` [#8872](https://github.com/JuliaLang/julia/pull/8872)
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,4 +1,5 @@
 using Compat
+import Compat.String
 using Base.Test
 
 v = 1
@@ -1116,3 +1117,7 @@ if VERSION ≥ v"0.4.0-dev+3732"
     @test Symbol('c') === :c
     @test Symbol(1) === Symbol("1")
 end
+
+foostring(::String) = 1
+@test foostring("hello") == 1
+@test foostring("λ") == 1


src/Compat.jl
```diff
@@ -4,10 +4,12 @@ module Compat
 
 using Base.Meta
 
-if !isdefined(Core, :String) || !isdefined(Core, :AbstractString)
-    typealias String Base.UTF8String
-    export String
+if isdefined(Core, :String) && isdefined(Core, :AbstractString)
+    typealias String Core.String
+else
+    typealias String Base.ByteString
```
yuyichao: It's unclear which one is better.... There are uses of all of the three around that can't be replaced with other ones. At least `UTF8String` is what `String` is on 0.5....

timholy: This one works in every package deprecation I've tried. `UTF8String` definitely does not.

timholy: I guess we could add `CompatString` which always means `UTF8String`, as a way to declare containers with concrete type. Thoughts?

yuyichao: I think we basically need to make all three names available on all versions...

stevengj: I honestly doubt there are any packages where using `ByteString` vs. `UTF8String` in a `type` field is performance critical.  Nor does a slight temporary slowdown in 0.4 seem like a huge deal.   If it ever is a huge deal, any package that wants to can implement the `CompatString` logic themselves.

yuyichao: There are packages that uses a tight signature like `Vector{ASCIIString}` and passing in anything different will not work. (These packages will of course needs to be updated to work on 0.5 but it will break on 0.4 if some of it's dependencies are updated this way.)

timholy: I think any time one goes from 2 types to 1, there is going to be pain.

@yuyichao, to be concrete, is your thought that we'd add
```jl
if # julia-0.5
    typealias CompatUTF8String String
    typealias CompatASCIIString String
    typealias CompatByteString String
else
    typealias CompatUTF8String UTF8String
    typealias CompatASCIIString ASCIIString
    typealias CompatByteString ByteString
end
```
? I'm fine with doing that, go for it.

stevengj: @yuyichao, they can use `{T<:Compat.String}` ... `Vector{T}`.

timholy: Do note that `convert(Vector{UTF8String}, ::Vector{ASCIIString})` Just Works. So the only problem is overly-restrictive function definitions. Over at BinDeps I had to do this in one place:
```
 -        MakeTargets(target::Vector{ASCIIString};env = Dict{AbstractString,AbstractString}()) = new("",target,env)
 +        MakeTargets{S<:AbstractString}(target::Vector{S};env = Dict{AbstractString,AbstractString}()) = new("",target,env)
```
But until some changes ASCIIString->String (i.e., when updating the package), I can't see that this will bite anyone. In other words, I think this is just part of updating packages to work with multiple versions?

yuyichao: That's the way to fix the breakage introduced by another package. Not the way to prevent breakage. e.g.

```jl
# package 1
f() = ASCIIString[]


# package 2
g(::Vector{ASCIIString}) = ...
g(Pkg1.f())
```

Package 1 cannot be fixed in the way you suggested without breaking 2 first. (It's of course not a problem if you go and find all the code that uses package 1 first but that's hard.)

timholy: You've convinced me. Can you add suitable types to Compat.jl and tag a new release? Feel free to push directly to the branch in https://github.com/JuliaLang/BinDeps.jl/pull/217, too.

tkelman:
I think the fact that Compat unconditionally exports `String` breaks any existing package that was still using the old deprecated Base.String on 0.4 or earlier.

yuyichao:
> 0.4 or earlier.

Should be 0.4 only =)

tkelman:
What should? The error is `WARNING: both Compat and Base export "String"; uses of it in module MLBase must be qualified` then an `UndefVarError`.

yuyichao:
NVM, I somehow thought it is the same on 0.3...

timholy:
> I think the fact that Compat unconditionally exports String breaks any existing package that was still using the old deprecated Base.String on 0.4 or earlier.

Oops. Although folks have had about a year to fix that, so are we really worried?

tkelman:
Yes. Having `Pkg.update()` break code is totally preventable.

timholy:
Sure, if it's easily fixable, by all means let's do that. Didn't mean to be snarky.

tkelman:
I think so, working on it in #199. If we had removed the deprecation instead of reappropriated the name then these packages would have broken on nightly, but they should continue functioning on versions of Julia that they support. Particularly Dahua's packages that haven't seen much updating in a long time.

You're like the least snarky person on github, no worries.

PR #198, opened 2016-05-11
Add unexported alias for UTF8String and ASCIIString

yuyichao:
To help package to upgrade without breaking API on julia 0.4.

As discussed in https://github.com/JuliaLang/Compat.jl/pull/197#discussion_r62842265

The name and the expected usage are picked using the follow reasoning,

1. Since this is a reverse `Compat` usage the usage should have `Compat` in it so it is easy to search and fix when 0.4 support is dropped from the package that uses this. The package can still `import` the symbol.
2. `Compat.*` is as easy to write as `Compat*`, might as well make it simpler (and reduce a few lines of code for 0.4)
3. The binding is not exported in order to not break packages on 0.5. Otherwise, a package that `using Compat` without importing explicitly will get a error when using `UTF8String` or `ASCIIString` on 0.5 instead of a deprecation.

Also fix a few warnings in the test.

@timholy @stevengj

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,12 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `String` has undergone multiple changes: in julia 0.3 it was an abstract type and then got renamed to `AbstractString`; later, `ASCIIString` and `UTF8String` got merged into a concrete type, `String`.
 
+    For packages that still need `ASCIIString` or `UTF8String` on julia 0.4 and
+    want to avoid the deprecation warning on julia 0.5,
+    use `Compat.ASCIIString` and `Compat.UTF8String` instead.
+    Note that `Compat.ASCIIString` does **not** guarantee `isascii` on julia 0.5.
+    Use `isascii` to check if the string is pure ASCII if needed.
+
 * `typealias AbstractString String` - `String` has been renamed to `AbstractString` [#8872](https://github.com/JuliaLang/julia/pull/8872)
 
 * `typealias AbstractFloat FloatingPoint` - `FloatingPoint` has been renamed to `AbstractFloat` [#12162](https://github.com/JuliaLang/julia/pull/12162)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -6,6 +6,9 @@ using Base.Meta
 
 if isdefined(Core, :String) && isdefined(Core, :AbstractString)
     typealias String Core.String
+    # Not exported in order to not break code on 0.5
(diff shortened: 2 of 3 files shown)


src/Compat.jl
```diff
@@ -6,6 +6,9 @@ using Base.Meta
 
 if isdefined(Core, :String) && isdefined(Core, :AbstractString)
     typealias String Core.String
+    # Not exported in order to not break code on 0.5
+    typealias UTF8String Core.String
+    typealias ASCIIString Core.String
```
stevengj: This doesn't seem right.   If a package declares something as `ASCIIString`, they might be expecting an `isascii` assert.

yuyichao: That's true. I think this is still useful for containers or fields though.

This basically means the package author can't blindly replace `ASCIIString` with `Compat.ASCIIString`, which is not a good idea anyway since the `isascii` info is not available in the type anymore.

timholy: Right, on julia-0.5 there is no way to do this anyway; this is just to avoid breaking packages that used to work on julia-0.4.

timholy:
LGTM. I'm fine with merging & tagging when tests pass.

src/Compat.jl
stevengj: (That is, you can't do it at the dispatch level without defining a new string type.   You can still do an `isascii` assert.)

stevengj:
The README should say explicitly that `Compat.ASCIIString` does not guarantee ASCII in 0.5, and that if you want to guarantee an ASCII string, you need to add an `isascii(s)` check.

yuyichao:
Doc updated.

README.md
```diff
@@ -77,6 +77,12 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `String` has undergone multiple changes: in julia 0.3 it was an abstract type and then got renamed to `AbstractString`; later, `ASCIIString` and `UTF8String` got merged into a concrete type, `String`.
 
+    For packages that still need `ASCIIString` or `UTF8String` on julia 0.4 and
+    want to avoid the deprecation warning on julia 0.5,
+    use `Compat.ASCIIString` and `Compat.UTF8String` instead.
+    Note that `Compat.ASCIIString` does **not** grantee `isascii` on julia 0.5.
```
tkelman: guarantee

yuyichao: .... this one was auto correction's fault......

stevengj:
LGTM.

PR #199, opened 2016-05-11
Don't export String since it is already exported by Base

tkelman:
otherwise `Pkg.test("MLBase")` for example will hit
`WARNING: both Compat and Base export "String"; uses of it in module MLBase must be qualified`
then fail with an `UndefVarError: String not defined`

ref https://github.com/JuliaLang/Compat.jl/pull/197#issuecomment-218547177

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -5,14 +5,12 @@ module Compat
 using Base.Meta
 
 if isdefined(Core, :String) && isdefined(Core, :AbstractString)
-    typealias String Core.String
     # Not exported in order to not break code on 0.5
     typealias UTF8String Core.String
     typealias ASCIIString Core.String
 else
     typealias String Base.ByteString
 end
-export String
 
 if VERSION < v"0.4.0-dev+2340"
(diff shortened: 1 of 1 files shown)


yuyichao:
You can just delete it.

src/Compat.jl
```diff
@@ -12,7 +12,9 @@ if isdefined(Core, :String) && isdefined(Core, :AbstractString)
 else
     typealias String Base.ByteString
 end
-export String
+if !isdefined(Base, :String)
```
yuyichao: This condition is basically false everywhere.

yuyichao:
And you can also delete the `typealias String Core.String`....

tkelman:
Wouldn't that be needed for `import Compat.String` to work?

yuyichao:
It shouldn't. Importing a symbol that is imported from another module should work.

```
julia> module M
       end
M

julia> import M.String

julia> String
String
```

timholy:
Great work both of you in fixing this!

PR #201, opened 2016-05-12
Compat for unsafe_write

yuyichao:
Ref https://github.com/JuliaLang/julia/pull/14766 @vtjnash 

Needed by ZipFile.jl.


yuyichao:
@vtjnash Is this (and the ZipFile PR) OK with you now?

vtjnash:
seems fine

PR #202, opened 2016-05-21
return bytes written from Base.write(::IO,::IO)

jkroso:


tkelman:
Thanks, could you add a test?

jkroso:
Sweet done

stevengj:
Failing on Julia 0.3

jkroso:
Looks like it's because of this method defined in base:

```julia
function write(to::IOBuffer, from::IOBuffer)
    write(to, pointer(from.data,from.ptr), nb_available(from))
    from.ptr += nb_available(from)
end
```

I've overwritten it for now but it does cause a warning about it to be printed. Is there a good way to get around that?

tkelman:
Rather than overwriting, maybe we have to define it as `Compat.write` ?

src/Compat.jl
```diff
@@ -956,6 +956,15 @@ if VERSION < v"0.5.0-dev+2228"
         s = open(filename)
         EachLine(s, ()->close(s))
     end
+
+    if VERSION < v"0.4"
```
tkelman: VERSION cutoffs here should be specific about when the change was made in base julia

jkroso: Super hard to hunt that commit down though

tkelman: Was it mentioned in NEWS.md? Or would git blame be usable to find when the change was made?

src/Compat.jl
```diff
@@ -945,15 +945,26 @@ if VERSION < v"0.5.0-dev+2228"
...
         end
+        return b
     end
 
     function Base.eachline(filename::AbstractString)
         s = open(filename)
         EachLine(s, ()->close(s))
     end
+
+    if VERSION < v"0.4.0-dev+5223"
```
TotalVerb: I think we can get rid of this; I doubt this is of much use to backport, and it overrides a method anyway. That's probably more trouble than it's worth.

test/runtests.jl
```diff
@@ -785,7 +785,7 @@ mktempdir() do dir
 
         verbose && println("$name write(::IOBuffer, ...)")
         @compat to = IOBuffer(UInt8[convert(UInt8, _) for _ in text], false, true)
-        write(to, io())
+        @test write(to, io()) == length(text)
```
TotalVerb: It doesn't matter so much with the current data, but this should be `length(text.data)` in case someone adds a unicode test text data.

TotalVerb:
I didn't realize this was an open PR in Julia: https://github.com/JuliaLang/julia/pull/18559. This should probably wait for that to be merged first then, and the version adjusted to the correct dev version.

PR #203, opened 2016-05-21
Add bytestring deprecation methods for 0.5

quinnj:


quinnj:
Ref: https://github.com/JuliaLang/julia/pull/16453

quinnj:
Hmmm.....how to handle the `Cstring` issue for 0.3? `Cstring` compat isn't defined until line ~698. Do we just have ot move that `bytestring` lower?

yuyichao:
Or move `Cstring` higher. You probably also need a rewrite rule in order to call the type on 0.3

quinnj:
Hmm.....now I'm running into
```julia
julia> reload("Compat")
Warning: replacing module Compat
ERROR: cannot redefine non-generic function
```
I guess we need to handle `bytestring` totally separately on 0.3

quinnj:
Hmm.......I'm struggling with how to get 0.3 compat here. We can't do `typealias String Base.ByteString` because we can't do `call` overloading in 0.3, but if we do `typealias String Union(ASCIIString,UTF8String)`, then we can't define the appropriate constructor methods.

yuyichao:
That's the rewrite rule I'm talking about. You need to use `@compat` to transform a call to `String` to appropriate (helper?) functions on 0.3.

quinnj:
Got it, let's see if I finally got this right.

quinnj:
Anything else we're missing here?

src/Compat.jl
```diff
@@ -4,12 +4,36 @@ module Compat
...
+    export Cstring, Cwstring
+end
+
 if isdefined(Core, :String) && isdefined(Core, :AbstractString)
     # Not exported in order to not break code on 0.5
     typealias UTF8String Core.String
     typealias ASCIIString Core.String
-else
+elseif v"0.4.0" <= VERSION < v"0.5.0"
     typealias String Base.ByteString
```
yuyichao: `String` should still be defined on 0.3.

src/Compat.jl
```diff
@@ -4,12 +4,36 @@ module Compat
...
+    call(::Type{Base.ByteString},s::Cstring) = bytestring(s::Cstring)
```
yuyichao: Minor issues.

The assertion on the rhs is a little weird. Simple outer constructor syntax should also work.

src/Compat.jl
yuyichao: And we can test that with `String == @compat(Union{Compat.UTF8String,Compat.ASCIIString})`.

src/Compat.jl
```diff
@@ -485,6 +499,8 @@ function _compat(ex::Expr)
             end
         elseif VERSION < v"0.4.0-dev+4389" && f === :withenv
             rewrite_pairs_to_tuples!(ex)
+        elseif VERSION < v"0.4.0" && f == :String
```
yuyichao: It's unfortunate that this is potentially breaking on 0.3 ...

yuyichao: Or does `bytestring` covers all usage of `String` on 0.3?

src/Compat.jl
quinnj: Are you saying something needs to be changed here? Or just a test added?

src/Compat.jl
quinnj: Hmmm....good question. I think in terms of function calls, it does, but I'm not 100% on that.

src/Compat.jl
yuyichao: the `typealias` shouldn't be conditional.

src/Compat.jl
yuyichao: Function call is all what matters here so it should be good then.

src/Compat.jl
quinnj: Just pushed the change.

src/Compat.jl
```diff
@@ -26,14 +26,16 @@ if isdefined(Core, :String) && isdefined(Core, :AbstractString)
     # Not exported in order to not break code on 0.5
     typealias UTF8String Core.String
     typealias ASCIIString Core.String
-elseif v"0.4.0" <= VERSION < v"0.5.0"
+elseif VERSION < v"0.5.0"
```
yuyichao: `else` is enough.

src/Compat.jl
```diff
@@ -4,12 +4,38 @@ module Compat
...
+end
+
 if isdefined(Core, :String) && isdefined(Core, :AbstractString)
     # Not exported in order to not break code on 0.5
     typealias UTF8String Core.String
     typealias ASCIIString Core.String
-else
+elseif VERSION < v"0.5.0"
     typealias String Base.ByteString
+    if VERSION >= v"0.4.0"
```
tkelman: please make all VERSION cutoffs as specific as possible

yuyichao: This is call overload ecf5426a07bbdd2ab000c9ddc3a461e1cf284027 . Version `0.4.0-dev+1246`

src/Compat.jl
tkelman: the version cutoff should be the latest version where `String` doesn't cover all signatures of `bytestring`

src/Compat.jl
```diff
@@ -4,12 +4,38 @@ module Compat
...
+    const Cwstring = Ptr{Cwchar_t}
+    export Cstring, Cwstring
+end
+
 if isdefined(Core, :String) && isdefined(Core, :AbstractString)
     # Not exported in order to not break code on 0.5
     typealias UTF8String Core.String
     typealias ASCIIString Core.String
-else
+elseif VERSION < v"0.5.0"
```
tkelman: is this needed?

src/Compat.jl
yuyichao: This is only needed before call overload. Otherwise the definition above should be enough.

src/Compat.jl
yuyichao: I don't think so .... https://github.com/JuliaLang/Compat.jl/pull/203#discussion_r64139174

yuyichao:
Ping I think this should be good to go after a rebase with the two loose version check fixed and one unnecessary version check removed?

quinnj:
Weird. I did a rebase and force push, but it's not showing up here.....

quinnj:
There should be a single commit now: b6edf4b146adb79d1a87f2532e42d0bf5c31fd10

yuyichao:
Maybe wait longer or try again? It takes some time for your comments to appear on the web interface too....

PR #205, opened 2016-05-23
Make String an alias for UTF8String, add ByteString and improve instr…

nalimilan:
`String` is really `UTF8String`, so this is the best replacement.
The only case where this is an issue is method signatures, which
should generally be ported to `AbstractString` anyway.

-------

The current `typealias String ByteString` is a trap in most cases when typing variables and fields. It's only advantage is to be a non-breaking replacement in method signatures. But packages should generally only rely on the `AbstractString` interface and not over-type their method arguments: this is the occasion to ensure maximal generality of the APIs.

Also, it's better to get an obvious immediate `NoMethod` error when passing an `ASCIIString` (the default on 0.4, so easy to spot) than more subtle type-instability issues -- which can only be fixed by using `Compat.UTF8String` and `Compat.ASCIIString` (i.e. by not using the new `String` type at all).

Cf. https://github.com/JuliaStats/DataFrames.jl/pull/970, where most uses of `String` turned out to be incorrect or sub-optimal. The current type alias is a disaster waiting to happen when more package authors will start porting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,15 +75,13 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## Type Aliases
 
-* `String` has undergone multiple changes: in julia 0.3 it was an abstract type and then got renamed to `AbstractString`; later, `ASCIIString` and `UTF8String` got merged into a concrete type, `String`.
+* `String` has undergone multiple changes: in Julia 0.3 it was an abstract type and then got renamed to `AbstractString` in 0.4; in 0.5, `ASCIIString` and `ByteString` were removed, and `UTF8String` was renamed to the (now concrete) type `String`.
 
-    For packages that still need `ASCIIString` or `UTF8String` on julia 0.4 and
-    want to avoid the deprecation warning on julia 0.5,
-    use `Compat.ASCIIString` and `Compat.UTF8String` instead.
-    Note that `Compat.ASCIIString` does **not** guarantee `isascii` on julia 0.5.
-    Use `isascii` to check if the string is pure ASCII if needed.
+    Compat provides a `String` type alias (equivalent to `UTF8String` on Julia 0.3 and 0.4), which should be used in most cases as a replacement for `ASCIIString`, `UTF8String` and `ByteString`. But note that *several particular cases should be handled with care*:
+    - Method signatures using `ASCIIString` and `ByteString` should be changed to `AbstractString` where possible. If not, use the `Compat.ASCIIString` and `Compat.ByteString` type aliases, which are equivalent to `String` on Julia 0.5, but to `ASCIIString` and `ByteString` on previous versions.
+    - Exported types with `ASCIIString` fields which are considered part of the public API should be changed to use the `Compat.ASCIIString` type to preserve compatibility on Julia 0.3 and 0.4. This type alias is equivalent to `String` on Julia 0.5, but to `ASCIIString` on previous versions. Though note that `Compat.ASCIIString` does **not** guarantee that the string only contains ASCII characters on Julia 0.5: call `isascii` to check if the string is pure ASCII if needed.
 
-* `typealias AbstractString String` - `String` has been renamed to `AbstractString` [#8872](https://github.com/JuliaLang/julia/pull/8872)
+* `typealias AbstractString String` - `String` has been renamed to `AbstractString` in Julia 0.4 [#8872](https://github.com/JuliaLang/julia/pull/8872)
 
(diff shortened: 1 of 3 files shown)


yuyichao:
Moving my comment from the out-of-date https://github.com/JuliaLang/Compat.jl/commit/8ec6bdd8b507137e00505e4455747aaf89e2159a

> I'm afraid doing this now is way too breaking at this point....

yuyichao:
Replying to https://github.com/JuliaLang/Compat.jl/commit/8ec6bdd8b507137e00505e4455747aaf89e2159a#commitcomment-17578380

> Hey, I hadn't even made the PR yet. :-)

Disclaimer: I don't have Tony's magic power. The commit just happens to be the first item on the github feed when I woke up....

> Yes, that's going to be breaking, but with some coordination it shouldn't be a major problem.

There are already many packages that depend on this behavior. A incomplete list from the package I fixed includes PyCall, Conda, BinDeps, ZMQ, FileIO, Images, ImageMagick, Clang. I'm not sure what coordination can fix all of the packages and non-registered user code.

nalimilan:
...and copying my reply from there:

> Yes, that's going to be breaking, but with some coordination it shouldn't be a major problem. OTC, I'm afraid it's going to be a nightmare if we leave all package authors get confused by the current type alias. Even somebody as familiar with Julia as @andreasnoack, with reviews from others insiders (including myself), introduced bugs into DataFrames when porting. Imagine the result with people less aware of the changes...

tkelman:
UTF8String doesn't have as many necessary constructor methods as ByteString, does it?

nalimilan:
> UTF8String doesn't have as many necessary constructor methods as ByteString, does it?

Rather the contrary AFAICT. On Julia 0.4:
```julia
julia> methods(ByteString)
2-element Array{Any,1}:
 call{T}(::Type{T}, arg) at essentials.jl:56    
 call{T}(::Type{T}, args...) at essentials.jl:57

julia> methods(UTF8String)
3-element Array{Any,1}:
 call(T::Type{UTF8String}, d::Array{UInt8,1}) at essentials.jl:37
 call{T}(::Type{T}, arg) at essentials.jl:56                     
 call{T}(::Type{T}, args...) at essentials.jl:57                 

julia> methodswith(Type{UTF8String})
9-element Array{Method,1}:
 call(T::Type{UTF8String}, d::Array{UInt8,1}) at essentials.jl:37                                    
 convert(::Type{UTF8String}, s::UTF8String) at unicode/utf8.jl:234                                   
 convert(::Type{UTF8String}, s::ASCIIString) at unicode/utf8.jl:235                                  
 convert(::Type{UTF8String}, dat::Array{UInt8,1}) at unicode/utf8.jl:241                             
 convert(::Type{UTF8String}, a::Array{UInt8,1}, invalids_as::AbstractString) at unicode/utf8.jl:293  
 convert(::Type{UTF8String}, str::UTF16String) at unicode/utf16.jl:160                               
 convert(::Type{UTF8String}, str::UTF32String) at unicode/utf32.jl:27                                
 convert(::Type{UTF8String}, s::AbstractString) at unicode/utf8.jl:313                               
 isvalid(::Type{UTF8String}, s::Union{ASCIIString,Array{UInt8,1},UTF8String}) at strings/basic.jl:229

julia> methodswith(Type{ByteString})
1-element Array{Method,1}:
 convert(::Type{ByteString}, s::AbstractString) at strings/basic.jl:29


```

There's also the `bytestring` function, but it can return either `ASCIIString` or `UTF8String`.

tkelman:
Ah, I'm thinking of the `bytestring` function methods which are currently all in the `String` constructor on 0.5, right? So we need a Compat mechanism for those.

nalimilan:
Any other opinions? Maybe @StefanKarpinski?

Disregarding the fact that several packages have been ported already, don't you agree this would have been a better strategy?

yuyichao:
Whether or not this is a better way, I don't think the fact that we've got 4 releases with this and more than a dozen packages depending on this can be disregarded.

stevengj:
@nalimilan's argument makes a certain amount of sense to me, although unfortunately it will be easier to see the effects only after a change is merged.

I think the transition should be manageable with careful coordination, but it will be painful — best to avoid tagging anything until there are PRs ready for all affected packages.

nalimilan:
Any other opinions? Maybe @StefanKarpinski?

Disregarding the fact that several packages have been ported already, don't you agree this would have been a better strategy?

PR #206, opened 2016-05-23
Add 0.5 compat for bytestring changes

quinnj:


src/Compat.jl
```diff
@@ -485,6 +501,8 @@ function _compat(ex::Expr)
             end
         elseif VERSION < v"0.4.0-dev+4389" && f === :withenv
             rewrite_pairs_to_tuples!(ex)
+        elseif VERSION < v"0.4.0" && f == :String
```
tkelman: what does this correspond to, the String => AbstractString rename?

yuyichao: Code that use `@compat` but not updated for the abstract string renaming can be broken by this.

quinnj: This is for 0.3 compatibility since `call` overload wasn't available. @yuyichao, do I need to change something here?

yuyichao: .... I forgot that you already confirmed that `bytestring` should be good enough on 0.3 so this should be fine then. https://github.com/JuliaLang/Compat.jl/pull/203#discussion_r64138813

tkelman: what is the cutoff version number though?

quinnj:
Good to merge?

src/Compat.jl
quinnj: Ah, got it. It would just be the same as we use above. I'll update.

quinnj:
Any other comments?

nalimilan:
Would be good to merge https://github.com/JuliaLang/Compat.jl/pull/208 too before tagging a new release.

src/Compat.jl
```diff
@@ -4,12 +4,38 @@ module Compat
...
     typealias ASCIIString Core.String
 else
     typealias String Base.ByteString
+    if VERSION >= v"0.4.0-dev+1246"
+        call(::Type{Base.ByteString},s::Cstring) = bytestring(s::Cstring)
+        call(::Type{Base.ByteString},v::Vector{UInt8}) = bytestring(v::Vector{UInt8})
+        call(::Type{Base.ByteString},io::Base.AbstractIOBuffer) = bytestring(io::Base.AbstractIOBuffer)
+        call(::Type{Base.ByteString},p::Union{Ptr{Int8},Ptr{UInt8}}) = bytestring(p::Union{Ptr{Int8},Ptr{UInt8}})
+        call(::Type{Base.ByteString},p::Union{Ptr{Int8},Ptr{UInt8}}, len::Integer) = bytestring(p::Union{Ptr{Int8},Ptr{UInt8}}, len::Integer)
+        call(::Type{Base.ByteString},s...) = bytestring(s::AbstractString...)
```
stevengj: This method does not seem to have been tested... see #210

PR #207, opened 2016-05-24
Add, test, and document three-argument tryparse

TotalVerb:
Used by JuliaLang/JSON.jl#140.


tkelman:
apologies for the noise, checking whether close-then-reopen causes appveyor to restart

TotalVerb:
It's interesting that Appveyor tends to randomly freeze at that particular spot. I've seen that several times. Perhaps a random problem with the network?

PR #208, opened 2016-05-24
Add Compat.repeat() accepting any AbstractArray

nalimilan:
Compatibility for JuliaLang/julia#14082. Since the new version in
Julia 0.5 calls similar() internally, it is not enough to wrap it
into a compatibility function to backport the fix. This is required
to deprecate rep() from DataArrays.jl, which provides similar
functionality for two custom array types.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1157,4 +1157,49 @@ if !isdefined(Base, @compat Symbol("@static"))
     export @static
 end
 
+# JuliaLang/julia#14082
+if VERSION < v"0.5.0-dev+4295"
+    function repeat(A::AbstractArray;
+                    inner=ntuple(x->1, ndims(A)),
+                    outer=ntuple(x->1, ndims(A)))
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -1138,4 +1138,49 @@ if !isdefined(Base, @compat Symbol("@static"))
     export @static
 end
 
+# JuliaLang/julia#14082
+if VERSION < v"0.5.0-dev+4295" && VERSION >= v"0.4"
```
tkelman: is it not possible to make this work on 0.3?

nalimilan: 0.3 doesn't have `sub2ind!`, so it forces `repeat` to be a bit less efficient. Though since you asked, I changed the code to use `sub2ind`, and now it works there too. The performance loss shouldn't be a big deal.

nalimilan:
Good to go?

tkelman:
Sure, merge when green.

PR #212, opened 2016-05-26
Fix for removal of limit_output

timholy:
This is only needed to support multiple versions of julia-0.5, since `limit_output(io)` was introduced (+1936) and then removed (+4305).

For, e.g., BenchmarkTools, it seems important to be able to support a range of julia-0.5 versions.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -357,6 +357,26 @@ function rewrite_pairs_to_tuples!(expr::Expr)
     return expr
 end
 
+# rewrites accesses to IOContext dicts
+function rewrite_iocontext!(expr::Expr)
+    if expr.head == :call && expr.args[1] == :get
+        key = expr.args[3]
+        if     (((isa(key, QuoteNode) && key.value == :limit) ||
+                 (isa(key, Expr) && key.head == :quote && key.args[1] == :limit))
+                && expr.args[4] == false)
+            if VERSION >= v"0.5.0-dev+1936" && VERSION < v"0.5.0-dev+4305"
+                expr.args[1] = :(Base.limit_output)
+                deleteat!(expr.args, 3:4)
+            elseif VERSION < v"0.5.0-dev+1936"
+                expr.head = :quote
+                expr.args[1] = false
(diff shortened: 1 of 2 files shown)


src/Compat.jl
```diff
@@ -357,6 +357,23 @@ function rewrite_pairs_to_tuples!(expr::Expr)
...
 end
 
+# rewrites accesses to IOContext dicts
+function rewrite_iocontext!(expr::Expr)
+    if expr.head == :call && expr.args[1] == :get
+        dump(expr)
+        key = expr.args[3]
+        if (((isa(key, QuoteNode) && key.value == :limit) ||
+             (isa(key, Expr) && key.head == :quote && key.args[1] == :limit))
+            && expr.args[4] == false)
```
tkelman: more readable if line continuations of block openers are indented one extra level IMO, so it's set off from the body

tkelman:
Is there anything we could try to make `get(io, :limit, false)` do something useful on 0.3 and/or 0.4?

timholy:
Maybe rewrite it as `false`?

In some places, [BenchmarkTools](https://github.com/JuliaCI/BenchmarkTools.jl/blob/60dfe83e5434c87b7311ca5d9f185f45752ed510/src/groups.jl#L237-L281) has whole separate sections for printing on 0.4. This may therefore be the best that's possible.

timholy:
This is all green. Good to go?

src/Compat.jl
tkelman: I'd have sacrificed the vertical alignment of the parens rather than adding a bunch of spaces between `if` and the start of the condition, but good enough

tkelman:
How urgently is this needed? I usually like to give Yichao, Milan et al a chance to review things here since they have a pretty good eye.

timholy:
That's fine, I was just pinging to let you know of the update.

timholy:
See also https://github.com/JuliaCI/BenchmarkTools.jl/pull/9

tkelman:
Oh right should this have a readme item?

PR #213, opened 2016-05-28
Change os dependent definitions to new syntax

andreasnoack:
and move them to the end of the source file.

I've also added a condition for the `chol` test to avoid the warning on newer versions.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -128,86 +128,6 @@ elseif VERSION < v"0.4.0-dev+1039"
     Base.rem{T<:Integer}(n::Integer, ::Type{T}) = mod(n, T)
 end
 
-import Base: srand, rand, rand!
-if VERSION < v"0.4.0-dev+1373" # PR #8854
-    function make_seed()
-        try
-            @unix_only seed = open("/dev/urandom") do io
-                a = Array(UInt32, 4)
-                read!(io, a)
-                a
-            end
-            @windows_only seed = let a = zeros(UInt32, 2)
-                Base.dSFMT.win32_SystemFunction036!(a)
-                a
-            end
-            return seed
-        catch
-            println(STDERR, "Entropy pool not available to seed RNG; using ad-hoc entropy sources.")
-            seed = reinterpret(UInt64, time())
-            seed = hash(seed, convert(UInt64, getpid()))
-            try
-                seed = hash(seed, parse(UInt64, readall(`ifconfig` |> `sha1sum`)[1:40], 16))
-            end
-            return seed
-        end
-    end
-    if VERSION < v"0.4.0-dev+1352" # PR #8832
-        function srand(r::MersenneTwister, seed::Vector{UInt32})
-            r.seed = seed
-            Base.dSFMT.dsfmt_init_by_array(r.state, seed)
-            return r
-        end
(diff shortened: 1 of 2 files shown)


src/Compat.jl
```diff
@@ -1219,4 +1139,88 @@ else
     const KERNEL = Sys.KERNEL
 end
 
+import Base: srand, rand, rand!
+if VERSION < v"0.4.0-dev+1373" # PR #8854
+    function make_seed()
+        try
+            if is_unit()
```
tkelman: typo

src/Compat.jl
```diff
@@ -132,14 +132,18 @@ import Base: srand, rand, rand!
 if VERSION < v"0.4.0-dev+1373" # PR #8854
     function make_seed()
         try
-            @unix_only seed = open("/dev/urandom") do io
-                a = Array(UInt32, 4)
-                read!(io, a)
-                a
+            @static if is_unix()
```
yuyichao: ....... note that I don't think `@static` is defined until much later ..... You can reverse the `if` and `function` if you don't want to move this.

src/Compat.jl
```diff
@@ -130,19 +130,27 @@ end
...
+                seed = let a = zeros(UInt32, 2)
+                    Base.dSFMT.win32_SystemFunction036!(a)
+                    a
+                end
             end
             return seed
-        catch
+        end
+    catch
+        function make_seed()
```
yuyichao: Errr..... the try catch should still be in the function....

andreasnoack:
This one is ready. The AV error seems unrelated.

test/runtests.jl
```diff
@@ -406,11 +406,13 @@ let s = "abcdef", u8 = "abcdef\uff", u16 = utf16(u8), u32 = utf32(u8),
     @test isvalid(UTF32String, u32)
 end
 
-# chol
-let A = rand(2,2)
-    B = A'*A
-    U = @compat chol(B, Val{:U})
-    @test_approx_eq U'*U B
+if VERSION < v"0.5.0-dev+907"
```
tkelman: did this get deprecated?

src/Compat.jl
```diff
@@ -1242,4 +1162,88 @@ else
...
+    function make_seed()
+        try
+            @static if is_unix()
+                seed = open("/dev/urandom") do io
+                    a = Array(UInt32, 4)
+                    read!(io, a)
+                    a
+                end
+            end
+            @static if is_windows()
```
tkelman: better as an `else`

test/runtests.jl
andreasnoack: Yes. The new version only accepts the matrix and not an `uplo` argument.

tkelman: is it just cholfact that lets you pick now, or only at the lapack level?

andreasnoack: It is still allowed in `cholfact`. In contrast to the Cholesky factor, the factorization as such doesn't depend on the `uplo` argument so we still allow specifying `uplo` in `cholfact`.

src/Compat.jl
andreasnoack: Done!

PR #219, opened 2016-06-02
rewrite show as writemime for Julia v0.4

dancasimiro:
writemime was deprecated in JuliaLang/julia#16563 in julia commit
ae62bf0b813afbf32402874451e55d16de909bd4. This compatibility change
allows code that is written in the Julia v0.5 style to continue working
on Julia v0.4. I have not tested this on Julia v0.3.

The compat macro must be used on the import statement and the function
definition. For example:

```
@compat import Base.show
@compat show(io::IO, ::MIME"text/plain", ::TestCustomShowType) = print(io, "MyTestCustomShowType")
```

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -446,7 +446,7 @@ function _compat(ex::Expr)
             rewrite_pairs_to_tuples!(ex)
         elseif VERSION < v"0.4.0-dev+1246" && f == :String
             ex = Expr(:call, :bytestring, ex.args[2:end]...)
-        elseif VERSION < v"0.5.0-dev+4356" && length(ex.args) > 2 && ex.args[1] === :show
+        elseif VERSION < v"0.5.0-dev+4340" && length(ex.args) > 2 && ex.args[1] === :show
             ex = rewrite_show(ex)
         end
         if VERSION < v"0.5.0-dev+4305"
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -187,6 +187,10 @@ if VERSION < v"0.5.0-dev+961"
     end
 end
 
+function rewrite_show(ex)
+    return Expr(:call, :writemime, ex.args[2], ex.args[3], ex.args[4])
+end
+
 function rewrite_dict(ex)
     length(ex.args) == 1 && return ex
 
@@ -438,6 +442,8 @@ function _compat(ex::Expr)
             rewrite_pairs_to_tuples!(ex)
         elseif VERSION < v"0.4.0-dev+1246" && f == :String
             ex = Expr(:call, :bytestring, ex.args[2:end]...)
+        elseif length(ex.args) == 4 && ex.args[1] === :show
+            ex = rewrite_show(ex)
         end
         if VERSION < v"0.5.0-dev+4305"
             rewrite_iocontext!(ex)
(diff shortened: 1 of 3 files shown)


TotalVerb:
Could you also support the syntax `@compat Base.show(io::IO, [...])`, which would be rewritten to `Base.writemime([...])`?

TotalVerb:
Also, might it be easier simply to define a generic `writemime` method that delegates to `show`? I don't know if this will cause problems; maybe ambiguity?

```julia
Base.writemime{M}(io::IO, ::Type{MIME{M}}, x) = Base.show(io, MIME{M}, x)
```

stevengj:
In 0.5, you are supposed to define `show(io::IO, ::MIME"text/plain", ::TestCustomShowType)`, not `show(io::IO, ::TestCustomShowType)`, so `@compat` should support the latter.

stevengj:
@TotalVerb, Base no longer calls `writemime` at all, so adding another `writemime` method won't help.

TotalVerb:
@stevengj I meant that it may be possible for that code would be included in `Compat` for older versions, so that one can just define a method of `show` and the fallback `writemime` will call it.

src/Compat.jl
```diff
@@ -438,6 +442,8 @@ function _compat(ex::Expr)
             rewrite_pairs_to_tuples!(ex)
         elseif VERSION < v"0.4.0-dev+1246" && f == :String
             ex = Expr(:call, :bytestring, ex.args[2:end]...)
+        elseif length(ex.args) == 4 && ex.args[1] === :show
```
tkelman: this should be version dependent otherwise it's calling a deprecated function on 0.5

TotalVerb:
This seems to work in some situations, but sometimes more code is needed for it to work cleanly. It might be better to just to go along with the proposal here.

```
# Provide writemime ⇒ show fallback
if VERSION < v"0.5.0-dev+4356"
    Base.show(x::IO, y::MIME"text/plain", z) = show(x, z)
    Base.show(x::IO, y::AbstractString, z) = show(x, MIME(y), z)
    Base.writemime(x::IO, y, z) = show(x, y, z)
end
```

TotalVerb:
@dancasimiro Do you have code that demonstrates the problem? I have a similar solution in place on Currencies.jl's master branch, and it works for my (very limited) use case.

stevengj:
@TotalVerb, that will (a) break `mimewritable` (which is used e.g. by IJulia to detect which mime formats are available for a given type); (b) not work if there is an more specific `writemime` method (such as the one for `AbstractVector`) if you want to override it for an even more specific type (e.g. your own vector type); and (c) does not seem like it should work even for overriding the default `writemime(io::IO, ::MIME"text/plain", x) = show(io, x)` that is already in Base in 0.4, though if you define a 2-argument `show` for text/plain it will continue to work (unless there is already a more specific `writemime`, as for `AbstractVector`).

I really think we need a `@compat show(io, ...) = ....` rewriting macro to cover all of the bases here.

TotalVerb:
Point taken. I think the solution in this PR is best then.

dancasimiro:
The new tests that I wrote are passing on 0.3, 0.4, and 0.5. Is there anything else that I need to address before squashing and rebasing?

TotalVerb:
I still would like a `Base.show` rewriting, but having something is good. :+1:

nalimilan:
Could you add a short explanation to README.md?

timholy:
Thanks @dancasimiro!

stevengj:
Note that we should also have a `show(io::IO, m, x) = writemime(io, m, x)` method on 0.4 for people that want to *call* the three-argument `show`, not to define new methods.  You shouldn't be forced to call `@compat` for this.

PR #224, opened 2016-06-09
Compat for new unsafe wrapper methods

malmaud:


src/Compat.jl
```diff
@@ -1245,4 +1245,13 @@ else
     end
 end
 
+if VERSION < v"0.4.0-dev+4612"
```
ivarne: Seems like ``v"0.4.0-dev+4612"`` was a really long time ago

stevengj: You probably need git pull --tags

yuyichao: The version script doesn't use tag to determine the name. It's probably just an old version of the script.

tkelman: the version of the script that had been in contrib was wrong until just a few minutes ago

src/Compat.jl
```diff
@@ -1245,4 +1245,13 @@ else
     end
 end
 
+if VERSION < v"0.5.0-dev+4612"
+    export unsafe_string, unsafe_wrap
+    unsafe_wrap(::Type{String}, p::Ptr, own=false) = pointer_to_string(p, own)
+    unsafe_wrap(::Type{String}, p::Ptr, len, own=false) = pointer_to_string(p, len, own)
```
yuyichao: I think there's a method overwrite here. You should probably have `own::Bool` or just forward all the arguments.

test/runtests.jl
```diff
@@ -1188,3 +1188,11 @@ end
 
 io = IOBuffer()
 @test @compat(get(io, :limit, false)) == false
+
+let
+    s = "test"
```
omus: I think the test is more informative if the data contains the null character. Something like `"test\0foo"`

src/Compat.jl
malmaud: Ah, thanks.

yuyichao:
Windows nightly failure is probably https://github.com/JuliaLang/julia/issues/16921

malmaud:
Ugh, do we really still need to support .3?

yuyichao:
If `pointer_to_string` doesn't exist on 0.3 than the new version shouldn't be supported. `unsafe_string` should be replaced with `bytestring` though. Also need a test for `unsafe_wrap(Array, ...)`

malmaud:
What do you mean? Which `unsafe_string` should be replaced?

yuyichao:
So `pointer_to_array` does exist on 0.3. In this case

`unsafe_wrap(::String)` should be implemented as `String(unsafe_wrap(Array, ...))` (plus some `strlen` `ccall`s, following the base definitions)

`unsafe_string` can either be implemented on top of this (note that you need to copy the array and make a string instead of copying the string, which is a no-op) or using `bytestring` instead.

yuyichao:
I would slightly prefer not exporting another name on 0.3 that is deprecated on master but this LGTM. Needs a update to README.md (and might as well squash it although squash at merge is also fine)

tkelman:
@omus' now-collapsed point that it would be worth testing against input with nulls is still unaddressed

yuyichao:
The embeded NUL isn't part of this API so doesn't has to be tested by Compat.

malmaud:
@tkelman Can this be merged now?

src/Compat.jl
```diff
@@ -1269,4 +1269,26 @@ else
...
+
+end
+
+if VERSION < v"0.5.0-dev+4612"
+    export unsafe_string, unsafe_wrap
+    unsafe_wrap(::Type{Compat.String}, p::Ptr, own::Bool=false) = pointer_to_string(p, own)
+    unsafe_wrap(::Type{Compat.String}, p::Ptr, len, own::Bool=false) = pointer_to_string(p, len, own)
+    unsafe_wrap(::Type{Array}, p::Ptr, dims, own::Bool=false) = pointer_to_array(p, dims, own)
+    unsafe_string(p::Ptr{UInt8}) = bytestring(p)
+    unsafe_string(p::Ptr{UInt8}, len) = bytestring(p, len)
```
stevengj: There are also `unsafe_string` methods that take a `Cstring` argument instead of `Ptr{UInt8}`.

PR #225, opened 2016-06-11
Support get(io, :compact, false) and get(io, :multiline, false)

nalimilan:
These are the two other IOContext properties at the moment.
limit_output() was used to mean both limit=true and compact=true.


TotalVerb:
What about `IOContext` itself? I think that might be a useful backport?

nalimilan:
I guess it could be useful in some advanced cases, by in most of them you can use `showcompact` instead to tweak `IOContext` under the hood, and it already exists in 0.4.

nalimilan:
OK to merge?

nalimilan:
I'll merge tomorrow if nobody objects.

timholy:
Thanks!

PR #226, opened 2016-06-15
Reword description of writemime changes

timholy:
@dancasimiro, I thought this might be slightly clearer? (I thought your description in your PR was very good, so it seems worth linking.)


dancasimiro:
This change looks good to me.

PR #229, opened 2016-06-18
[WIP, RFC] mock for IOContext

TotalVerb:
This is what's being used by Currencies.jl right now. Defining `get` on IO doesn't seem to be an issue since it doesn't conflict with existing methods.

If there's demand for this kind of thing, I'll add tests and docs.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1296,4 +1296,22 @@ if !isdefined(Base, :allunique)
     export allunique
 end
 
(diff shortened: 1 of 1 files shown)


TotalVerb:
I've realized that `IOContext` is something that's incredibly easy to implement in a package itself. In fact, usually the default `IOContext` is not even used, so I don't think it's worth providing it.

PR #230, opened 2016-06-19
Fix qualified show rewrite

TotalVerb:
This addresses the missing case in #219.

Could a new version be tagged soon? I got hit by this when updating Hiccup. Hiccup tests also didn't catch it, so luckily I double checked! (I noticed the omission in the original PR, but since tests passed I thought it might have been fixed in the interim.)

While debugging this, I noticed that the `Base.Timer` rewrite was untested and didn't actually work. I decided to fix that along the way.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -192,7 +192,7 @@ function rewrite_show(ex)
     if length(argtypes) == 2
         insert!(argtypes, 2, Expr(:(::), :macrocall, MIME"text/plain"))
     end
-    return Expr(:call, :writemime, argtypes...)
+    return Expr(:call, :(Base.writemime), argtypes...)
 end
 
 function rewrite_dict(ex)
@@ -456,7 +456,8 @@ function _compat(ex::Expr)
             rewrite_pairs_to_tuples!(ex)
         elseif VERSION < v"0.4.0-dev+1246" && f == :String
             ex = Expr(:call, :bytestring, ex.args[2:end]...)
-        elseif VERSION < v"0.5.0-dev+4340" && length(ex.args) > 2 && ex.args[1] === :show
+        elseif VERSION < v"0.5.0-dev+4340" && length(ex.args) > 2 &&
+                (macroexpand(ex.args[1]) in (:show, :(Base.show)))
             ex = rewrite_show(ex)
         end
         if VERSION < v"0.5.0-dev+4305"
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -491,7 +491,7 @@ function _compat(ex::Expr)
         end
     elseif VERSION < v"0.4.0-dev+5322" && ex.head === :(::) && isa(ex.args[end], Symbol)
         # Replace Base.Timer with Compat.Timer2 in type declarations
-        if ex.args[end] === :Timer || ex.args[end] == :(Base.Timer)
+        if ex.args[end] === :Timer || macroexpand(ex.args[end]) == :(Base.Timer)
             ex.args[end] = :(Compat.Timer2)
         end
     elseif ex.head === :quote && isa(ex.args[1], Symbol)
--- a/src/nullable.jl
+++ b/src/nullable.jl
@@ -1,6 +1,10 @@
 import Base: eltype, convert, get, isequal, ==, hash, show
 export Nullable, NullException, isnull
 
+# FIXME: v0.3 on linux possibly has strange failure if this line, seemingly a
+# no-op, is excluded
+export eltype
+
 immutable Nullable{T}
     isnull::Bool
     value::T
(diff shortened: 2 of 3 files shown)


TotalVerb:
CI failures are due to unrelated issues.

timholy:
Really nice, @TotalVerb. I also got bit by this in https://github.com/JuliaGraphics/Cairo.jl/issues/143. Unlike you, I got bit :frowning:.

The test failure on julia-0.3 (linux-only, curiously) is a little bit alarming: "`Nullable{Float32}` has no method matching Nullable{Float32}()". Did something else change?

TotalVerb:
I cannot reproduce the problem locally.

```
Julia Version 0.3.12
Commit 80aa779 (2015-10-26 12:41 UTC)
Platform Info:
  System: Linux (x86_64-unknown-linux-gnu)
  CPU: Intel(R) Core(TM) i7-3632QM CPU @ 2.20GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas
  LIBM: libopenlibm
  LLVM: libLLVM-3.3
```

TotalVerb:
Let's restart Travis.

tkelman:
I can reproduce the error on Ubuntu 14.04, julia 0.3

TotalVerb:
I wish I could say this was another symlink failure. But unfortunately I am actually testing the right package. I can't reproduce.

Does Travis rebase on master? Maybe it was another change that did this. I will rebase and check.

TotalVerb:
Also, we should fix the nightly failures. I think those are in master.

timholy:
@tkelman, is the 0.3 failure dependent on this PR?

TotalVerb:
Nightly failures see #232.

TotalVerb:
v0.3 failure is hard to reproduce (I have to go through Travis) but I think I've got something licked. It is however extremely strange.

Insertion of the following, after the definition of `Nullable`:

```julia
immutable Dummy end
```

fixes the issue on Travis, if and only if it is exported. I hope this is not just random chance that creates this strange outcome. This may possibly be a bug in Julia 0.3 itself.

TotalVerb:
To add, I don't think this PR is actually the cause of the issue. I suspect that the underlying issue is something very sensitive to small, possibly unrelated, changes. Memory corruption being the underlying issue also explains why it's so platform dependent.

TotalVerb:
Travis is green now (except for v0.5 failure, which is #232). This workaround is not ideal, but I don't want what looks like a probable Julia v0.3 bug holding up this PR, especially if it is hard to fix.

tkelman:
https://github.com/JuliaLang/Compat.jl/pull/230/commits/80d99b477760341dd4ca7d8b9daca3d35fadc2b0 seems good enough for me. The reproduction was from checking out this branch of @TotalVerb's fork. Thanks for looking into it and finding a fix, @timholy and I had mentioned in person how much we both appreciate the great work you've been doing.

tkelman:
I'm actually tempted to disallow new tags of packages that support 0.3 once master of Julia has moved on to be 0.6-dev and I change PkgEval to test vs 0.4, 0.5 and 0.6. Right now it still tests against 0.3 but doesn't report it on the web site, I manually check the commit log and so far it's been pretty much only the flaky packages that ever fail on 0.3 since most other packages have moved on to not actively supporting new tags on it any more.

tkelman:
Oh dear, I think this managed to break JuMP's tests on 0.3? I'm confused how though. Pinning Compat back to v0.8.1 makes them pass. cc @mlubin @joehuchette 

https://gist.github.com/b57fce33157184dbd3f090b41a52bbce

timholy:
Plenty of evidence that this somehow triggered corruption on 0.3; I am reminded of https://github.com/JuliaLang/julia/issues/13306.

If no simple solution presents itself, I vote we retrospectively put a julia-0.4 requirement on Compat, and drop 0.3 support. Probably should retrospectively bump the version to 0.9.

TotalVerb:
I do wonder whether corruption was actually introduced in this PR, or if it was always there and this PR just managed to expose it on certain configurations somehow. I agree that dropping 0.3 support might be the best option.

TotalVerb:
For what it's worth, I can reproduce the JuMP failure.

tkelman:
I was going to propose adding upper bounds on the Compat requirement to the 0.3-supporting past tags of JuMP. If this change to Compat has introduced some un-fixable issue that only manifests on Julia 0.3, that's as good a reason as any to declare package versions that support 0.3 as frozen from now on. We should raise this somewhere visible, julia-users and maybe also julia-dev and julia-news just so people are aware.

tkelman:
Likely related - https://github.com/JuliaDiff/TaylorSeries.jl/issues/54 and https://github.com/JuliaDiff/TaylorSeries.jl/issues/53 cc @lbenet

TotalVerb:
If this problem occurs on 0.4, then that's another story.

TotalVerb:
As for JuMP... the problem occurs for me pinning to 0.8.1 also.

TotalVerb:
The JuMP failure is an off-by-one error introduced in #219, that was exposed in this PR. Phew! Should be an easy fix.

PR #234, opened 2016-06-26
Check argument count before accessing expression arguments

### LINKED ISSUE #228: @compat get for Nullable types throws a BoundsError
I am the maintainer of SimJulia and recently the @compat macro for getting a Nullable type in Julia v0.3 and Julia v0.4 throws a BoundsError.

I isolated the problem with following small example:

`using Compat

type Test
a :: Float64
end

type TestCompat
t :: @compat Nullable{Test}
function TestCompat()
tc = new()
tc.t = @compat Nullable{Test}()
return tc
end
end

tc = TestCompat()
@compat get(tc.t)`

In Julia v0.4 without the @compat macro everything works fine (a NullException is thrown) but with the macro I get the following error message:
ERROR: LoadError: BoundsError: attempt to access 2-element Array{Any,1}:
:get
:(tc.t)
at index [3]

yuyichao:
Fix #228

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -269,34 +269,39 @@ function rewrite_pairs_to_tuples!(expr::Expr)
     return expr
 end
 
+function is_quote_symbol(ex::ANY, val::Symbol)
+    if isa(ex, QuoteNode)
+        return (ex::QuoteNode).value === val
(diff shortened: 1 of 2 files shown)


tkelman:
LGTM

PR #237, opened 2016-06-27
Remove rewriting of two-argument show

TotalVerb:
On v0.4 and before, it is correct to define two-argument `Base.show`, and there is an automatic fallback for three-argument `Base.writemime` with `MIME("text/plain")`. However the previous rewriting was rewriting two-argument `Base.show` to three-argument `Base.writemime`, which broke JuMP's tests. This removes that unnecessary rewriting.

The rest of the changes are general cleanups that are unrelated to the JuMP failure. I removed the extra `export` in hopes that fiddling with the code would fix whatever problem required it.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -189,10 +189,7 @@ end
 
 function rewrite_show(ex)
     argtypes = ex.args[2:end]
-    if length(argtypes) == 2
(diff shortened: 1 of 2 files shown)


tkelman:
Great. We should probably tag this today?

PR #240, opened 2016-06-28
Add `threadid()` and `nthreads()`

### LINKED ISSUE #223: provide Threads.nthreads(), Threads.threadid()
We already have a `Compat.Threads` module, and it would be good to include more functions.   In particular, `nthreads()` and `threadid()` are useful for defining thread-local variables, and can default to `1` on older versions.

ranjanan:
Fixes #223 .

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1078,7 +1078,9 @@ if !isdefined(Base, :Threads)
         macro threads(expr)
             return esc(expr)
         end
-        export @threads
+        const threadid() = 1
+        const nthreads() = 1
+        export @threads, threadid, threadid
     end
     export Threads
 end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1080,7 +1080,7 @@ if !isdefined(Base, :Threads)
         end
         threadid() = 1
         nthreads() = 1
-        export @threads, threadid, threadid
+        export @threads, threadid, nthreads
     end
     export Threads
 end
@@ -1279,7 +1279,7 @@ else
 end
 
 if !isdefined(Base, :view)
-    const view = slice 
+    const view = slice
 end
 
 if !isdefined(Base, :pointer_to_string)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1078,8 +1078,8 @@ if !isdefined(Base, :Threads)
         macro threads(expr)
             return esc(expr)
         end
-        const threadid() = 1
-        const nthreads() = 1
+        threadid() = 1
+        nthreads() = 1
         export @threads, threadid, threadid
     end
     export Threads
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1142,6 +1142,12 @@ using Compat.Threads
     @test true
 end
 
+# Issue #223
+if VERSION < v"0.5"
+    @test threadid() == 1
+    @test nthreads() == 1
+end
+
 @test @compat(Symbol("foo")) === :foo
 @test @compat(Symbol("foo", "bar")) === :foobar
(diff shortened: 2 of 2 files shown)


src/Compat.jl
```diff
@@ -1078,7 +1078,9 @@ if !isdefined(Base, :Threads)
         macro threads(expr)
             return esc(expr)
         end
-        export @threads
+        const threadid() = 1
+        const nthreads() = 1
```
yuyichao: No const

test/runtests.jl
```diff
@@ -1142,6 +1142,12 @@ using Compat.Threads
     @test true
 end
 
+# Issue #223
+if VERSION < v"0.5"
```
yuyichao: Fwiw this condition is always true even on master.

ranjanan: Right, sorry. I'll change that to VERSION > v"0.4"

ranjanan: Yeah, I'll change this to `VERSION < v"0.5.0"

yuyichao:
Maybe aa well just relax the test

ranjanan:
@yuyichao You meant not have the test? Because in v0.5, you can start Julia with more than one thread. Then Compat tests would fail.

yuyichao:
No i mean sth like just making sure both exist and 0 < id < n

ranjanan:
Umm, but what would `n` be? Shall I simply say `nthreads() >= 1` and `threadid() == 1` and relax the condition?

yuyichao:
n was short for `nthreads` (and the second `<` was supposed to be `<=`) we can do `1 == threadid() <= nthreads()` too if we want to guarantee that the test will be run on the first thread.

yuyichao:
Also, I think `v"0.5.0" === v"0.5"`

ranjanan:
Okay, I'll simply remove the condition and do `1 == threadid() >= nthreads()`

ranjanan:
@yuyichao I made the same mistake again. Could you please switch off the CI for the previous commits and enable it for the latest commit? Thanks.

yuyichao:
I canceled the extra travis build. I can't do that for AppVeyor ones but I think that one has auto detection.

Anyway, this LGTM now.

test/runtests.jl
```diff
@@ -1142,6 +1142,9 @@ using Compat.Threads
     @test true
 end
 
+# Issue #223
+@test 1 == threadid() <= nthreads()
+
```
yuyichao: Need import or qualified name.

ranjanan: I think the bug is in a previous line.

src/Compat.jl
```diff
@@ -1078,7 +1078,9 @@ if !isdefined(Base, :Threads)
         macro threads(expr)
             return esc(expr)
         end
-        export @threads
+        threadid() = 1
+        nthreads() = 1
+        export @threads, threadid, threadid
```
ranjanan: Yeah, this is the bug. The second `threadid` should be `nthreads`

ranjanan:
It seems `tryparse(Int32, "0", 1)` does not throw an error. It used to throw an `ArgumentError` on v0.4? Is this expected behavior on v0.5?

yuyichao:
Probably https://github.com/JuliaLang/julia/pull/17078 @quinnj

ranjanan:
@yuyichao I see that #17333 has been merged. Could you restart the build for this PR please?

yuyichao:
I can't restart the appveyor here so it's easier if you can rebase this pr.

TotalVerb:
New failure should be fixed by #249.

PR #244, opened 2016-07-01
unsafe_string should accept both UInt8 and Int8 pointer type

r9y9:
As Julia master does:

```jl
julia> methods(unsafe_string)
# 3 methods for generic function "unsafe_string":
unsafe_string(s::Cstring) at c.jl:69
unsafe_string(p::Union{Ptr{Int8},Ptr{UInt8}}) at strings/basic.jl:56
unsafe_string(p::Union{Ptr{Int8},Ptr{UInt8}}, len::Integer) at strings/basic.jl:52

julia> versioninfo()
Julia Version 0.5.0-dev+5063
Commit 8ecd657* (2016-07-01 04:21 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin15.5.0)
  CPU: Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, haswell)
```

Without this, e.g. the following doesn't work (on v0.4.5 at least):

```jl
using Compat
unsafe_string(ccall((:getenv, "libc"), Ptr{Cchar}, (Ptr{UInt8},),
            Compat.unsafe_convert(Ptr{UInt8}, "HOME")))
```
```
LoadError: MethodError: `unsafe_string` has no method matching unsafe_string(::Ptr{Int8})
while loading In[28], in expression starting on line 1
```

when Cchar is aliased to Int8.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1277,7 +1277,7 @@ else
 end
 
 if !isdefined(Base, :view)
-    const view = slice 
+    const view = slice
 end
 
 if !isdefined(Base, :pointer_to_string)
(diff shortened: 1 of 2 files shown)


stevengj:
You need `@compat` to use `Union{...}`, since that syntax won't work on 0.3.  Also, `unsafe_wrap` needs the same fix.

I think #245 should supersede this.

PR #248, opened 2016-07-09
AbstractIOBuffer was not created until after call overloading

tkelman:
ref https://github.com/JuliaLang/julia/pull/11554


tkelman:
Failures are all unrelated.

TotalVerb:
Does Compat.jl still support v0.4 prereleases?

tkelman:
As much as possible it should try to not break on them. I found this because I was bisecting a bug on Base, and it's a pain to also have to worry about changing package versions at the same time.

PR #249, opened 2016-07-11
fix tests to work with JuliaLang/julia#16590

stevengj:
This fixes the test script, which was failing on nightly due to JuliaLang/julia#16590.  Also got rid of a couple of `String -> unsafe_string` deprecations in the tests.


stevengj:
Also fixes another failure due to JuliaLang/julia#16893

PR #252, opened 2016-07-13
Provide promote_eltype_op

martinholters:
Needed due to https://github.com/JuliaLang/julia/pull/17398.

For Julia 0.3, which lacks `promote_op`, the definition is questionable. Maybe better only provide `promote_eltype_op` for Julia 0.4?


tkelman:
I wouldn't worry all that much about 0.3. Only a few packages are actively tagging new versions that support it. We haven't formally done so yet, but I anticipate "freezing" metadata for new package tags that support Julia 0.3 shortly after 0.5 comes out, though with a possibility for exceptions on a case-by-case basis. So the minimal thing you're doing here for now seems fine to me.

tkelman:
Will merge later if no one objects.

TotalVerb:
@martinholters Needs a rebase.

martinholters:
Ok, rebased.

martinholters:
Someone care merging this?

PR #253, opened 2016-07-15
Add `normalize` and `normalize!`

### LINKED ISSUE #246: Add Base.normalize and normalize! to Compat
It would be great to have Base.normalize and normalize! in Compat. Normalization is a very basic and common concept - having it in Compat would allow 0.4-compatible packages to use it and to provide new methods for it. It would be very natural to have `normalize!(h::StatsBase.Histogram)`, for example.

ranjanan:
Add `normalize` and `normalize!`, which normalizes a vector according to the p-norm. Fixes #246 .

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1112,6 +1112,9 @@ if !isdefined(Base, :normalize)
         v
     end
 
+    copy_oftype{T,N}(A::AbstractArray{T,N}, ::Type{T}) = copy(A)
+    copy_oftype{T,N,S}(A::AbstractArray{T,N}, ::Type{S}) = convert(AbstractArray{S,N}, A)
+
     function normalize(v::AbstractVector, p::Real = 2)
         nrm = norm(v, p)
         if !isempty(v)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1091,6 +1091,77 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
+if !isdefined(Base, :normalize)
+    @doc """
+        normalize!(v, [p=2])
+
+    Normalize the vector `v` in-place with respect to the `p`-norm.
+
+    Inputs:
+
+    - `v::AbstractVector` - vector to be normalized
+    - `p::Real` - The `p`-norm to normalize with respect to. Default: 2
+
+    Output:
+
+    - `v` - A unit vector being the input vector, rescaled to have norm 1.
+            The input vector is modified in-place.
+
+    See also:
+    `normalize`, `qr`
+    """ -> 
+    function normalize!(v::AbstractVector, p::Real=2)
+        nrm = norm(v, p)
+        __normalize!(v, nrm)
+    end
+
+    @inline function __normalize!(v::AbstractVector, nrm::AbstractFloat)
+        #The largest positive floating point number whose inverse is less than
+        #infinity
+        δ = inv(prevfloat(typemax(nrm)))
+        if nrm ≥ δ #Safe to multiply with inverse
+            invnrm = inv(nrm)
+            scale!(v, invnrm)
+        else # scale elements to avoid overflow
+            εδ = eps(one(nrm))/δ
+            scale!(v, εδ)
+            scale!(v, inv(nrm*εδ))
+        end
+        v
+    end
+
+    @doc """
+        normalize(v, [p=2])
+
+    Normalize the vector `v` with respect to the `p`-norm.
+
+    Inputs:
+
+    - `v::AbstractVector` - vector to be normalized
+    - `p::Real` - The `p`-norm to normalize with respect to. Default: 2
+
+    Output:
+
+    - `v` - A unit vector being a copy of the input vector, scaled to have norm 1
+
+    See also:
+    `normalize!`, `qr`
+    """ ->
+    function normalize(v::AbstractVector, p::Real = 2)
+        nrm = norm(v, p)
+        if !isempty(v)
+            vv = copy_oftype(v, typeof(v[1]/nrm))
+            return __normalize!(vv, nrm)
+        else
+            T = typeof(zero(eltype(v))/nrm)
+            return T[]
+        end
+    end
+
+    export normalize, normalize! 
+    
+end
+
 if !isdefined(Base, :AsyncCondition)
     const AsyncCondition = Base.SingleAsyncWork
 else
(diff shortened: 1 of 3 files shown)


ranjanan:
Failing on v0.3 :( I added docstrings to the functions, but there's no `@doc` defined for v0.3.

TotalVerb:
@ranjanan I think the standard way to deal with docstrings is to use the syntax without `@doc`, that is:

```julia
"""
Docstring
"""
function x() end
```

This degrades to a no-op on v0.3, and will produce the correct docstring on v0.4.

ranjanan:
Yes, I tried that first, but the docs won't show up at `?help`. @yuyichao told me this was by design. That was why I shifted to using `@doc`.

TotalVerb:
It is really no surprise that docs won't show up on v0.3, as docstrings were not implemented until v0.4. The `@doc` macro won't really help, as that was also implemented in v0.4.

The docs should show up on v0.4 without issue using the docstring syntax.

ranjanan:
@TotalVerb I meant that on v0.4, the normal docstring syntax within an `if` block does _not_ work by design. On v0.3 `@doc` wasn't part of Base Julia so it's not going to work. That is why, at @tkelman 's suggestion, I removed it and linked to it in the README.

TotalVerb:
@ranjanan Ah, you are correct. I apologize for the misunderstanding.

src/Compat.jl
```diff
@@ -1085,6 +1085,43 @@ if !isdefined(Base, :Threads)
     export Threads
 end
 
+if !isdefined(Base, :normalize)
+
```
tkelman: don't put empty lines at the beginning or end of blocks like this

tkelman:
needs rebasing

TotalVerb:
You might need to provide copy_oftype too.

test/runtests.jl
```diff
@@ -1286,3 +1286,12 @@ let
     @test norm(w) === 1.0
     @test norm(normalize!(v) - w, Inf) < eps()
 end
+
+# 0.5 style single argument `@boundscheck`
```
tkelman: rebasing, not merging

ranjanan:
@tkelman , yes, but I screwed up. I rebased on my local branch and my push got rejected, so I  ended merging stuff from my remote branch on my local branch. I need to get better at this.

tkelman:
when you rebase you need to force push over your branch

ranjanan:
So the `norm` test is failing. Which is kinda weird, I [copied this test](https://github.com/JuliaLang/julia/blob/master/test/linalg/generic.jl#L254) from the Julia tests

test/runtests.jl
```diff
@@ -1280,3 +1280,31 @@ end
...
+        end
+    end
+end
+
+#Test potential overflow in normalize!
+let
+    δ = inv(prevfloat(typemax(Float64)))
+    v = [δ, -δ]
+
+    @test norm(v) === 7.866824069956793e-309
```
ranjanan: This is where the v0.3 test fails. `norm(v)` gives `Inf`

ranjanan: You know what, I'm not sure why I'm adding this test for norm. I'm testing normalize and normalize!

ranjanan: Well, actually it seems that `norm` itself behaves differently on v0.3 as compared to v0.4 and v0.5.

ranjanan:
Okay, the function `norm` behaves differently in v0.3 as compared to v0.4 and v0.5. It seems to me that to support `normalize` and `normalize!`, we drop v0.3 support, or warn against flaky behaviour.

stevengj:
At some point the `norm` function's behavior was improved to reduce overflow/underflow problems.   I don't think it's a big deal if `Compat.normalize` has the same limitations as `norm` in a given Julia version.

test/runtests.jl
```diff
@@ -1301,9 +1301,13 @@ end
 let
     δ = inv(prevfloat(typemax(Float64)))
     v = [δ, -δ]
-
+    if VERSION > v"0.4.0-"
```
stevengj: `VERSION` >= v"0.4.0-pre+7164", for JuliaLang/julia#1861

test/runtests.jl
```diff
@@ -1301,9 +1301,13 @@ end
...
+        @test norm(v) === 7.866824069956793e-309
+    end
     w = normalize(v)
     @test w ≈ [1/√2, -1/√2]
-    @test norm(w) === 1.0
+    if VERSION > v"0.4.0-"
+        @test norm(w) === 1.0
```
stevengj: This seems too strict.  Should be `@test norm(w) ≈ 1.0`, and this should work on any version of Julia.

test/runtests.jl
```diff
@@ -1280,3 +1280,32 @@ end
...
+let
+    δ = inv(prevfloat(typemax(Float64)))
+    v = [δ, -δ]
+    if VERSION > v"0.4.0-"
+        @test norm(v) === 7.866824069956793e-309
+    end
+    w = normalize(v)
+    @test w ≈ [1/√2, -1/√2]
+    @test_approx_eq norm(w) 1.0
+    @test norm(normalize!(v) - w, Inf) < eps()
```
TotalVerb: It doesn't matter in this case, but ≈ is preferred to `@test_approx_eq` which is broken on some numeric types.

ranjanan: @TotalVerb You're right. I guess I'll use `isapprox` instead from now on. Would you like me to change it in this PR?

KristofferC: Better to use `@test norm(w) ≈ 1.0`

ranjanan: @TotalVerb @KristofferC it's weird, but the test fails on Julia v0.4 using `isapprox` but passes with `@test_approx_eq`

TotalVerb:
@ranjanan The problem you are seeing on v0.4 is not caused by this change, merely exposed by it. It is rather a leaked rounding mode introduced by a different test. [Here](https://github.com/JuliaLang/Compat.jl/pull/253/files#diff-fce720c43af3c52c862fd7451c7374b8R866), try adding the following line to reset the rounding mode, so

```julia
    setrounding(T, RoundDown)
    @test rounding(T) == RoundDown
```

becomes

```julia
    setrounding(T, RoundDown)
    @test rounding(T) == RoundDown
    setrounding(T, RoundNearest)
```

TotalVerb:
LGTM. Thanks for the contribution!

PR #256, opened 2016-07-20
Fix static parameters in show rewriting

### LINKED ISSUE #255: `show` -> `writemime` for 0.4 doesn't trigger for parametric show methods
It seems that the re-writing of `show` methods to `writemime` methods on 0.4 doesn't get triggered if the method is parametric:

```
julia> type MyType{T} x::T end

julia> using Compat

julia> @compat import Base.show

# this one works correctly:
julia> @compat show(io::IO, ::MIME"text/plain", x::MyType) = print(io, "testing")
writemime (generic function with 35 methods)

# this doesn't seem to be getting translated:
julia> @compat show{T}(io::IO, ::MIME"text/plain", x::MyType{T}) = print(io, "testing")
show (generic function with 108 methods)
```

@dancasimiro, @TotalVerb

TotalVerb:
Closes #255. cc @ssfrr

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -4,6 +4,9 @@ module Compat
 
 using Base.Meta
 
+"""Get just the function part of a function declaration."""
+withincurly(ex) = isexpr(ex, :curly) ? ex.args[1] : ex
+
 if VERSION < v"0.4.0-dev+1419"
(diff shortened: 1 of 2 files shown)


ssfrr:
Thanks! I can confirm that this fixes the issue with my package.

TotalVerb:
Will merge tomorrow if no objections.

PR #259, opened 2016-07-26
Update version.sh script

martinholters:
See https://github.com/JuliaLang/julia/pull/17531

### DIFF
--- a/bin/version.sh
+++ b/bin/version.sh
@@ -1,11 +1,38 @@
-#!/bin/bash
+#!/bin/sh
 
-# Need to be run from a julia repo clone
-# First argument (Optional) is a ref to the commit
+# Needs to be run from a julia repo clone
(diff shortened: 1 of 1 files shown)


bin/version.sh
```diff
@@ -1,11 +1,38 @@
-#!/bin/bash
+#!/bin/sh
 
 # Need to be run from a julia repo clone
```
tkelman: I guess you're just copy-pasting the existing content, but this should be "Needs"

martinholters: Fixed. Also "Optional" -> "optional" in the next line.

tkelman: maybe sync those (and anything else that might come up before this gets merged) into https://github.com/JuliaLang/julia/pull/17434 with a [ci skip] later?

PR #261, opened 2016-07-26
adds AsyncCondition that's more compliant with 0.5 version

ssfrr:
This adds the ability to use `AsyncCondition()` (without giving a callback function) to make a thing that can be `wait`ed on, like the `AsyncCondition` type in 0.5. To simplify, it's built on top of `SingleAsyncWork`.


TotalVerb:
LGTM.

PR #268, opened 2016-08-09
Add entry for the cholfact(HermOrSym) family

andreasnoack:
As discussed in https://groups.google.com/forum/#!topic/julia-users/0VVcvDgeTLM


andreasnoack:
This has to wait for https://github.com/JuliaLang/julia/pull/17909

TotalVerb:
@andreasnoack Now that JuliaLang/julia#17909 is merged, is this ready?

andreasnoack:
Unfortunately, it will have to wait a little longer. Just found https://github.com/JuliaLang/julia/pull/18142 as well. We are also having [problems with the nightly builds](https://status.julialang.org/) of Julia which means the all package testing is unreliable right now. In particular, it affects this PR since it relies on some recent bug fixes in Base.

PR #275, opened 2016-08-23
Support do-block syntax for redirect_std*

timholy:
Ref https://github.com/JuliaLang/julia/pull/18165

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1452,4 +1452,15 @@ else
     end
 end
 
+import Base: redirect_stdin, redirect_stdout, redirect_stderr
+if VERSION < v"0.6.0-dev.374"
+    for (F,S) in ((:redirect_stdin, :STDIN), (:redirect_stdout, :STDOUT), (:redirect_stderr, :STDERR))
+        @eval function $F(f::Function, stream)
+            STDOLD = $S
+            $F(stream)
+            try f() finally $F(STDOLD) end
+        end
+    end
+end
+
 end # module
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * [`normalize`](http://docs.julialang.org/en/latest/stdlib/linalg/?highlight=normalize#Base.normalize) and [`normalize!`](http://docs.julialang.org/en/latest/stdlib/linalg/?highlight=normalize#Base.normalize!), normalizes a vector with respect to the p-norm ([#13681](https://github.com/JuliaLang/julia/pull/13681))
 
+* `redirect_stdout`, `redirect_stderr`, and `redirect_stdin` take an optional function as a first argument, `redirect_std*(f, stream)`, so that one may use `do` block syntax (as first available for Julia 0.6).
+
 ## Renamed functions
 
 * `pointer_to_array` and `pointer_to_string` have been replaced with `unsafe_wrap(Array, ...)` and `unsafe_wrap(String, ...)` respectively.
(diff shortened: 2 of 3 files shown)


timholy:
I don't have a copy of julia 0.3 lying around; anyone have any idea about the error? I tried adding a `@compat` but that didn't seem to help.

TotalVerb:
v0.3 does not support defining a function without methods.

src/redirect.jl
```diff
@@ -0,0 +1,34 @@
...
+Run the function `f` while redirecting `STDERR` to `stream`. Upon completion, `STDERR` is restored to its prior setting.
+"""
+redirect_stderr(f::Function, stream)
+
+"""
+    redirect_stdin(f::Function, stream)
+
+Run the function `f` while redirecting `STDIN` to `stream`. Upon completion, `STDIN` is restored to its prior setting.
+"""
+redirect_stdin(f::Function, stream)
```
stevengj: It seems like overkill to have docstrings in Compat.jl.    We don't do that for any other functions that we copy from newer versions.   

And if you delete the docstrings, you can just put the code directly into `src/Compat.jl`.   And it should fix things on 0.3.

timholy:
Not sure I know what you mean. This is just adding a method to each of the existing `redirect_std*` functions.

Or do you mean it's the docstring that's likely the problem?

timholy:
Sorry, browser refresh delay left things out of sync. I'll delete the docstrings.

timholy:
Hmm, still no dice. I worry it's because of the `Base.$F` part, but I've tried with and without `@compat` and it doesn't seem to matter.

src/Compat.jl
```diff
@@ -1452,4 +1452,20 @@ else
     end
 end
 
+if VERSION < v"0.6.0-dev.374"
+    for (F,S) in ((:redirect_stdin, :STDIN), (:redirect_stdout, :STDOUT), (:redirect_stderr, :STDERR))
+        @eval @compat function Base.$F(f::Function, stream)
```
yuyichao: `function Base.$F() ...` doesn't parse on 0.3. You can use `import` (note that `@eval import Base: $F` won't work on 0.3 either so you need to either use `Expr(:import)` or just list the functions manually outside the loop)

stevengj: I would just import the functions outside the loop.

timholy: If it's a parsing thing, that explains why `@compat` doesn't help.

src/Compat.jl
```diff
@@ -1452,4 +1452,21 @@ else
...
 
+import Base: redirect_stdin, redirect_stdout, redirect_stderr
+if VERSION < v"0.6.0-dev.374"
+    for (F,S) in ((:redirect_stdin, :STDIN), (:redirect_stdout, :STDOUT), (:redirect_stderr, :STDERR))
+        @eval function $F(f::Function, stream)
+            STDOLD = $S
+            local ret
+            $F(stream)
+            try
+                ret = f()
```
TotalVerb: Could save a few lines with `try f() finally $F(STDOLD) end`.

tkelman: it should return the value of `f()` not the value of `$F` though, shouldn't it?

TotalVerb: A `try` block's value is the tail statement in `try` or `catch` position, not the value in `finally` position.

timholy:
Something killed the OSX builds, but it passed on all 3 versions on Linux and Windows.

timholy:
Oh, and thanks for the help, folks!

PR #280, opened 2016-09-05
Compat for at-vectorize_(1|2)arg deprecation

yuyichao:
Provide `Compat.@vectorize_1arg` and `Compat.@vectorize_2arg` as an upgrade
path for packages.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1464,8 +1464,49 @@ if VERSION < v"0.6.0-dev.374"
 end
 
 if VERSION < v"0.6.0-dev.528"
-    macro __DIR__() Base.source_dir() end
+    macro __DIR__()
+        Base.source_dir()
+    end
     export @__DIR__
+end
+
+# PR #17302
+# Provide a non-deprecated version of `@vectorize_(1|2)arg` macro which defines
+# deprecated version of the function so that the depwarns can be fixed without
+# breaking users.
+# Packages are expected to use this to maintain the old API until all users
+# of the deprecated vectorized function have migrated.
+# These macros should raise a depwarn when the `0.5` support is dropped from
+# `Compat` and be dropped when the support for `0.6` is dropped from `Compat`.
+# Modified based on the version copied from 0.6 Base.
+macro dep_vectorize_1arg(S, f)
+    S = esc(S)
+    f = esc(f)
+    T = esc(:T)
+    x = esc(:x)
+    AbsArr = esc(:AbstractArray)
+    ## Depwarn to be enabled when 0.5 support is dropped.
+    # depwarn("Implicit vectorized function is deprecated in favor of compact broadcast syntax.",
+    #         Symbol("@dep_vectorize_1arg"))
+    :(@deprecate $f{$T<:$S}($x::$AbsArr{$T}) @compat($f.($x)))
+end
+
+macro dep_vectorize_2arg(S, f)
+    S = esc(S)
+    f = esc(f)
+    T1 = esc(:T1)
+    T2 = esc(:T2)
+    x = esc(:x)
+    y = esc(:y)
+    AbsArr = esc(:AbstractArray)
+    ## Depwarn to be enabled when 0.5 support is dropped.
+    # depwarn("Implicit vectorized function is deprecated in favor of compact broadcast syntax.",
+    #         Symbol("@dep_vectorize_2arg"))
+    quote
+        @deprecate $f{$T1<:$S}($x::$S, $y::$AbsArr{$T1}) @compat($f.($x,$y))
+        @deprecate $f{$T1<:$S}($x::$AbsArr{$T1}, $y::$S) @compat($f.($x,$y))
+        @deprecate $f{$T1<:$S,$T2<:$S}($x::$AbsArr{$T1}, $y::$AbsArr{$T2}) @compat($f.($x,$y))
+    end
 end
 
 end # module
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -7,6 +7,11 @@ using Base.Meta
 """Get just the function part of a function declaration."""
 withincurly(ex) = isexpr(ex, :curly) ? ex.args[1] : ex
 
+if VERSION < v"0.4.0-dev+2254"
+    immutable Val{T} end
+    export Val
+end
+
 if VERSION < v"0.4.0-dev+1419"
     export UInt, UInt8, UInt16, UInt32, UInt64, UInt128
     const UInt = Uint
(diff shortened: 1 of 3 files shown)


yuyichao:
We also need a better `@compat` for the broadcast syntax for this to be usable.

tkelman:
This is backwards. Should try to provide the new syntax if possible, not just move the old macro here.

yuyichao:
I know this is backward but that's the point.

The issue is that even though packages can use the 0.5 broadcast syntax with `Compat` (well, apart from the fact that the return type is wrong, which I'm trying to fix now) they can't remove the vectorized function without breaking all the users, which means that without the macro in this PR the package will either

1. Break user code
2. Can't fix depwarn
3. Roll their own `@vectorize_*` macro or write them out manually, only to be removed later once all users are upgraded.

As the commit message says, this is basically an easy upgrade path for packages that uses the macro so that they can fix their own use of the vectorized form and make sure user code won't break with minimum change of code.

yuyichao:
The `provide the new syntax if possible` part is https://github.com/JuliaLang/Compat.jl/pull/196, which is not good enough for packages that uses `@vectorize_*` to use yet.

yuyichao:
OK, the return type of the broadcast syntax is fixed in most cases. This makes it as good as comprehension on 0.3/0.4 although not as good as 0.5.

yuyichao:
Haven't added test for the splatting case yet, will do that later.

tkelman:
What's the upgrade path? Packages can just replace uses of the vectorize macros with calls to broadcast.

yuyichao:
Not for packages that uses the macro to define public functions

tkelman:
Is the problem just that the base deprecated version of the macro doesn't give deprecation warnings when you call the vectorized functions that it defines? We could change that. I don't think Compat should be a holding ground for deprecated code.

tkelman:
As it should, it's deprecated. There have only been a handful of cases of putting the deprecated version here, and those were due to major differences in what would work on old julia versions - functorize and Compat.ASCIIString. The deprecated macro now giving deprecation warnings isn't a good reason to put it here without the warnings. There's an easy workaround here of defining vectorized methods manually.

yuyichao:
> There's an easy workaround here of defining vectorized methods manually.

That could be said for the macro before it was deprecate, as well as the aliases for string renaming. In fact, I was doing the string deprecation fix by defining the type in packages before all of them are available in Compat.

The defining vectorized method manually is also not that "easy". For one, the deprecation should only be raised on 0.6 (if not more precise than that) so there has to be a version check. The broadcast / shape / promotion code is also different between versions so it's non-trivial to manually defining these functions in packages in a way that works on all julia versions.

In general, I think this package should define common things that help people fixing depwarns. In most case that's by defining new functions on old julia versions. However, there are some cases when that's not possible and it should be ok to define necessary piece that helps in those cases.

yuyichao:
Also, being a clear import/qualified use from Compat, it's a clear self documentation that something can be removed when support for old julia version is dropped

tkelman:
I don't think Compat should host code that's obsolete from the very start. The way in which this package is intended to help people fix deprecation warnings is by supporting the new syntax on old Julia versions. A LegacyMacros package under JuliaArchive would be a better place for things that are deprecated in base that you still want packages to use without changes. People should stop using these macros, that's the point of deprecating them. If we put them here they'll never die.

yuyichao:
> obsolete from the very start.

I don't see why that's the case. I'm only saying that it's not easier to define those functions manually than what it was before it was deprecated. It was useful which is why people use it. And FWIW, whether the code is obsolete from the start has little to do with how the upgrade should be done, how many users exists does.

> The way in which this package is intended to help people fix deprecation warnings is by supporting the new syntax on old Julia versions.

That's one way and happens to be what works most of the time. It's unfortunately not always the case.

> A `LegacyMacros` package under `JuliaArchive` would be a better place for things that are deprecated in base that you still want packages to use without changes.

That could be a good place for the `@ngenerate` and related macros. This is not one of those since this does not make it possible to use the old code without depwarn. The macro still give a depwarn when the function defined is called and people should be aware of that and move away from actually using the function defined.

As mentioned before, the correct upgrade path for a package that uses these macros is,

1. Deprecate the use of the defined functions

    This is done automatically with both `Compat.@vectorize_*` and `Base.@vectorize_*`

2. Fix use of the defined functions

3. Mark that the function defined by the macro are deprecated and should be removed.

    This is achieve by the `Compat.`, which should be a mark for something needs to be done, and possibly some comment where the macro is used.

4. When enough/all users are migrated, remove the old definitions (i.e. the use of the `Compat.` macros).

It is true that it can theoretically take one release cycle longer to be removed them from `Compat`. Whether that will actually happen depends on how many users of the macro have got to step 4 when we drop 0.5 support from `Compat`. Given that most (all?) users of the macro drops julia version supports much faster than `Compat`, I don't think that's a real issue. (In the worst case, we just need to raise a depwarn from the Compat version for one release cycle after 0.5 support is dropped from Compat).

TotalVerb:
I wonder if it might be worth having an additional module `Deprecations.jl` (or perhaps `Compat.Deprecations`) that contains helper functions to deprecate functionality. Examples of exported macros may include `Deprecations.@deprecate`, `Deprecations.@deprecate_vectorize_1arg`, `Deprecations.@deprecate_binding`, etc. To me, this functionality does not seem like "compatibility" functionality, but rather convenient macros to generate methods that have `depwarn` calls included.

yuyichao:
Well, it's "compatibility" in that it's to make the code being able to run on multiple julia versions.

I don't see why putting it in a module would help anything apart from making using Compat harder but if that makes people happy I can certainly do that.

TotalVerb:
Putting it in a module keeps related items together, and I think this `@deprecate_vectorize_(1|2)arg` is definitely more similar to `@deprecate` or `@deprecate_binding` (which the new module can import from `Base`, and re-export) than it is to the existing compatibility code. It also opens the way for more complex deprecation patterns to be added, such as the currently-non-existent `@deprecate_without_replacement`, or `@deprecate_with_message`, which cannot be included in `Compat` because they make little sense here.

yuyichao:
Also it's not always deprecated, only on 0.6 (to avoid breaking users, ref https://github.com/JuliaLang/ZMQ.jl/pull/109), so calling it `deprecate` will be strange.

TotalVerb:
Hmm right, that is an argument for keeping it in `Compat`, since it has different behaviour across versions.

tkelman:
Very few users or package developers should be caring about deprecation warnings on 0.6-dev yet. By the time they do, a much simpler upgrade path is just not calling the base vectorize macro on 0.6-dev or newer.

yuyichao:
What's the "Base vectorize macro"?

tkelman:
put the call to `vectorize_1arg` behind a version conditional in the packages that use it.

yuyichao:
How is that different from manually defining the function in packages which I've already argued against above?

yuyichao:
Also, `Very few users or package developers should be caring about deprecation warnings on 0.6-dev yet.` is not a very good argument to be made if the argument of base bisectability on packages needs to be made at other places.

tkelman:
Deprecation warnings don't usually make much difference to bisectability. Migrating package code to the new syntax is the right way to get rid of the deprecations, and going through putting this code here is a distraction from accomplishing that.

yuyichao:
Return type of the broadcasting syntax with scalar input is fixed. Also add more tests.

yuyichao:
> Migrating package code to the new syntax is the right way to get rid of the deprecations

So how do you propose the fix for a package that use the macro to define public functions? How can it make sure users are not broken?

> going through putting this code here is a distraction from accomplishing that.

How so? Putting the code here makes fixing those depwarns significantly easier and should encourage people from actually fixing it instead of deciding if the extra complexity of including the code in (the first commit of) this PR in the package is worth doing. I've seen multiple times that some package will simply waiting for the actual release to come out and suddenly switch the API breaking all users on the new release.

yuyichao:
Also

> Deprecation warnings don't usually make much difference to bisectability

is not true. Printing the back trace involves calling a significant part of the runtime (backtrace, debug info, IO) and can easily cause/hide many problems. Let along the annoyance of finding the actually error messages / warnings walls of unrelated depwarns.

tkelman:
Don't define those functions any more on 0.6 because it's obsolete there. Before doing that, change the base deprecation of vectorize_1arg so the functions it defines give warnings, so it's more visible.

yuyichao:
> so the functions it defines give warnings

It already does.

> so it's more visible.

How do you expect people using the defined functions to care if the packages that uses the macro already have depwarns due to the macro itself.

I see you point being that we don't need this now. I disagree for a few reasons,

1. I don't think fixing depwarn just before 0.6 is release is a better option than fixing them now.

    We obiously shouldn't require packages to be depwarn free on master but being able to do so or earlier should be encourage. Fixing the depwarn later in the package doesn't mean the support for the previous version can be immediately dropped, but just much less time to work on it and potentially exposing much more problems that needs to be fixed in a tighter time window. The return type of the broadcast syntax on 0.3/0.4 is a good example of such problems. I wouldn't notice it and fix it earlier if I didn't need to use it in this PR.

2. I believe fixing the depwarn should start from the leaf dependency down. At least so that it's much easier to see all the warnings. The macro and the proposed upgrade path in this PR does exactly that, while fixing it in other ways and fixing them later does exactly the opposite, leaving users in a broken state and possibly less time to be fix.

tkelman:
Not suggesting the deprecation shouldn't be fixed now, if people want to fix them then go ahead. But I don't think adding this transition code to Compat is valuable for doing that. Fix the uses of the deprecated vectorized functions, then make the macro calls conditional on julia version after that has been done.

yuyichao:
> Fix the uses of the deprecated vectorized functions, then make the macro calls conditional on julia version after that has been done.

So if I want to make a PR that fix this, I should drop the support for 0.5 on that package so that working user code won't break?

yuyichao:
I see that I don't need to drop 0.5 support but I disagree that this is the right direction to fix the depwarn, It requires fixing all users before fixing the package itself, which is totally backward.

tkelman:
Basically I see your proposal here as equivalent to removing the warnings at macro call time from the base deprecation of vectorize_1arg, but with a timeline that makes less sense since anyone who starts using the code from Compat needs to be fixed before it could ever be removed here.

yuyichao:
And I just don't understand why putting ~20 line of self containing code that significantly helps package upgrades is called not "valuable".

stevengj:
Hopefully, by the time avoiding the warning on 0.6 becomes important (i.e., close to 0.6's release), people will be ready to drop support for 0.4, and hence they can just use the new dot-call syntax.

yuyichao:
The point is that the Base change adds two depwarn and they can't be fixed at the same time without breaking code. If the depwarn is only doing it one at a time then the function here is not necessary. The change here is basically two provide the right change sequence with the same timeline as what it would have been done in base,

First we deprecate the functions so that users can migrate. Then we deprecate the macros so that the packages that defines those functions can migrate. Finally we remove the micro.

The second step can be skipped if the third step is known to not cause much breakage, which should be helped by doing this earlier.

yuyichao:
> hence they can just use the new dot-call syntax.

With the fix in this PR they can do that on 0.3. The point is that not doing the change in this PR requires fixing the users first, which is backward.

tkelman:
I don't think working around the macro call time depwarn is worth duplicating the code here. I think we both agree that the call of vectorize_1arg isn't the actual problem here, it's the calls to the functions it defines. Until those are fixed, a depwarn in the package about its use of vectorize_1arg is serving a valid purpose, and changing it to call Compat.vectorize_1arg to make the macro call time warning go away just makes the issue less visible.

yuyichao:
> Until those are fixed, a depwarn in the package about its use of vectorize_1arg is serving a valid purpose,

No because the use can be in other packages.

yuyichao:
And FWIW this is exactly what base do. The macros here is just the `@dep_vectorize_*` macros on the master. I'm copying it instead of renaming it only because this potentially needs to be kept for one version longer than the base one and I want to avoid breaking Compat as much as possible.

TotalVerb:
If this function were renamed `Compat.@deprecate_vectorize_1arg`, I can get behind having it here. This makes it more clear that the extra methods defined are deprecated.

yuyichao:
Sure. I don't personally think we need more clarification here compare to the string change but I can add the extra characters if that makes everyone happy.

tkelman:
Why should Compat continue providing old code after it goes through a base deprecation cycle? Implementing the equivalent of master's dep_vectorize seems better.

TotalVerb:
I think the improved clarity is important. Reading

```julia
Compat.@vectorize_1arg sin(x) = ...
```

seems like additional vectorized versions are being defined. Whereas

```julia
Compat.@deprecate_vectorize_1arg sin(x) = ...
```

(or `@dep_vectorize_1arg`) makes it clear that any new definitions are deprecated and subject to eventual removal, much like definitions hidden made by `Base.@deprecate`.

yuyichao:
> Why should Compat continue providing old code after it goes through a base deprecation cycle?

Because the base change is actually two sequential deprecations. Of course they are closely related which is why this might not need to stay for longer than base. Another potential reason is that Compat support more julia versions (3 now) than Base (only 1).

> Implementing the equivalent of master's dep_vectorize seems better.

Good. As I said, this is exactly what this PR does.

tkelman:
> Good. As I said, this is exactly what this PR does.

But with a different name from what master uses, which is confusing.

tkelman:
In other words I think it should be called dep_vectorize_1arg here, and the functions it defines should also be deprecations on earlier julia versions if we really want people to move away from them.

yuyichao:
The `@dep_vectorize_*` is a internal name that users shouldn't actually care about which is why I picks the name that should be more clear what it is doing on earlier julia versions.

Now if I use `dep_....` here, what should it do on 0.3-0.5? and (less importantly) should it be exported?

yuyichao:
> the functions it defines should also be deprecations on earlier julia versions if we really want people to move away from them.

It'll be much longer than 30 lines but sure.

TotalVerb:
It should deprecate the vectorized versions on 0.3 to 0.5 also.

yuyichao:
> It'll be much longer than 30 lines but sure.

Actually maybe not assuming we want the right depwarn and can accept minor behavior difference.

src/Compat.jl
```diff
@@ -1463,4 +1601,44 @@ if VERSION < v"0.6.0-dev.374"
...
 
+# PR #17302
+# Provide a non-deprecated version of `@vectorize_(1|2)arg` macro which defines
+# deprecated version of the function so that the depwarns can be fixed without
+# breaking users.
+# Packages are expected to use this to maintain the old API until all users
+# of the deprecated vectorized function have migrated.
+# These macros should raise a depwarn when the `0.5` support is dropped from
+# `Compat` and be dropped when the support for `0.6` is dropped from `Compat`.
+export @dep_vectorize_1arg, @dep_vectorize_2arg
```
tkelman: these aren't exported from base so I don't think they should be exported here

yuyichao: Sure. Do note that the version here and the version in base serves completely different purposes. The base version isn't a public API, this one is.

tkelman: We could make the base one more official, in which case packages could choose to call that on 0.6 instead of the one that throws a deprecation warning.

yuyichao: Well, a Compatible layer in base that doesn't work on multiple julia version isn't very useful.
Packages will have to do version check in order to use the macro in which case they might as well use the Compat version or break API on 0.6.
This will basically means that the `@dep_*` macro needs to be kept in Base for one version longer than the `@vectorize_` which I think is a worse choice, at least for compatibility layer, than putting them here.

yuyichao:
The performance of the broadcast syntax should almost match that of comprehension on 0.4 and the same syntax on master now. The main reason it doesn't is due to `broadcast_shape` being much slower than master. Since that is a Base issue that's fixed, I don't feel like backporting a better `broadcast_shape` implementation to 0.3/0.4. It's not a regression compare to current Compat master and can always be done later if anyone needs it.

yuyichao:
Will merge later today.

Note that this is backportable to 0.8.x if people want it on 0.3.

src/Compat.jl
```diff
@@ -1464,8 +1602,49 @@ if VERSION < v"0.6.0-dev.374"
...
 end
 
+# PR #17302
+# Provide a non-deprecated version of `@vectorize_(1|2)arg` macro which defines
+# deprecated version of the function so that the depwarns can be fixed without
+# breaking users.
+# Packages are expected to use this to maintain the old API until all users
+# of the deprecated vectorized function have migrated.
+# These macros should raise a depwarn when the `0.5` support is dropped from
+# `Compat` and be dropped when the support for `0.6` is dropped from `Compat`.
```
stevengj: Shouldn't they raise a depwarn when 0.4 support is dropped?   dot-call syntax is implemented in 0.5, so once people are using 0.5+ there is no need for `@vectorize`.

yuyichao: This is just the most conservative option since hypothetical a package that is deprecated on 0.6 but works on 0.5 don't need to remove the use of `@dep_vectorize_` since `@vectorize_` is not deprecated there. Hopefully packages will migrate fast enough and we don't need to worry about this too much....

PR #281, opened 2016-09-07
Add @__DIR__ macro.

helgee:
Another nightly PR but this time with a branch and all...

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1463,7 +1463,7 @@ if VERSION < v"0.6.0-dev.374"
     end
 end
 
-if VERSION < v"0.6.0-dev.404"
+if VERSION < v"0.6.0-dev.528"
     macro __DIR__() Base.source_dir() end
     export @__DIR__
 end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1463,4 +1463,9 @@ if VERSION < v"0.6.0-dev.374"
     end
 end
 
+if VERSION < v"0.6.0-dev.404"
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -1463,4 +1463,9 @@ if VERSION < v"0.6.0-dev.374"
     end
 end
 
+if VERSION < v"0.6.0-dev.404"
```
tkelman: did you run the commit-name script at the merge commit?

TotalVerb: This should probably be `!isdefined(Base, Symbol("@__DIR__"))` for clarity

helgee: @tkelman yes. Since CI says that it was wrong, how do I do it correctly? Or should I use @TotalVerb's suggestion?

tkelman: `isdefined` sounds fine to me. Suggests there may be a bug in that script though...

tkelman: hm, I get
```
$ git checkout -q 11ff9abd013e8c8f745616057c1076e6c6b761c5 && contrib/commit-name.sh
0.6.0-dev.528
```

helgee: Obviously I did not run it at the merge commit. Sorry, my bad!

helgee:
With the correct version CI passes 👍

src/Compat.jl
yuyichao: https://github.com/JuliaLang/Compat.jl/pull/281#discussion_r77931523 is still better

PR #288, opened 2016-10-18
add transcode

### LINKED ISSUE #284: transcode function
We should add Compat for the `transcode` function in Julia 0.5, so that packages don't need to add a dependency on the LegacyStrings package just to convert data from UTF-16 or UTF-32 into a `String`.  (See e.g. JuliaPackaging/WinRPM.jl#79)

stevengj:
Adds the `transcode` function; fixes #284.


nalimilan:
Any idea why tests failed on 0.4 on Linux but not on Windows?

stevengj:
wchar_t is different on Linux and Windows, so it is probably that case that is buggy.

stevengj:
Yup, there was a typo in the 32-bit wchar_t code.

PR #289, opened 2016-11-14
Deprecation of `$` in favour of `xor`

simonbyrne:
https://github.com/JuliaLang/julia/pull/18977

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `writemime` has been merged into `show` [#16563](https://github.com/JuliaLang/julia/pull/16563). Note that to extend this function requires `@compat`; see the [Supported Syntax](#supported-syntax) section for more information.
 
+* `$` is now `xor` or `⊻` [#18977](https://github.com/JuliaLang/julia/pull/18977).
+
 ## New macros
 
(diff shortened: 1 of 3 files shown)


nalimilan:
Could you add a test?

GunnarFarneback:
This doesn't cover `⊻=`. Is it at all possible to support that with Compat?
```
julia> using Compat

julia> x = true
true

julia> x $= true
false

julia> x ⊻= true
ERROR: syntax: unexpected "="
```

stevengj:
No, `⊻=` isn't parsed as an operator.  You have to write `x = x ⊻ true`

GunnarFarneback:
Ok. That's a bit unfortunate for the specific piece of code I'm working with since `⊻=` will look nicer once 0.5 support can be dropped.

In case someone else should run into this, this is the alternative solution I chose for the 0.5 to 0.6 transition (skipping Compat):
```
if isdefined(:xor)
    macro x(ex)
        arg1 = ex.args[1]
        arg2 = ex.args[2]
        return esc(:($(arg1) = $(arg1) ⊻ $(arg2)))
    end
else
    const ⊻ = $
    macro x(ex)
        return esc(ex)
    end
end
```
and decorating all use of `$=` with the `@x` macro.

PR #290, opened 2016-11-26
A few fixes for 0.6

yuyichao:
* `num` and `den` rename
* `takebuf_array` rename
* Fix `$` in the test

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,10 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `$` is now `xor` or `⊻` [#18977](https://github.com/JuliaLang/julia/pull/18977).
 
+* `num` and `den` are now `numerator` and `denominator` [#19246](https://github.com/JuliaLang/julia/pull/19246).
+
+* `takebuf_array` is now a method of `take!`. `takebuf_string(io)` becomes `String(take!(io))` [#19088](https://github.com/JuliaLang/julia/pull/19088).
+
 ## New macros
 
 * `@static` has been added [#16219](https://github.com/JuliaLang/julia/pull/16219).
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -1512,3 +1512,13 @@ end
...
 @test 1 ⊻ 5 == 4
+
+# julia#19246
+@test numerator(1//2) === 1
+@test denominator(1//2) === 2
+
+# julia#19088
+let io = IOBuffer()
+    write(io, "aaa")
+    @test take!(io) == UInt8['a', 'a', 'a']
```
yuyichao: Sure. It's actually tested twice above already.

nalimilan: Yes, but it doesn't hurt to do that explicitly here too.

README.md
```diff
@@ -126,6 +126,10 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `$` is now `xor` or `⊻` [#18977](https://github.com/JuliaLang/julia/pull/18977).
 
+* `num` and `den` are now `numerator` and `denominator` [#19246](https://github.com/JuliaLang/julia/pull/19246).
+
+* `takebuf_array` is now a method of `take!`. `takebuf_string(io)` becomes `String(take!(io))` [#19088](https://github.com/JuliaLang/julia/pull/19088).
```
stevengj: Shouldn't this be `Compat.UTF8String(take!(io))` thanks to the `Compat.String` mess?

yuyichao: No, I believe `takebuf_string` on 0.4 is actually type unstable. It can return either `ASCIIString` or `UTF8String`.

stevengj: Then we need to define a ByteString(::Array{UInt8}) method

yuyichao: [Yes?](https://github.com/JuliaLang/Compat.jl/blob/master/src/Compat.jl#L1516)

TotalVerb: Don't we already?

yuyichao: Is there anything that's broken? This won't work before the specific 0.4 commit but Compat.jl already require 0.4.0 now. Travis tests also seems to be passing on 0.4?

stevengj:
Right, sorry, we already have it

PR #293, opened 2016-12-22
fix test script for changes in warn() escape codes

stevengj:
One of the tests got broken by JuliaLang/julia#18628, as discussed in JuliaLang/julia#19682


tkelman:
there's a mac backlog that I'm inclined not to wait for

PR #294, opened 2016-12-27
Add ShortCircuiting

ararslan:
This works around the removal of `Base.ShortCircuiting` in Julia 0.6 by providing `Compat.ShortCircuiting`, which retains the definition from Base on previous versions.

cc @tkelman


tkelman:
ColorVectorSpace looks like the only use of this in packages - https://github.com/search?l=&q=ShortCircuiting+language%3AJulia&ref=advsearch&type=Code&utf8=%E2%9C%93

ararslan:
Closing since this seems to be unnecessary.

PR #292, opened 2016-12-22
add @compat for .=

stevengj:
(closes #285)

PR #295, opened 2016-12-29
fix use of `.primary` field for jb/subtype

JeffBezanson:
This is a simple change that allows the JLD tests to pass on my branch.


tkelman:
Assuming you'll want this tagged nowish? #292 and #288 were features, so I'll tag as 0.10.0 (cc @stevengj)

(almost all Compat tags add features so we should stop using patch increments - this and #293 would be counterexamples in isolation, but those aren't as common as adding new compat backport-features)

PR #296, opened 2016-12-31
Backport the Iterators module to 0.4/0.5

TotalVerb:
This introduces `Compat.Iterators` with the intention that code can do

```julia
using Compat.Iterators
```

instead of

```julia
using Base.Iterators
```

to preserve 0.5 compatibility.

A question: Should the missing iterators `partition`, `product`, and `flatten` be backported to 0.4? These iterators have never been available for 0.4 and nobody has complained so far. `Compat.Iterators` will still correctly contain the iterators that have been available since 0.4.


stevengj:
I think the expectation is that anything in Compat will support all versions of Julia supported by Compat (currently ≥ 0.4).

vchuravy:
bump, this needs a rebase.

ararslan:
I've resolved the branch conflicts.

timholy:
Tests fail.

TotalVerb:
Thanks for the heads up. I'll get this fixed.

tkelman:
what's with the empty file?

TotalVerb:
I'll remove it, my bad.

TotalVerb:
I have tried my best to get all the missing iterators working on 0.4. It's not feasible to replicate the 0.5/0.6 behaviour entirely, because that would require backporting the iterator traits system. In particular, `partition`, `flatten`, and `product` all differ from 0.5/0.6 behaviour in several ways, most visibly being `product`'s lack of shape. (The differences for `partition` and `flatten` occur mostly for odd cases like infinite iterators.)

TotalVerb:
I'll merge this later today if there are no remaining comments.

TotalVerb:
There seems to be a delay with the Travis mac builds. Travis linux and Windows are passing, so I will merge this.

PR #298, opened 2017-01-08
Backport ∘ and ! from #17155

TotalVerb:
This makes ∘ and ! available for 0.5 and prior.


tkelman:
Whoever tags this, we should get in the habit of making them minor tags for features.

There are a few packages defining these locally that this could also break when released.

TotalVerb:
Yeah, packages that this broke include my own, so a minor tag would indeed be a good idea.

TotalVerb:
OK, I tagged it [v0.11.0](https://github.com/JuliaLang/Compat.jl/releases/tag/v0.11.0). Let's let attobot work its magic.

PR #300, opened 2017-01-08
Fix some test errors and deprecation warnings on 0.6

TotalVerb:
This fixes uses of `s.data` that prevent Compat tests from passing. In particular, we no longer check that `unsafe_wrap` aliases properly on sufficiently new versions of 0.6.

This also fixes several deprecation warnings from `@test_approx_eq`.

This does not fix the following remaining deprecation warnings:

- `floor`, `round`, `ceil`, etc. as vectorized functions
- `promote_eltype_op`

because both fixes will likely require (non-test) code changes. I have a solution to `promote_eltype_op` that I will open as a PR shortly.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -402,19 +402,19 @@ end
 # qr, qrfact, qrfact!
 let A = [1.0 2.0; 3.0 4.0]
     Q, R = qr(A, Val{false})
-    @test_approx_eq Q*R A
+    @test Q*R ≈ A
     Q, R, p = qr(A, Val{true})
-    @test_approx_eq Q*R A[:,p]
+    @test Q*R ≈ A[:,p]
     F = qrfact(A, Val{false})
-    @test_approx_eq F[:Q]*F[:R] A
+    @test F[:Q]*F[:R] ≈ A
     F = qrfact(A, Val{true})
-    @test_approx_eq F[:Q]*F[:R] A[:,F[:p]]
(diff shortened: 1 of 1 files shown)


TotalVerb:
Closing and reopening to restart Appveyor.

stevengj:
Since `unsafe_wrap(String, ...)` has disappeared in 0.6, should we drop it from Compat?

TotalVerb:
For as long as we continue to support 0.4, it would break packages that have not yet upgraded to 0.6 if we deprecated or dropped `unsafe_wrap(String, ...)`. However I do agree that we should document that this feature is no longer available in 0.6 or later.

test/runtests.jl
```diff
@@ -1304,19 +1304,29 @@ io = IOBuffer()
...
+    end
     new_str = unsafe_string(ptr)
     cstr = convert(Cstring, ptr)
     new_str2 = unsafe_string(cstr)
     @test wrapped_str == "test"
     @test new_str == "test"
     @test new_str2 == "test"
-    @test ptr == pointer(wrapped_str)  # Test proper pointer aliasing behavior
+    if VERSION < v"0.6.0-dev.1988"
+        # Test proper pointer aliasing behavior, which is of course not
```
tkelman: "of course" is debatable 😉

TotalVerb: Removed this part.

TotalVerb:
Is this good to merge? I'll handle the remaining deprecations and documentation issues later but it's useful to get Travis working again.

PR #301, opened 2017-01-09
Small improvements to the README

TotalVerb:
The first commit removes outdated mentions of 0.3. We have not supported 0.3 for a long time.

The second commit replaces the current long links with shorter `[#nnnnn]` links, for which we run NEWS-update.jl (from Julia base) to convert into footnote references. This makes the README easier to edit. This commit also includes several formatting changes.

There is no code change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## Type Aliases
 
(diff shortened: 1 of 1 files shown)


TotalVerb:
Travis failure is addressed by #300 and obviously unrelated.

PR #305, opened 2017-01-14
added iszero

stevengj:
From JuliaLang/julia#19950

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1708,6 +1708,14 @@ if !isdefined(Base, :numerator)
     export numerator, denominator
 end
 
+# julia #19950
+if !isdefined(Base, :iszero)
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -1708,6 +1708,14 @@ if !isdefined(Base, :numerator)
     export numerator, denominator
 end
 
+# julia #19950
+if !isdefined(Base, :iszero)
+    iszero(x) = x == iszero(x)
```
tkelman: this is a stack overflow isn't it?

stevengj: whoops, typo

stevengj:
Failures on nightlies seem to be due to unrelated problems with `consume`.

PR #306, opened 2017-01-14
Add .& and .|

nalimilan:
These replace element-wise & and |, which are deprecated in Julia 0.6.
This syntax is not supported on 0.4 (raises a deprecation warning).

(Failure on nightlies is unrelated.)

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1759,4 +1759,11 @@ if VERSION < v"0.5.0-dev+3669"
         scale!(similar(A, promote_op(*, eltype(A), eltype(D.diag))), D.diag, A)
 end
 
+if VERSION >= v"0.5.0-dev+5509" && VERSION < v"0.6.0-dev.1614"
+    # To work around unsupported syntax on Julia 0.4
+    include_string("export .&, .|")
+    include_string(".&(xs...) = broadcast(&, xs...)")
+    include_string(".|(xs...) = broadcast(|, xs...)")
+end
+
 end # module
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1577,3 +1577,9 @@ D = Diagonal(x)
 A = view(rand(5,5), 1:3, 1:3)
 @test D*A == Diagonal(copy(x)) * copy(A)
 @test A*D == copy(A) * Diagonal(copy(x))
+
+# julia#17623
+@static if VERSION >= v"0.5.0-dev+5509" # To work around unsupported syntax on Julia 0.4
+    @test [true, false] .& [true, true] == [true, false]
+    @test [true, false] .| [true, true] == [true, true]
+end
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * The method of `!` to negate functions (typically used as a unary operator, as in `!isinteger`) can be used in 0.5 and earlier. [#17155](https://github.com/JuliaLang/julia/pull/17155)
 
+* `.&` and `.|` are short syntax for `broadcast(&, xs...)` and `broadcast(|, xs...)` (respectively) in Julia 0.6 (only supported on Julia 0.5 and above) [#17623](https://github.com/JuliaLang/julia/pull/17623)
+
 ## Renamed functions
 
 * `pointer_to_array` and `pointer_to_string` have been replaced with `unsafe_wrap(Array, ...)` and `unsafe_wrap(String, ...)` respectively


tkelman:
can this work in an at-compat macro on 0.4?

nalimilan:
Not that I know of unfortunately. `[1] .& [2]` prints a deprecation warning on 0.4, even inside a macro.

nalimilan:
And even worse, it sometimes does not parse at all:
```julia
julia> X .& Y

WARNING: deprecated syntax "X .".
Use "X." instead.
ERROR: syntax: extra token "Y" after end of expression
```

test/runtests.jl
```diff
@@ -1577,3 +1577,11 @@ D = Diagonal(x)
 A = view(rand(5,5), 1:3, 1:3)
 @test D*A == Diagonal(copy(x)) * copy(A)
 @test A*D == copy(A) * Diagonal(copy(x))
+
+# julia#17623
+if VERSION >= v"0.5.0"
+    @eval quote # To work around unsupported syntax on Julia 0.4
```
tkelman: would at-static work here?

src/Compat.jl
```diff
@@ -1759,4 +1759,11 @@ if VERSION < v"0.5.0-dev+3669"
         scale!(similar(A, promote_op(*, eltype(A), eltype(D.diag))), D.diag, A)
 end
 
+if VERSION >= v"0.5.0" && VERSION < v"0.6.0-dev.1614"
+    # To work around unsupported syntax on Julia 0.4
+    parse("export .&, .|")
+    parse(".&(xs...) = broadcast(&, xs...)")
+    parse(".|(xs...) = broadcast(|, xs...)")
```
tkelman: these will return Exprs, did you mean include_string ?

nalimilan: Of course, I got that totally wrong. Since I made the same mistake in tests, no wonder why they passed. :-)

test/runtests.jl
nalimilan: Doesn't exist on 0.4...

tkelman: compat implements it

src/Compat.jl
tkelman: at-static would look better here too

test/runtests.jl
nalimilan: Ah, great. That works here, but not in the other file, since just parsing the syntax is an error.

tkelman: I thought that's what at-static was supposed to help with. and it's exported so shouldn't need at-compat to use

yuyichao: `@static` doesn't really help parsing error. It can avoid lowering error if something parses but raises error/depwarn during lowering or type inference.

nalimilan: In the present case it fixes the error on 0.4, though the deprecation warnings remain.

nalimilan:
Another question to address: the PR currently replaces `.&` with `broadcast(&, ...)` on 0.5, but it could equally replace it with `&`. Which is better? Cc: @stevengj

nalimilan:
Bump. This is needed to fix DataFrames on 0.6, among others.

src/Compat.jl
```diff
@@ -1759,4 +1759,11 @@ if VERSION < v"0.5.0-dev+3669"
         scale!(similar(A, promote_op(*, eltype(A), eltype(D.diag))), D.diag, A)
 end
 
+if VERSION >= v"0.5.0" && VERSION < v"0.6.0-dev.1614"
+    # To work around unsupported syntax on Julia 0.4
```
tkelman: do we know what would have made these first parse?

nalimilan: Sorry, I don't understand the question.

tkelman: it was some pr during 0.5-dev that first made these parse, right? "parse more dot operators" by stevengj I want to say?

ararslan: Yes: https://github.com/JuliaLang/julia/pull/17393

stevengj:
Would be nice if we could get CI working on nightly... I don't see a Compat issue for this?

src/Compat.jl
nalimilan: OK, I've made the check specific to that commit.

nalimilan:
AppVeyor failure due to network error.

nalimilan:
Good to go now?

src/Compat.jl
```diff
@@ -1759,4 +1759,11 @@ if VERSION < v"0.5.0-dev+3669"
         scale!(similar(A, promote_op(*, eltype(A), eltype(D.diag))), D.diag, A)
 end
 
+if VERSION >= v"0.5.0-dev+5509" && VERSION < v"0.6.0-dev.1614"
```
tkelman: how did you get the 1614 cutoff? I think that's wrong, I get 1632 at the merge commit  b0a41531fc4a0345710375adb3d7f40b6220ae91

stevengj:
@nalimilan, `&` is probably faster on 0.5, but I don't think it's critical for Compat to use the fastest possible code on older releases.

src/Compat.jl
nalimilan: 1614 corresponds to commit https://github.com/JuliaLang/julia/pull/17623/commits/f216cbb7d3eb5a070ae84e18662c9702b8ba7561, from https://github.com/JuliaLang/julia/pull/17623 for which 1632 is the merge commit. Which is the most correct?

tkelman: The merge commit. Otherwise you're counting relative to where work on the PR branched off from, and there's going to be overlap for commits from master that don't have the feature. Linear numbering isn't quite sufficient to capture "newer or older than a given commit" so we should probably move towards feature tests where possible.

nalimilan: OK, see https://github.com/JuliaLang/Compat.jl/pull/318.

Issue #297, opened 2017-01-01
2-arg vectorized trunc doesn't work in Julia 0.5


giordano:
In Julia 0.5
```julia
julia> trunc.(Int, 10rand(10))
ERROR: MethodError: no method matching size(::Type{Int64})
Closest candidates are:
  size{N}(::Any, ::Integer, ::Integer, ::Integer...) at abstractarray.jl:48
  size(::BitArray{1}) at bitarray.jl:39
  size(::BitArray{1}, ::Any) at bitarray.jl:43
  ...
 in broadcast_shape(::Type{T}, ::Array{Float64,1}) at ./broadcast.jl:31
 in broadcast_t(::Function, ::Type{Any}, ::Type{T}, ::Vararg{Any,N}) at ./broadcast.jl:213
 in broadcast(::Function, ::Type{T}, ::Array{Float64,1}) at ./broadcast.jl:230
```
Vectorized version of `trunc` is now preferred in master, after (probably) https://github.com/JuliaLang/julia/pull/19791.  Any chance to make it work in Julia 0.5 using `Compat.jl`?

TotalVerb:
See also related #291. Most of the broadcasting infrastructure would be quite difficult to backport, unfortunately.

giordano:
> See also related #291.
I know, I'm the same who reported that one ;-)

> Most of the broadcasting infrastructure would be quite difficult to backport, unfortunately.
I suspected that, but keeping compatibility in packages for both 0.5 and 0.6 is hard.  I have to choose between dropping support for 0.5 now and updating the package later.  I think I'll do the former as soon as 0.6 is out.

stevengj:
We could define a limited number of `broadcast` and `broadcast!` methods just so that people can remove deprecations.  e.g. `broadcast(f::Function, T::Type, A::AbstractArray) = broadcast(x -> f(T, A), A)` for the `trunc`, `round`, etcetera methods.

PR #307, opened 2017-01-18
Fix Compat on master by updating walkdir

andreasnoack:
Since `walkdir` returns a `Task` in 0.5 and a `Channel` in 0.6 the implementation in Compat would need to be updated. For now, I've done this by not exporting the `Compat` version such that you'd have to use `Compat.walkdir` instead of just `walkdir`.

`promote_eltype_op` is now deprecated on master so I've also deprecated the `Compat` version.

Testing of vectorized `round`, `ceil`, `floor`, and `trunc` is disabled on 0.6 since the dot syntax should be used.

Some minor syntax deprecations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.isapprox` with `nans` keyword argument [#20022](https://github.com/JuliaLang/julia/pull/20022)
 
+* `take!` method for `Task`s since some functions now return `Channel`s instead of `Task`s [#19841](https://github.com/JuliaLang/julia/pull/19841).
+
 ## Renamed functions
 
 * `pointer_to_array` and `pointer_to_string` have been replaced with `unsafe_wrap(Array, ...)` and `unsafe_wrap(String, ...)` respectively
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -194,6 +194,9 @@ if VERSION < v"0.5.0-dev+961"
         Task(_it)
     end
 end
+if VERSION < v"0.6.0-dev.2043"
+    Base.take!(t::Task) = consume(t)
+end
 
 function rewrite_show(ex)
     if isexpr(ex, :call)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -983,29 +983,29 @@ cd(dirwalk) do
     follow_symlink_vec = has_symlinks ? [true, false] : [false]
     has_symlinks && symlink(abspath("sub_dir2"), joinpath("sub_dir1", "link"))
     for follow_symlinks in follow_symlink_vec
-        task = walkdir(".", follow_symlinks=follow_symlinks)
-        root, dirs, files = consume(task)
+        chnl = walkdir(".", follow_symlinks=follow_symlinks)
+        root, dirs, files = take!(chnl)
         @test root == "."
         @test dirs == ["sub_dir1", "sub_dir2"]
         @test files == ["file1", "file2"]
 
-        root, dirs, files = consume(task)
+        root, dirs, files = take!(chnl)
         @test root == joinpath(".", "sub_dir1")
         @test dirs == (has_symlinks ? ["link", "subsub_dir1", "subsub_dir2"] : ["subsub_dir1", "subsub_dir2"])
         @test files == ["file1", "file2"]
 
-        root, dirs, files = consume(task)
+        root, dirs, files = take!(chnl)
         if follow_symlinks
             @test root == joinpath(".", "sub_dir1", "link")
             @test dirs == []
             @test files == ["file_dir2"]
-            root, dirs, files = consume(task)
+            root, dirs, files = take!(chnl)
         end
         for i=1:2
             @test root == joinpath(".", "sub_dir1", "subsub_dir$i")
             @test dirs == []
             @test files == []
-            root, dirs, files = consume(task)
+            root, dirs, files = take!(chnl)
         end
 
         @test root == joinpath(".", "sub_dir2")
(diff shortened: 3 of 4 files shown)


TotalVerb:
We shouldn't deprecate `promote_eltype_op` because that adds deprecation warnings to packages that use it on 0.5.

TotalVerb:
I have an alternative approach here: https://github.com/TotalVerb/Compat.jl/commit/8f6d272edd06f845fbab66f9d138027fa3622186

This will deprecate `promote_eltype_op` only on 0.6. Do you think that approach would work?

andreasnoack:
I'm not completely sure about what is the right solution here but `Compat` provides most recent Julia features on older Julia version. `promote_eltype_op` is no longer a Julia feature so I don't think that `Compat` should provide the feature.

TotalVerb:
There may be packages that still support 0.4 and 0.5, and have yet to upgrade to 0.6. Compat should not break these packages or add deprecation warnings to them until Compat drops support for those versions.

That this feature is deprecated in Julia 0.6 means that we should deprecate it for Julia 0.6, but not 0.5 or earlier.

src/Compat.jl
```diff
@@ -152,21 +152,21 @@ elseif VERSION < v"0.4.0-dev+6987"
     export pipeline
 end
 
-if VERSION < v"0.5.0-dev+961"
-    export walkdir
-
+if VERSION < v"0.6.0-dev.2043"
     function walkdir(root; topdown=true, follow_symlinks=false, onerror=throw)
```
TotalVerb: should document that `walkdir` should be used as `Compat.walkdir` on 0.5

src/Compat.jl
```diff
@@ -152,21 +152,21 @@ elseif VERSION < v"0.4.0-dev+6987"
     export pipeline
 end
 
-if VERSION < v"0.5.0-dev+961"
```
TotalVerb: should still export walkdir on 0.4

TotalVerb:
It would be breaking to unexport `walkdir` on 0.4. It's fine to restrict it to `Compat.walkdir` for 0.5.

andreasnoack:
I think it is flawed to support the syntax of several versions of Julia at the same time and not only the latest version but I guess we are already doing that so I'll remove the parts of this PR that touches `promote_eltype_op` and adjust the `walkdir` part.

src/Compat.jl
```diff
@@ -152,21 +152,21 @@ elseif VERSION < v"0.4.0-dev+6987"
...
-            return Task(()->())
+            # Need to return an empty closed channel to skip the current root folder
+            chnl = Channel(0)
+            close(chnl)
+            return chnl
         end
-        dirs = Array(eltype(content), 0)
-        files = Array(eltype(content), 0)
+        dirs = Array{eltype(content)}(0)
+        files = Array{eltype(content)}(0)
```
tkelman: Vector

test/runtests.jl
```diff
@@ -98,35 +98,37 @@ if VERSION < v"0.4.0-dev+1387"
...
-@test floor(Int, [1.1 1.1]) == [1 1]
-@test floor(Int, fill(1.1, 2, 3, 4)) == fill(1, 2, 3, 4)
-@test trunc(Int, 1) == 1
-@test trunc(Int, 1.1) == 1
-@test trunc(Int, [1, 1]) == [1, 1]
-@test trunc(Int, [1.1, 1.1]) == [1, 1]
-@test trunc(Int, [1 1]) == [1 1]
-@test trunc(Int, [1.1 1.1]) == [1 1]
-@test trunc(Int, fill(1.1, 2, 3, 4)) == fill(1, 2, 3, 4)
+if VERSION < v"0.6.0-"
```
tkelman: be more specific, and anything called on scalars should run everywhere. or better, do `@compat` and dot syntax everywhere

src/Compat.jl
TotalVerb: This code is copied from Base; I don't think Compat should deviate from it.

test/runtests.jl
TotalVerb: Dot syntax doesn't work on 0.4 and 0.5 yet for these functions, see #297.

src/Compat.jl
tkelman: why doesn't the base walkdir work on 0.5?

src/Compat.jl
tkelman: it should be Vector in base then

src/Compat.jl
tkelman: isn't the semantic change breaking for anyone using this on 0.4?

src/Compat.jl
TotalVerb: you can't `take!` from a `Task`, so the two walkdirs are not compatible

src/Compat.jl
TotalVerb: Yes, any code using `consume(walkdir(...))` will break.

src/Compat.jl
TotalVerb: The `Array{T}(0)` pattern is used very frequently in Base. Should that be changed in Base, then it can be changed here too.

src/Compat.jl
TotalVerb: If it's any consolation, _most_ uses of `walkdir` discard the return value. The return value is not even documented.

src/Compat.jl
tkelman: it should, and has been in many places - it's easier on the compiler to call the constructor of a fully parameterized type instead of having to infer the dimensionality

src/Compat.jl
tkelman: export an old version on 0.4, have a non exported new version

src/Compat.jl
TotalVerb: I couldn't find a Base issue so opened https://github.com/JuliaLang/julia/issues/20114.

test/runtests.jl
```diff
@@ -981,29 +983,29 @@ cd(dirwalk) do
     follow_symlink_vec = has_symlinks ? [true, false] : [false]
     has_symlinks && symlink(abspath("sub_dir2"), joinpath("sub_dir1", "link"))
     for follow_symlinks in follow_symlink_vec
-        task = walkdir(".", follow_symlinks=follow_symlinks)
-        root, dirs, files = consume(task)
+        task = Compat.walkdir(".", follow_symlinks=follow_symlinks)
```
amitmurthy: Should also rename "task", "task_error", etc to "chnl", "chnl_error" everywhere.

andreasnoack:
I've updated the PR such that it doesn't touch `promote_eltype_op `. Unfortunately, https://github.com/JuliaLang/Compat.jl/issues/315 happened meanwhile. I'm not sure yet what the right fix for that is.

andreasnoack:
I've added a fix for #315 as well. It disables the same tests as in https://github.com/JuliaLang/julia/pull/18777/files#diff-3d2d2cb368aae0105589988195f6b11aL341 .

tkelman:
> Compat should not break these packages or add deprecation warnings to them until Compat drops support for those versions.

And multiple outstanding line comments.

test/runtests.jl
tkelman: okay, then anything on scalars should run everywhere, anything vectorized on the specific julia versions where it's not deprecated

andreasnoack:
@amitmurthy I guess that `Channel`s have changed quite a bit on master. On 0.5 they don't accept a function argument. Is at all feasible to support this in Compat?

amitmurthy:
task-channel binding is not supported on 0.5

Compat's implementation of walkdir should
- create an explicit task
- close the channel before exiting - normal or on error.

I can push a commit to your branch later if you don't mind.

README.md
```diff
@@ -81,7 +81,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `foreach`, similar to `map` but when the return value is not needed ([#13744])
 
-* `walkdir`, returns an iterator that walks the directory tree of a directory ([#13707])
+* `walkdir`/`Compat.walkdir`, returns an iterator that walks the directory tree of a directory. For compatibility with Julia 0.6- `Compat.walkdir` should be used and `walkdir` should be used for compatibility with the Julia 0.5. ([#13707])
```
tkelman: "for compatibility with Julia 0.5"

test/runtests.jl
```diff
@@ -101,32 +101,35 @@ end
...
 @test trunc(Int, 1) == 1
 @test trunc(Int, 1.1) == 1
-@test trunc(Int, [1, 1]) == [1, 1]
-@test trunc(Int, [1.1, 1.1]) == [1, 1]
-@test trunc(Int, [1 1]) == [1 1]
-@test trunc(Int, [1.1 1.1]) == [1 1]
-@test trunc(Int, fill(1.1, 2, 3, 4)) == fill(1, 2, 3, 4)
+
+if VERSION < v"0.6.0-dev.1825"
+    @test round(Int, [1, 1]) == [1, 1]
```
Keno: `@compat round.(`?

src/Compat.jl
```diff
@@ -194,6 +194,9 @@ if VERSION < v"0.5.0-dev+961"
         Task(_it)
     end
 end
+if VERSION < v"0.6.0-dev.2043"
+    Base.take!(t::Task) = consume(t)
```
tkelman: should be mentioned in readme

test/runtests.jl
tkelman: ref https://github.com/JuliaLang/Compat.jl/issues/297

Keno: I see thanks. I don't care too strongly about these tests on 0.6, so I'm fine with this.

src/Compat.jl
andreasnoack: It is not mentioned in the `README.md`.

Keno:
Tests failing on nightly now.

src/Compat.jl
tkelman: not = now ?

andreasnoack:
Just realized. I think the deprecations caused `centralizedabs2fun` to be exported from `Base` so when the deprecations were deleted, `centralizedabs2fun` was no longer exported. I'm trying to come up with a fix.

Keno:
I think this is good to go. @andreasnoack do the honors?

PR #309, opened 2017-01-20
Compat.isapprox method with nans keyword argument

andreasnoack:


src/Compat.jl
```diff
@@ -947,6 +947,16 @@ if VERSION < v"0.4.0-dev+6578"
     ≉(x,y) = !(x ≈ y)
     export ≈, ≉
 end
+if VERSION < v"0.6.0-dev.2093" # Compat.isapprox to allow for NaNs
+    if VERSION >= v"0.4.0-dev+6578"
```
martinholters: We require `julia 0.4`, so this should always be true?

andreasnoack: Thanks. Then I might as well just remove the previous definition as part of this PR.

src/Compat.jl
```diff
@@ -1774,4 +1761,13 @@ if VERSION >= v"0.5.0-dev+5509" && VERSION < v"0.6.0-dev.1614"
     include_string(".|(xs...) = broadcast(|, xs...)")
 end
 
+if VERSION < v"0.6.0-dev.2093" # Compat.isapprox to allow for NaNs
+    using Base.rtoldefault
+    function isapprox(x::Number, y::Number; rtol::Real=rtoldefault(x,y), atol::Real=0, nans::Bool=false)
```
tkelman: don't want to overwrite the default when nans is not specified, right?

andreasnoack: Yes. That was the plan.

tkelman: is it possible to define a kwarg method without a default value?

ararslan: Nope, that's a syntax error saying that the kwarg needs a default value.

andreasnoack:
Can we merge this? The tests are failing because of the `wakedir` breakage.

src/Compat.jl
tkelman: Let me phrase the original question differently then. Doesn't the default-value method here overwrite the base method without the nan kwarg?

andreasnoack: I'm not importing the Base method so I wouldn't think so.

PR #317, opened 2017-02-03
at-compat for Foo{<:Bar} sugar

stevengj:
`@compat` support for https://github.com/JuliaLang/julia/pull/20414 (at least for non-nested cases that are possible to support).


stevengj:
Will merge in a couple of days if there are no objections here.

PR #304, opened 2017-01-13
Add Compat for TypeUtils

Keno:
See https://github.com/JuliaLang/julia/pull/20006

PR #319, opened 2017-02-04
Add Compat for TypeUtils: fix for julia 0.4

timholy:
Closes #304. CC @Keno.


Keno:
I'll go ahead and tag a new Compat version that includes these changes.

PR #320, opened 2017-02-07
Add isabstract(::TypeConstructor)

timholy:
CC @Keno, for https://github.com/JuliaGraphics/ColorTypes.jl/pull/68.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1783,6 +1783,8 @@ module TypeUtils
         using Base: isabstract, parameter_upper_bound, typename
     else
         isabstract(t::DataType) = t.abstract
+        isabstract(t::TypeConstructor) = isabstract(t.body)
+        isabstract(t::ANY) = false
         parameter_upper_bound(t::DataType, idx) = t.parameters[idx].ub
         typename(t::DataType) = t.name
     end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1614,8 +1614,13 @@ end
 
 # julia#20006
 abstract AbstractFoo20006
-immutable ConcreteFoo20006{T<:Int} <: AbstractFoo20006
-end
+immutable ConcreteFoo20006{T<:Int} <: AbstractFoo20006 end
+immutable ConcreteFoo20006N{T<:Int,N} <: AbstractFoo20006 end
+typealias ConcreteFoo200061{T<:Int} ConcreteFoo20006N{T,1}
 @test Compat.TypeUtils.isabstract(AbstractFoo20006)
+@test !Compat.TypeUtils.isabstract(ConcreteFoo20006)
+@test !Compat.TypeUtils.isabstract(ConcreteFoo20006N)
+@test !Compat.TypeUtils.isabstract(ConcreteFoo200061)
+@test !Compat.TypeUtils.isabstract(StridedArray)
 @test Compat.TypeUtils.parameter_upper_bound(ConcreteFoo20006, 1) == Int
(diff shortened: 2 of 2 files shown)


timholy:
On 0.6 we get this:
```julia
julia> typeof(ConcreteFoo200061)
UnionAll

julia> Base.isabstract(ConcreteFoo200061)
false
```

src/Compat.jl
```diff
@@ -1783,6 +1783,7 @@ module TypeUtils
         using Base: isabstract, parameter_upper_bound, typename
     else
         isabstract(t::DataType) = t.abstract
+        isabstract(t::TypeConstructor) = true
```
yuyichao: `isabstract` checks if the base type is abstract so this should recursively access the `body` like `unwrap_unionall`.

test/runtests.jl
```diff
@@ -1614,8 +1614,10 @@ end
 
 # julia#20006
 abstract AbstractFoo20006
-immutable ConcreteFoo20006{T<:Int} <: AbstractFoo20006
-end
+immutable ConcreteFoo20006{T<:Int} <: AbstractFoo20006 end
+immutable ConcreteFoo20006N{T<:Int,N} <: AbstractFoo20006 end
+typealias ConcreteFoo200061{T<:Int} ConcreteFoo20006N{T,1}
 @test Compat.TypeUtils.isabstract(AbstractFoo20006)
+@test Compat.TypeUtils.isabstract(ConcreteFoo200061)
```
yuyichao: Should be

```julia
@test !Compat.TypeUtils.isabstract(ConcreteFoo20006)
@test !Compat.TypeUtils.isabstract(ConcreteFoo20006N)
@test !Compat.TypeUtils.isabstract(ConcreteFoo200061)
```

Keno:
I'm in a agreement with @yuyichao

test/runtests.jl
timholy: Thanks, @yuyichao

src/Compat.jl
```diff
@@ -1783,6 +1783,7 @@ module TypeUtils
         using Base: isabstract, parameter_upper_bound, typename
     else
         isabstract(t::DataType) = t.abstract
+        isabstract(t::TypeConstructor) = isabstract(t.body)
```
yuyichao: Just wondering is it possible for the body to be neither datatype or typeconstrunctor. Maybe a Union .

timholy: What would you do, map it over the types? With `any` or `all`?
```julia
julia> using Compat.TypeUtils

julia> all(isabstract, StridedArray.body.types)
false

julia> any(isabstract, StridedArray.body.types)
true
```

yuyichao: Following `Base.isabstract` it should return `false` for anything else although I feel like it should actually return `true` in those cases.@Keno?

timholy: How about for now we don't define that method and wait to see if we need it. My guess is that few packages will use this functionality anyway.

timholy: Or perhaps this is the intended meaning of this function: `isabstract(T) == true` means it's valid to say `immutable S <: T`. Which for a `Union`, it's not.

yuyichao: I guess it agrees with the the doc

> Determine whether T was declared as an abstract type (i.e. using the  abstract keyword).

So returning `false` for anything else should be correct.

timholy: Hmm, I swear that docstring wasn't there is an earlier 0.6 version I was looking at...or this would have been a shorter conversation. Fix coming.

PR #321, opened 2017-02-09
try fix undeffed isabstract on certain julia versions

KristofferC:
Ref https://github.com/JuliaLang/julia/commit/e2cceb619d100d57ccdd06aa1fdc6b4ec6d650a5#commitcomment-20813338 cc @tkelman

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1797,7 +1797,7 @@ end
 
 module TypeUtils
     using ..Compat: @static
-    @static if isdefined(Core, :isabstract)
+    @static if isdefined(Base, :isabstract)
         using Base: isabstract, parameter_upper_bound, typename
     else
         isabstract(t::DataType) = t.abstract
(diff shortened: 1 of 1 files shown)


src/Compat.jl
```diff
@@ -1797,7 +1797,7 @@ end
 
 module TypeUtils
     using ..Compat: @static
-    @static if isdefined(Core, :UnionAll)
+    @static if isdefined(Core, :isabstract)
```
tkelman: `isabstract` isn't in Core

KristofferC:
Right you are

src/Compat.jl
KristofferC: Right you are, thanks

tkelman:
(probably should have squashed)

KristofferC:
isch, indeed.. im so just to squash being the default

PR #322, opened 2017-02-10
only add method for TypeConstructor if type exists

KristofferC:
cc @tkelman

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1801,7 +1801,7 @@ module TypeUtils
         using Base: isabstract, parameter_upper_bound, typename
     else
         isabstract(t::DataType) = t.abstract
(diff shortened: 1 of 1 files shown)


src/Compat.jl
```diff
@@ -1801,7 +1801,9 @@ module TypeUtils
         using Base: isabstract, parameter_upper_bound, typename
     else
         isabstract(t::DataType) = t.abstract
-        isabstract(t::TypeConstructor) = isabstract(t.body)
+        @static if isdefined(Base, :TypeConstructor)
+            isabstract(t::TypeConstructor) = isabstract(t.body)
```
yuyichao: Shouldn't need the static

KristofferC: removed

PR #323, opened 2017-02-10
Clean out pre-0.4 cruft

timholy:
Puts Compat on a bit of a diet. I wasn't sure whether we should deprecate old functionality (do we need `CompatCompat.jl`?), but I did so for safety. I am also not sure whether we want to run tests of deprecated functionality, but again included it for safety.


timholy:
Looks like this will trigger warnings in StatsBase (re `@ngenerated`).

timholy:
I modified the deprecation warning for `@ngenerate` to urge people to try to switch to tuples/`CartesianIndex` rather than running straight for `@generated function` (though I listed that as an option).

Among the packages installed on my machine, the other remaining consumer of `@ngenerate` is DataArrays.

tkelman:
I'd like to avoid introducing deprecation warnings on Julia 0.4, since many packages aren't making new releases there. When we drop support for 0.4 would be a good time to do this, though this likely should be one of the last packages to do that.

timholy:
Very reasonable choice. I commented out the depwarns and renamed the files `to-be-deprecated.jl`. Once this is merged you/I should file an issue to remind us to uncomment the depwarns once 0.4 support is dropped.

src/Compat.jl
```diff
@@ -752,251 +465,7 @@ end
...
-    end
-    @eval module Libdl
-        using Base.Sys: dlext, dllist, dlpath
-        export DL_LOAD_PATH, RTLD_DEEPBIND, RTLD_FIRST, RTLD_GLOBAL, RTLD_LAZY,
-               RTLD_LOCAL, RTLD_NODELETE, RTLD_NOLOAD, RTLD_NOW, dlclose, dlext,
-               dllist, dlopen, dlopen_e, dlpath, dlsym, dlsym_e, find_library
-    end
-    export Libc, Libdl
-else
-    import Base: Libc, Libdl
```
tkelman: should be kept, or `Compat.Libc` moved to to-be-deprecated

src/Compat.jl
```diff
@@ -752,251 +465,7 @@ end
...
-    Base.real{T<:Real}(::Type{T}) = T
-    Base.real{T<:Real}(::Type{Complex{T}}) = T
-end
-
-# Use as Compat.Linspace to get improved linspace in both 0.3 and 0.4
-if VERSION < v"0.4.0-dev+6110"
-    include("linspace.jl")
-    const linspace = LazyLinSpace.linspace
-else
-    const linspace = Base.linspace
```
tkelman: likewise should keep `Compat.linspace` available, possible future deprecation candidate

test/to-be-deprecated.jl
```diff
@@ -0,0 +1,61 @@
+# Once we drop 0.4 support, we want the following to return depwarns
+# info("Beginning of tests with deprecation warnings")
+
+v = 1
+d = Dict{Int,Int}()
+d[1] = 1
+@test Compat.@Dict(1 => 1) == d
+@test Compat.@Dict(1 => v) == d
+
+@test typeof(@compat(Dict(1 => 1))) == Dict{Int,Int}
```
tkelman: won't this be a no-op rather than a depwarn?

src/Compat.jl
```diff
@@ -752,251 +465,7 @@ end
...
-    isvalid(::Type{Char}, ch)         = is_valid_char(ch)
-    isvalid(::Type{ASCIIString}, str) = is_valid_ascii(str)
-    isvalid(::Type{UTF8String},  str) = is_valid_utf8(str)
-    isvalid(::Type{UTF16String}, str) = is_valid_utf16(str)
-    isvalid(::Type{UTF32String}, str) = is_valid_utf32(str)
-    export isvalid
-end
-
-if VERSION < v"0.4.0-dev+5322"
-    include("timer.jl")
```
tkelman: delete this file then i guess?

src/Compat.jl
timholy: Not necessary.
```julia
julia> Compat.Libc
Base.Libc
```

src/Compat.jl
timholy:
```julia
julia> Compat.linspace == Base.linspace
true
```

test/to-be-deprecated.jl
timholy: They won't trigger a deprecation warning, but conversely the tests aren't required since Compat doesn't do anything to `Dict` syntax anymore. No point hanging on to no-op tests.

timholy:
`timer.jl` deleted.

timholy:
Rebased. I'll merge later today unless I hear otherwise.

yuyichao:
Maybe merge after https://github.com/JuliaLang/Compat.jl/pull/326?

timholy:
That one is unquestionably of higher priority. This is just one of those PRs that shouldn't linger too long, since it's a magnet for merge conflicts.

yuyichao:
Actually I thought this removes 0.4 support which is not the case so nvm. I'll rebase this one if I want to merge #326 (probably early afternoon) before you merge this one.

tkelman:
Is (most of) this code hurting anything by being here but inactive?

yuyichao:
It shouldn't hurt anything though it should be done right after 0.3 support is dropped...

src/Compat.jl
```diff
@@ -378,41 +183,9 @@ if !isdefined(Base, :tryparse)
...
-                   (:Complex32,:complex32),
-                   (:Complex64,:complex64),
-                   (:Complex128,:complex128),
-                   (:Char,:char),
-                   (:Bool,:bool),
-                   (:Symbol,:symbol)]
-        calltypes[k] = v
-    end
-elseif VERSION < v"0.5.0-dev+3831"
+import Base.unsafe_convert
```
yuyichao: So what'll be the timeline for deprecating and removing this? (Assuming we don't want people to use `Compat.unsafe_convert` for ever, might also apply to a few other places).

I can see the argument of not giving new depwarn on 0.4 though waiting longer doesn't help since we always drop a version here only when virtually no one is using it anymore. Should we do that on the same release that drop 0.4 support assuming that's when 0.5 is still widely tested?

tkelman:
That was a while ago, and I argued against it at the time since leaving it in makes the possibility of restoring 0.3 support including (some of?) the latest features a little simpler, if a need ever arises for that. Code deletions could be reverted and people could use branches for such things, but if it's harmless to keep it...

timholy:
Proximal motivation was that I was entertaining the possibility of putting in the new `StepRangeLen` and `LinSpace`. But there was some old `linspace` Compat stuff, and I didn't want to mess with trying to potentially have two versions of that loaded.

tkelman:
So the old pre-0.4 linspace is hurting stuff by being here, I'm fine with deleting that if it's in your way.

timholy:
I should confess that I can't promise to get to this immediately. We could merge this when I'm sure I'll have time.

TotalVerb:
There has been an important change since 0.3 support was dropped: with 0.6 on the horizon, I think that the chances of ever needing to backport anything to 0.3 are now very low. So deleting the code makes sense to me.

timholy:
OK, let's do it now just so there aren't rebase conflicts in the future.

PR #325, opened 2017-02-10
Add at-compat for new type declaration syntax and fix depwarn in tests

yuyichao:


yuyichao:
Another question is if we should implement something like

```julia
@struct typename{T} <: basetype{T} begin
    field::T
end
@mutable typename2{T} <: basetype2{T} begin
    field::T
end
```

Similar to (but uglier than) the soon to be removed [`@Dict`](https://github.com/JuliaLang/Compat.jl/blob/58c52dc613fbc9e6a9943765d029f308ad79bb6c/src/Compat.jl#L76) macro etc.

README.md
```diff
@@ -71,6 +71,10 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `@compat Array{<:Real}` and similar uses of `<:T` to define a set of parameterized types ([#20414]).
   In 0.4 and 0.5, this only works for non-nested usages (e.g. you can't define `Array{<:Array{<:Real}}`).
 
+* `@compat abstract type T end` and `@compat primitive type T 8 end`
+  to declare abstract and primitive types. This only work when
```
tkelman: only works when

xref the base PR

test/runtests.jl
```diff
@@ -1353,8 +1353,10 @@ end
 end
 @test do_boundscheck() == true
 
-@test Compat.promote_eltype_op(@functorize(+), ones(2,2), 1) === Float64
-@test Compat.promote_eltype_op(@functorize(*), ones(Int, 2), zeros(Int16,2)) === Int
+if VERSION < v"0.6.0-dev.1886"
```
tkelman: what changed here?

pabloferz: `promote_eltype_op` was deprecated?

yuyichao: Right

README.md
yuyichao: Done.

test/runtests.jl
tkelman: would be worth a comment to that effect

yuyichao:
The 14 packages I've tested works well with this, with only `JLD` showing a small parser issue that isn't really caused by this macro (`type A $b end` is parsed as `type $(A, b) end`). Will merge shortly.

PR #324, opened 2017-02-10
Add broadcast for same-length tuples on 0.5

pabloferz:
Fixes #291

PR #326, opened 2017-02-12
Support new syntax for type alias with free parameters

yuyichao:


src/Compat.jl
```diff
@@ -1815,7 +1817,7 @@ end
 if VERSION < v"0.5.0-dev+3669"
     using Base: promote_op
     import Base: *
-    typealias SubMatrix{T} SubArray{T,2}
+    eval(Expr(:typealias, :(SubMatrix{T}), :(SubArray{T,2})))
```
TotalVerb: Why doesn't `@compat SubMatrix{T} = SubArray{T,2}` work here?

yuyichao: Just avoiding using `@compat` in this file in general.

TotalVerb: `@compat` is used 20 or so times in this file. Is there a reason to avoid it?

yuyichao: I just don't ever want to worry about ordering unless it's very hard to do without `@compat`.

yuyichao:
Added support for `@compat const A{T} = B{T}` since it might be useful to keep the style consistent...

tkelman:
before anyone tags a new version here, should be sure to resolve https://github.com/JuliaLang/Compat.jl/pull/324#issuecomment-279244196 in the readme

PR #329, opened 2017-02-17
Support IndexStyle, IndexLinear, IndexCartesian

timholy:
Ref https://github.com/JuliaLang/julia/pull/16378.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `@compat A{T} = B{T}` or `@compat const A{T} = B{T}` to declare type alias with free
   parameters. [#20500]. Use `const A = B` for type alias without free parameters.
 
+* `@compat Base.IndexStyle(::Type{<:MyArray}) = IndexLinear()` and `@compat Base.IndexStyle(::Type{<:MyArray}) = IndexCartesian()` to define traits for abstract arrays, replacing the former `Base.linearindexing{T<:MyArray}(::Type{T}) = Base.LinearFast()` and `Base.linearindexing{T<:MyArray}(::Type{T}) = Base.LinearSlow()`, respectively.
+
 ## Module Aliases
 
 * In 0.6, some 0.5 iterator functions have been moved to the `Base.Iterators`
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -147,6 +147,12 @@ function is_quote_symbol(ex::ANY, val::Symbol)
     return false
 end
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -101,6 +101,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bytestring` has been replaced in most cases with additional `String` construction methods; for 0.4 compatibility, the usage involves replacing `bytestring(args...)` with `Compat.String(args...)`. However, for converting a `Ptr{UInt8}` to a string, use the new `unsafe_string(...)` method to make a copy or `unsafe_wrap(String, ...)` to avoid a copy.
 
+* In 0.6, older non-exported indexing traits were renamed and exported as `IndexStyle`, `IndexLinear`, and `IndexCartesian`. Any code specializing `Base.linearindexing(::Type{T})` should switch to `Compat.IndexStyle(::Type{T})`.
```
nalimilan: The correct form is `::Type{<:T}` AFAIK.

timholy: I know, I was just being brief and didn't want to exclude notation that was valid on 0.5. I'll try to change it to be more explicit about old and new syntax.

nalimilan: Yeah, but I know I got it wrong the first time, so I would say it's worth being explicit since people are likely to copy/paste the suggested solution. You could use `@compat Compat.IndexStyle(::Type{<:T})` (or just use the old syntax).

README.md
```diff
@@ -78,6 +78,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `@compat A{T} = B{T}` or `@compat const A{T} = B{T}` to declare type alias with free
   parameters. [#20500]. Use `const A = B` for type alias without free parameters.
 
+* `@compat Compat.IndexStyle(::Type{<:MyArray}) = IndexLinear()` and `@compat Compat.IndexStyle(::Type{<:MyArray}) = IndexCartesian()` to define traits for abstract arrays, replacing the former `Base.linearindexing{T<:MyArray}(::Type{T}) = Base.LinearFast()` and `Base.linearindexing{T<:MyArray}(::Type{T}) = Base.LinearSlow()`, respectively.
```
timholy: Or should we allow people to write this as `@compat Base.IndexStyle(...)`, effectively lying about which module `IndexStyle` is defined in?

nalimilan: I would say either solution is fine. `Base.IndexStyle` might be a bit better since people will just have to remove `@compat` to get the official 0.6 syntax, but not a big deal.

timholy:
Tests pass. I have a growing collection of packages that I have PRs ready to queue, so would be great to merge this soon and tag a new release.

PR #330, opened 2017-02-24
Support new zeros/ones methods from #19635

timholy:
https://github.com/JuliaLang/julia/pull/19635. CC @jw3126.


tkelman:
add to readme

timholy:
Done

Issue #263, opened 2016-07-29
Document minimum Compat.jl versions for features


kmsquire:
When using Compat in packages, we require specifying a minimum version for Compat in the package's REQUIRE file. Finding this sometimes feels burdensome, and could be made a bit easier if the required Compat.jl version for each feature was included in the documentation.

stevengj:
I agree that this would be nice.  Unfortunately, the version number is not known at the time a feature is merged, so you'd have to go back somehow at the time the package is tagged.

I guess when the feature is merged we could add `(Compat ?)` under that feature in the README, and then have a script that replaces `Compat ?` with `Compat x.y.z` and tags a version?

ranjanan:
This may help:  To find the minimum Compat version, click on the blame icon in the README, find your function and click on the corresponding commit ID. This would then load a page that shows the minimum tagged version after this particular commit ID.

KristofferC:
We should run a whitespace checker then: https://github.com/JuliaLang/Compat.jl/commit/12c36d71182ff15fe1c13bc51ee0384603493f59

martinholters:
To repeat what I said in #670: The idea is to add the version information in the commit that sets the version number in `Project.toml`. I just wonder how we prevent forgetting to do so...

PR #331, opened 2017-03-02
Add a findversion script that determines the correct tag.

helgee:
Since I get the `Compat` version wrong every time I decided to automate it. If you pass the function one of the code snippets from the README it will return the correct tag, e.g.

```julia
julia> using Compat
julia> Compat.findversion("@static")
"v0.10.0"
```
Git needs to be installed.

Do we want this?

EDIT: This is now a script:

```bash
$ ./bin/version.jl @static
v0.10.0
```

## ToDo

* [X] Docs


yuyichao:
I don't think this should be included in the test since it requires the git repo and not just the package. It probably also shouldn't be in the package and should rather be a script since there's no reason to load it and basically no one should use this at runtime.

helgee:
Where to put the script?

yuyichao:
There was a `bin/version.sh` that's moved to base so this can be in `bin/`

src/version.jl
```diff
@@ -0,0 +1,48 @@
...
+    cd(repo) do
+        lines = readlines(`git blame --line-porcelain -L$idx,$idx -- $file`)
+    end
+    split(lines[1])[1]
+end
+
+function gettag(repo, commit)
+    lines = Vector{String}()
+    cd(repo) do
+        lines = readlines(`git tag --contains $commit`)
```
kmsquire: Wondering if this is possible through the LibGit2 interface?

tkelman:
Neat idea. It'll get fooled by whitespace changes in the readme, but so does manual blame, you just have to know to click through those

src/version.jl
helgee: Possibly, but right now it still has to work with v0.4.

kmsquire: Ok, makes sense.  Either way, this is nice!

kmsquire:
I'd say that, when merged, this is sufficient to address #263.

helgee:
Turned it into a script and documented it in the README.

README.md
```diff
@@ -279,19 +279,29 @@ specify the right lower bound.
...
-which can be found in the top-right corner.
+* Copy the code snippet that describes the feature and pass it to the `bin/version.jl`
+script which will return the correct version.
 
-* Now find your fix, and then find the corresponding commit ID of that fix on the
-left-hand side. Click on the commit ID. This navigates to a page which recorded
-that particular commit.
+```sh
+bash $ ./bin/version.jl @static
+v0.12.0
```
tkelman: so yeah, like I said, this is fooled by readme changes and wrong. can we blacklist specific commits and have the blame keep going?

helgee:
@tkelman I think I fixed it. This uses `git log` with pickaxe search now instead of `blame`.

yuyichao:
Given how this package is tested, wouldn't it be easier to just add a version to each readme line?

helgee:
Bump.

giordano:
@tkelman Whitespace changes can be ignored with `-w` option to `git blame`, but I've seen sometimes rephrasing changes that makes even more annoying blaming the README.  A reference in the README to the version in which a feature was introduced, as suggested by @yuyichao, or a proper ChangeLog, like `NEWS.md` file in Julia, would be much more useful than an automated script that can be fooled in a number of ways, IMHO.

helgee:
In its current form the script does not suffer from these problems since it uses the right tools for the job, i.e. `git log` instead of `git blame`.

giordano:
Don't take me wrong, I welcome any tool that could make determining the `Compat` version needed easier than it's now, but I'm not sure that an automatic tool is fool-proof.  For example,
```
./version.jl Compat.String
v0.10.0
```
but `Compat.String` was introduced in 0.7.15: https://github.com/JuliaLang/Compat.jl/commit/4c8c62d275fd11273124fcf4d76f2513eafa0ccc.

giordano:
I believe that in `gettag` you should convert each line to a `VersionNumber`, sort the vector and then take the first tag, otherwise you're going to pick-up often the wrong tag, i.e. something like:
```julia
sort(VersionNumber.(lines)))[1]
```
instead of `lines[1]`.

c42f:
This will be great. I've just made a Compat version mistake yet again (tagging PlyIO) and was looking to solve this exact same issue when I came across this PR.

I don't think the version numbers created by this script need to be exact lower bounds - an approximation will do, as long as it's newer than the true lower bound.  This is because almost no packages have upper bounds on Compat.

@giordano - it seems right that versions should be sorted by `VersionNumber`.  Can we get that small fix in and just merge this?  Tony's got his eagle eyes on METADATA, but I'm sure he's got better things to do than find my versioning mistakes ;-)

giordano:
The (fixed) script gives a completely wrong version for the feature

* `broadcast` is supported on tuples of the same lengths on 0.5. (#16986)

```
$ ./version.jl broadcast
0.7.15
```

Should have been `0.18.0` (see https://github.com/JuliaLang/Compat.jl/commit/b42521674f3c0482e784b3e93ad39bbacdd41b60).  Probably the word "broadcast" was used before in the README, but this makes the script useless in this case.

This script is definitely useful (but not complete nor always accurate) and is better than current situation, but I still believe that keeping a simple ChangeLog from Compat developers (or simply adding a version to each line, as suggested above) would greatly alleviate the pain for the users and solve all problems.

c42f:
Ah, that's not so good.  Maybe something like #263 is the only true solution.

PR #334, opened 2017-03-06
Add Compat.readline

### LINKED ISSUE #333: Compat.readline
It would be useful to get `Compat.readline` which is `readline` on 0.6, and `chomp!(readline(...))` on 0.5.

TotalVerb:
Fix #333.

`eachline` and `readlines` are significantly harder to do properly. We would either need to implement a lazy iterator map, or port the functions to `Compat`. I think we should add this first and leave those until someone wants them.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -597,8 +597,8 @@ if VERSION < v"0.5.0-dev+2228"
     Base.read(filename::AbstractString, args...) = open(io->read(io, args...), filename)
     Base.read!(filename::AbstractString, a) = open(io->read!(io, a), filename)
     Base.readuntil(filename::AbstractString, args...) = open(io->readuntil(io, args...), filename)
-    Base.readline(filename::AbstractString) = open(readline, filename)
-    Base.readlines(filename::AbstractString) = open(readlines, filename)
+    Base.readline(filename::AbstractString) = open(Base.readline, filename)
+    Base.readlines(filename::AbstractString) = open(Base.readlines, filename)
     Base.readavailable(s::IOStream) = read!(s, @compat Vector{UInt8}(nb_available(s)))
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -1411,6 +1411,18 @@ if VERSION < v"0.6.0-dev.2840"
     IndexStyle(args...) = Base.linearindexing(args...)
 end
 
+# https://github.com/JuliaLang/julia/pull/20203
+if VERSION < v"0.6.0-dev.2283"
+    # not exported
+    function readline(s::IO=STDIN; chomp::Bool=true)
+        if chomp
+            Base.chomp(Base.readline(s))
```
nalimilan: Can `Base.chomp!` be used? Else you don't need the namespace prefix.

TotalVerb: The namespace prefix is needed here because `chomp` is shadowed by the function argument :wink:.

I think `Base.chomp!` can probably be used safely here. I will make that change.

src/Compat.jl
```diff
@@ -1411,6 +1411,18 @@ if VERSION < v"0.6.0-dev.2840"
     IndexStyle(args...) = Base.linearindexing(args...)
 end
 
+# https://github.com/JuliaLang/julia/pull/20203
+if VERSION < v"0.6.0-dev.2283"
+    # not exported
```
nalimilan: Wait, don't you also need `const readline = Base.readline` for later versions? I don't understand how the tests passed on nightlies without it...

TotalVerb: No, it's not needed. `readline` is exported from Base so Compat.readline aliases it unless redefined.

nalimilan: Ah, funny!

PR #338, opened 2017-03-23
Type definition macros for 0.5-1.0

iamed2:
This allows you to write type definitions that work on 0.5 and 1.0. Not sure if this is actually desired but I wrote it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,8 @@ Currently, the `@compat` macro supports the following syntaxes:
   vectorized function have migrated. These macros will be dropped when the
   support for `0.6` is dropped from `Compat`.
 
+* `@struct begin ... end` is equivalent to `immutable ... end` in Julia 0.5 and `struct ... end` in Julia 0.6+. `@mutable_struct begin ... end` is equivalent to `type ... end` in Julia 0.5 and `mutable struct ... end` in Julia 0.6+.
+
 ## Other changes
 
 * `remotecall`, `remotecall_fetch`, `remotecall_wait`, and `remote_do` have the function to be executed remotely as the first argument in Julia 0.5. Loading `Compat` defines the same methods in older versions of Julia ([#13338])
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1423,6 +1423,46 @@ if VERSION < v"0.6.0-dev.2283"
     end
 end
 
+# https://github.com/JuliaLang/julia/pull/20418
+export @immutable_struct
+macro immutable_struct(ex)
+    _struct(ex, false)
+end
+
+export @mutable_struct
+macro mutable_struct(ex)
+    _struct(ex, true)
+end
+
+function _struct(ex, mutable::Bool)
+    function macro_name()
+        "@$(mutable ? "" : "im")mutable_struct"
+    end
+
+    if ex.head !== :block
+        throw(ArgumentError("$(macro_name()) can only be applied to block expressions"))
+    end
+
+    type_name = ex.args[2]
+
+    if !isa(type_name, Symbol) && type_name.head !== :curly && type_name.head !== :(<:)
+        throw(ArgumentError(
+            "The first expression after `$(macro_name()) begin` must be the type name"
+        ))
+    end
+
+    return esc(Expr(
+        :type,
+        mutable,
+        type_name,
+        Expr(
+            :block,
+            ex.args[1],
+            ex.args[3:end]...,
+        ),
+    ))
+end
+
 include("to-be-deprecated.jl")
 
 end # module Compat
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1768,4 +1768,28 @@ end
 @test Compat.readline(IOBuffer("x\n"), chomp=true) == "x"
 @test Compat.readline(IOBuffer("x\n"), chomp=false) == "x\n"
 
+# PR 20418
+@mutable_struct begin MutableTestType{T<:Real} <: Real
+    x::T
+
+    MutableTestType() = new(one(T))
+end
+@test isdefined(:MutableTestType)
+@test isa(MutableTestType, Type)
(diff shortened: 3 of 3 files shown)


tkelman:
Since these aren't actually deprecated yet, the recommended approach will be to change them whenever you drop support for 0.5, and not need Compat at all. Will people need to support for all of 0.5, 0.6, and the dev version of 1.0?

iamed2:
I honestly don't know. I wrote this in response to deprecation warnings from using the old syntax in 0.6...which don't exist (anymore?). This _would_ be useful if anyone wants to support those three versions but I'm not sure those people exist. I don't personally expect to use this but I thought I'd contribute it since it works and there might be someone who needs it sometime. 

Feel free to close and keep around until someone asks :)

martinholters:
FWIW, I usually code against master, but also keep compatibility with the oldest version still allowed into METADATA, so I would probably become a user of these.

martinholters:
After updating Julia to past the deprecation switch being flipped, it occurred to me that Compat itself might benefit from these macros, no? At least if we want to maintain that Compat master is free of Julia master deprecations.

TotalVerb:
I think we should drop support for 0.4 and 0.5 before those deprecations are flipped. The 0.7 and 1.0 dev cycles are anticipated to be quite short.

martinholters:
So we should bump REQUIRE from julia 0.4 to julia 0.6 now? That seems drastic. Or did you mean to drop 0.5 support when the deprecations are replaced with errors?

TotalVerb:
Are `type` and `immutable` deprecated already? I didn't realize that. If so, I would support these macros as a temporary stopgap. But I think we ought to drop 0.4 support, at least, sooner rather than later.

tkelman:
Compat itself has almost no types in it. I think I saw someone get away with using manually constructed Exprs since I don't think the AST heads have changed?

tkelman:
I don't think people who aren't working about 0.7 should worry about it at all just yet. These stopgaps mean you have a temporary fix just to keep things working on 0.5, and Compat has to carry them permanently, or eventually deprecate and remove them.

TotalVerb:
I guess a more conservative solution is to just fix the deprecations in Compat manually (e.g. `include_string` or manual AST manipulation and `eval`) without exporting these macros.

martinholters:
This is what triggered me to come back here:
```julia
WARNING: deprecated syntax "type" at ~/.julia/v0.7/Compat/src/Compat.jl:819.
Use "mutable struct" instead.

WARNING: deprecated syntax "immutable" at ~/.julia/v0.7/Compat/src/Compat.jl:1275.
Use "struct" instead.

WARNING: deprecated syntax "immutable" at ~/.julia/v0.7/Compat/src/Compat.jl:1321.
Use "struct" instead.

WARNING: deprecated syntax "immutable" at ~/.julia/v0.7/Compat/src/Compat.jl:1356.
Use "struct" instead.

WARNING: deprecated syntax "type" at ~/.julia/v0.7/Compat/src/Compat.jl:1376.
Use "mutable struct" instead.
```
Not urgent, but at the latest when 0.7 goes alpha, it would be nice not to have these anymore, but we might still want to support 0.5 until after 0.7 is released. And any solution we cook up in Compat could also be interesting for other packages. But I see the point of this not really belonging into Compat. Creating another package just for this also seems like overkill. Baseline: I don't kniow what I really want myself, but I'm unhappy about the current situation :smile:

tkelman:
those lines might be addressable with a `@static` or two, if they're all behind VERSION conditionals?

TotalVerb:
The deprecations are thrown in the parser and not at lowering, so that wouldn't work. We'd need `include_string`.

src/Compat.jl
```diff
@@ -1423,6 +1423,46 @@ if VERSION < v"0.6.0-dev.2283"
     end
 end
 
+# https://github.com/JuliaLang/julia/pull/20418
+export @immutable_struct
```
timholy: Shouldn't this just be called `@struct`?

iamed2: In 0.7:

```julia
julia> macro struct() end
ERROR: syntax: invalid name "struct"
```

yuyichao:
```julia
julia> @eval macro $(Symbol("struct"))()
           x
       end
@struct (macro with 1 method)
```

Or

```julia
julia> macro _struct()
           x
       end
@_struct (macro with 1 method)

julia> @eval const $(Symbol("@struct")) = $(Symbol("@_struct"))
@_struct (macro with 1 method)
```

README.md
```diff
@@ -214,6 +214,8 @@ Currently, the `@compat` macro supports the following syntaxes:
   vectorized function have migrated. These macros will be dropped when the
   support for `0.6` is dropped from `Compat`.
 
+* `@struct begin ... end` is equivalent to `immutable ... end` in Julia 0.5 and `struct ... end` in Julia 0.6+. `@mutable_struct begin ... end` is equivalent to `type ... end` in Julia 0.5 and `mutable struct ... end` in Julia 0.6+.
+
```
omus: Reference `@struct` doesn't currently match the macro name

src/Compat.jl
```diff
@@ -1423,6 +1423,46 @@ if VERSION < v"0.6.0-dev.2283"
...
+
+export @mutable_struct
+macro mutable_struct(ex)
+    _struct(ex, true)
+end
+
+function _struct(ex, mutable::Bool)
+    function macro_name()
+        "@$(mutable ? "" : "im")mutable_struct"
+    end
```
omus: Why is this a function?

iamed2: It's called in two different places but only needs to be called if there's an error

omus: Ah, ok. I just thought it should be saved to a variable but this makes sense if you only need it when there is an error

PR #346, opened 2017-04-12
Backport bswap method for complex numbers.

traktofon:
It was added to julia in [PR#21346](https://github.com/JuliaLang/julia/pull/21346), commit [b7836db](https://github.com/JuliaLang/julia/commit/b7836db4527e07a07d2415a8780294e9067a1f97).

Originally I had this method in my own package [FortranFiles.jl](https://github.com/traktofon/FortranFiles.jl) (which also supports julia 0.5), but because this defines a Base method for a Base datatype, @tkelman suggested it might go into Base, and if accepted there, also into Compat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `unsafe_trunc(::Type{<:Integer}, ::Integer)` is supported on 0.5. ([#18629])
 
+* `bswap` is supported for `Complex` arguments. ([#21346])
+
 ## Renamed functions
 
 * `pointer_to_array` and `pointer_to_string` have been replaced with `unsafe_wrap(Array, ...)` and `unsafe_wrap(String, ...)` respectively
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -161,6 +161,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `unsafe_trunc(::Type{<:Integer}, ::Integer)` is supported on 0.5. ([#18629])
 
+* `bswap` is supported for `Complex` arguments. ([#21346])
```
TotalVerb: maybe add "on 0.5 and earlier"?

traktofon: Since Compat now requires julia 0.4, I feel (but not very strongly) that specifying this is redundant.

traktofon:
Ok, I have integrated @TotalVerb's suggestion.

README.md
traktofon: Ok, I have integrated @TotalVerb's suggestion.

TotalVerb: The reason for this is that some features are available only on 0.5, and not on 0.4 (see a good chunk of the `broadcast` stuff), so we should be clear that this works on both.

stevengj: I think that the default should be that every Compat feature is supported on every Julia version supported by the REQUIRES file (i.e. v0.4+ currently).  A note should only be necessary when a given feature is *not* supported in 0.4.

PR #348, opened 2017-04-13
Add Compat.StringVector

TotalVerb:
This is just `Vector{UInt8}` on 0.5 and below, and `Base.StringVector` on 0.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
-* `Compat.StringVector` is supported on 0.5 and below. ([#19449])
+* `Compat.StringVector` is supported on 0.5 and below. On 0.6 and later, it aliases `Base.StringVector`. ([#19449])
 
 ## Renamed functions
 
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
+* `Compat.StringVector` is supported on 0.5 and below. ([#19449])
+
 ## Renamed functions
 
 * `pointer_to_array` and `pointer_to_string` have been replaced with `unsafe_wrap(Array, ...)` and `unsafe_wrap(String, ...)` respectively
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
-* `Compat.StringVector` is supported on 0.5 and below. On 0.6 and later, it aliases `Base.StringVector`. ([#19449])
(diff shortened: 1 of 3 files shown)


stevengj:
See also JuliaLang/julia#19945 ... it would be nice to decide whether we're actually going to document/export this in Base.

TotalVerb:
I think we should have this regardless of whether it will be documented/exported in `Base`. We had `promote_eltype_op` in `Compat`, after all. If the API changes, we can deprecate it as we've done with `promote_eltype_op` here.

TotalVerb:
The 0.4 test is failing because `String(x)` calls `bytestring(x)` which does not properly alias array `x`, instead making a copy. We should have `String(x)` call `UTF8String(x)` instead, which is the 0.4 equivalent of `String`. Any objections to making this change?

tkelman:
changing the return type like that seems risky. could this instead wait until Compat drops 0.4 support?

TotalVerb:
We can also disable that particular test on all versions of 0.4 that have `String(x::Vector{UInt8})` call `bytestring`, which should be fine.

README.md
```diff
@@ -163,6 +163,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
+* `Compat.StringVector` is supported on 0.5 and below. ([#19449])
```
stevengj: Should mention that it is an alias for `Base.StringVector` on 0.6

src/Compat.jl
```diff
@@ -1464,6 +1464,13 @@ if VERSION < v"0.6.0-pre.beta.102"
     Base.bswap(z::Complex) = Complex(bswap(real(z)), bswap(imag(z)))
 end
 
+# https://github.com/JuliaLang/julia/pull/19449
+if VERSION < v"0.6.0-dev.1988"
+    StringVector = Vector{UInt8}
```
stevengj: should probably be
```
StringVector(n::Integer) = Vector{UInt8}(n)
```
doing `StringVector = Vector{UInt8}` has the drawback that (a) it is not `const` and (b) it accepts multiple arguments, unlike `Base.StringVector` in 0.6.

TotalVerb:
Thanks, I addressed the comments.

README.md
```diff
@@ -163,6 +163,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
+* `Compat.StringVector` is supported on 0.5 and below. On 0.6 and later, it aliases `Base.StringVector`. ([#19449])
```
stevengj: Maybe it should suggest using `Compat.UTF8String(Compat.StringVector(n))` to create strings from arrays without making copies?  This way the code would work on 0.4 too.  The test could be modified accordingly.

TotalVerb: 0.4 is on its way out, and many users might not know/care about `Compat.UTF8String`. But you are right about the test. I'll change that.

TotalVerb:
The Mac failure is concerning. Does anyone have an idea what's going on?

README.md
stevengj: Until Compat drops support for 0.4, we should explain how to use it "on 0.5 and below".

TotalVerb: I've updated the README.

TotalVerb:
Although the Travis failure seems to have disappeared overnight, the [segfault](https://travis-ci.org/JuliaLang/Compat.jl/jobs/222271243) is still concerning. Should I raise an issue in Base?

TotalVerb:
@stevengj Do you think the current version is good to go?

PR #350, opened 2017-04-18
Provide 0.6 implementation of collect for previous releases

timholy:


timholy:
Cross-links: https://github.com/JuliaLang/julia/pull/21257, https://github.com/JuliaIO/ImageMagick.jl/pull/85.

src/Compat.jl
```diff
@@ -1471,6 +1471,20 @@ else
     using Base: StringVector
 end
 
+# https://github.com/JuliaLang/julia/pull/21257
+if v"0.5.0" <= VERSION < v"0.6.0-pre.beta.28"
```
tkelman: why the 0.5 lower bound?

timholy: OffsetArrays in its modern form only exists for 0.5 and above, though it's nice that older versions are installable on 0.4 and thus avoid giving us an error just from test/REQUIRE. Moreover, I'm not sure the necessary `copy!` method was available on 0.4. In general any support that 0.4 would have for arrays with nontraditional indices would be a fluke, and no one should even try to use them; this test is only relevant on 0.5 and above.

tkelman: It wasn't the release that made this method relevant though...

timholy: Do we really want to be encouraging 0.5 pre-releases at *this* point?

timholy: Note that in the README I specified that `Compat.collect` is for 0.5 and above.

tkelman: encourage no, but Compat's job is to work on as much of its supported version range as possible

tkelman: "even on 0.5" isn't the clearest way of phrasing that

timholy: #351

PR #351, opened 2017-04-19
Support `collect` for rc-releases of Julia 0.5

timholy:
CC @tkelman


tkelman:
thanks. it was that late?

timholy:
Yes.

PR #352, opened 2017-04-27
add invokelatest from Julia 0.6

stevengj:
This adds `Compat.invokelatest` for the `Base.invokelatest` function introduced in https://github.com/JuliaLang/julia/pull/19784

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
 * `Compat.invokelatest` is equivalent to `Base.invokelatest` in Julia 0.6,
-  but works Julia 0.4+, and allows you to guarantee that a function call
+  but works in Julia 0.4+, and allows you to guarantee that a function call
   invokes the latest version of a function ([#19784]).
 
 * `Compat.StringVector` is supported on 0.5 and below. On 0.6 and later, it aliases `Base.StringVector`. This function allocates a `Vector{UInt8}` whose data can be made into a `String` in constant time; that is, without copying. On 0.5 and later, use `String(...)` with the vector allocated by `StringVector` as an argument to create a string without copying. Note that if 0.4 support is needed, `Compat.UTF8String(...)` should be used instead. ([#19449])
--- a/README.md
+++ b/README.md
@@ -165,6 +165,10 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -165,6 +165,10 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `bswap` is supported for `Complex` arguments on 0.5 and below. ([#21346])
 
+* `Compat.invokelatest` is equivalent to `Base.invokelatest` in Julia 0.6,
+  but works Julia 0.4+, and allows you to guarantee that a function call
```
martinholters: "works _on_ Julia 0.4+" (or "in"?)

src/Compat.jl
```diff
@@ -1471,6 +1471,13 @@ else
     using Base: StringVector
 end
 
+# https://github.com/JuliaLang/julia/pull/19784
+if isdefined(Base, :invokelatest)
+    import Base.invokelatest
+else
+    invokelatest(f, args...) = eval(Expr(:call, f, map(QuoteNode, args)...))
```
martinholters: While this surely is the correct definition for `invokelatest`, prior to JuliaLang/julia#17057 just `f(args...)` would often also work and be more performant. I'd surely have use for a function that does just `f(args...)` in older Julia and `invokelatest` (or `eval(...)`) in newer Julia. Should we include this in Compat?

stevengj: @martinholters, that's not true.  Prior to that PR, you might get some random obsolete version of a function.  See, for example JuliaLang/julia#7884 ... we had to add something like in IJulia long before 17057.

martinholters: I know. But I use `eval` to create anonymous functions for which there cannot be obsolete versions. But that might be a very specific use case of mine.

stevengj: The point is, this seems to be about the simplest correct definition of `invokelatest` on earlier  Julia versions.   Your suggestion above would not work in general, hence it is not appropriate for Compat.

README.md
stevengj: whoops, fixed.

stevengj:
Travis failure on 0.4 OSX seems unrelated:
```
The command "curl -A "$CURL_USER_AGENT" -s -L -o julia.dmg 'https://s3.amazonaws.com/julialang/bin/osx/x64/0.4/julia-0.4-latest-osx10.7+.dmg'" failed and exited with 56 during .
```

src/Compat.jl
martinholters: Sorry for the misunderstanding: I didn't want to propose replacing what you have done, merely suggest to consider the simplified definition in addition. But admittedly, it doesn't really fit into Compat and I have no API proposal, so let's forget about it.

stevengj:
Okay to merge?

PR #355, opened 2017-05-10
Compat flag for adding source location argument to macros

vtjnash:
This is needed for JuliaLang/julia#21746 to pass CI, so I can't have an exact commit number and instead am providing this boolean flag for packages to use as a feature test.


tkelman:
should be mentioned in the readme, and tested once there's a version number to fill in

vtjnash:
failure appears to be unrelated

PR #358, opened 2017-05-14
Test on 0.6 too

TotalVerb:
It would be wise to test this package on 0.6 also, especially with it nearing release.


tkelman:
yep lgtm

PR #359, opened 2017-05-16
Changes invokelatest eval to run in current_module()

JobJob:
`invokelatest` runs in module Compat on julia 0.5, so this:
```
import Compat.invokelatest
invokelatest(current_module)
```
returns `Compat` in julia 0.5, but on 0.6-rc1.0 (which has `Base.invokelatest` defined) it returns `Main`.

After this PR they both return `Main`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1848,6 +1848,7 @@ let foo() = begin
     end
     @test foo() == 2
 end
+@test Compat.invokelatest(current_module) === current_module()
 
 # PR 21378
 let
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1475,7 +1475,7 @@ end
 if isdefined(Base, :invokelatest)
     import Base.invokelatest
 else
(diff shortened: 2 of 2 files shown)


test/runtests.jl
```diff
@@ -1870,6 +1870,8 @@ let
     @test_throws MethodError Dates.Month(1) < Dates.Day(1)
 end
 
+@test Compat.invokelatest(current_module) != Compat
```
yuyichao: `=== current_module()`?

JobJob: yeah, that's better

stevengj:
doh, my fault, sorry

tkelman:
worth a patch release if you're in a hurry to use this in ijulia

stevengj:
Okay, will tag a patch release.

stevengj:
Tagged 0.25.1 (JuliaLang/METADATA.jl#9334)

PR #362, opened 2017-05-17
Fix deprecation warnings in source

TotalVerb:
sprinkle some `include_string` to fix deprecation warnings in source

(there are still depwarns in test)

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -816,6 +816,7 @@ if !isdefined(Base, :normalize)
 end
 
 if !isdefined(Base, :AsyncCondition)
+    include_string("""
     type AsyncCondition
         cond::Condition
         handle::Ptr{Void}
@@ -844,6 +845,7 @@ if !isdefined(Base, :AsyncCondition)
             this
         end
     end
+    """)
 
     Base.wait(c::AsyncCondition) = wait(c.cond)
 else
@@ -1272,9 +1274,11 @@ if VERSION < v"0.6.0-dev.1024"
 
         # julia #14805
         if VERSION < v"0.5.0-dev+3256"
+            include_string("""
             immutable Flatten{I}
                 it::I
             end
+            """)
 
             flatten(itr) = Flatten(itr)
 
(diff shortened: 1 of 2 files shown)


martinholters:
I've pushed the alternative using manual `Expr` construction for the tests so that we can get a feel for it. (For most of these, `include_string` wouldn't work because they happen unconditionally, but we probably could tolerate the deprecations in the tests, so this might not be needed.)

martinholters:
Oh no, the `:(ConcreteFoo20006{T<:Int} <: AbstractFoo20006)` part doesn't work on 0.4 because it parses as an `Expr(:comparison)`, but 0.4 requires an `Expr(:<:)` for the type definition. I see these simple solutions:
* Drop my commit completely, live with the deprecation warnings in the tests until 0.5 support is dropped, switch to new syntax then.
* Drop my commit for now, live with the deprecation warnings in the tests until 0.4 support is dropped, reinstate my commit then.
* Only undo the two offending changes of my commit for now, live with these two deprecation warnings until 0.4 support is dropped, reinstate these changes then.

I tend to prefer second or third option. Working out a complete fix with a version dependency sounds like overkill with dropping 0.4 support to be expected more or less shortly. Opinions?

martinholters:
Just went with option three for now to (hopefully) get this back to a working (i.e. passing tests) state.

martinholters:
Derp. Let's see how many I missed this time.

TotalVerb:
I would suggest an explicit Expr(:(<:)) call rather than a halfway solution. By the way, could you squash the commits so that we can merge these changes as separate commits?

TotalVerb:
I went ahead and manually constructed `Expr`s for the remaining deprecation warnings.

TotalVerb:
I fixed the remaining deprecation warnings and squashed into three commits (the minimum necessary to preserve authorship information). Does this look good to you?

martinholters:
Yes, makes sense. I somehow was a bit confused and thought we would need to distinguish the Julia version, which is non-sense of course, the `Expr(:<:)` works everywhere.

PR #363, opened 2017-06-07
Support atsign-__MODULE__

timholy:
https://github.com/JuliaLang/julia/pull/22064. CC @vtjnash.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1481,6 +1481,20 @@ else
     using Base: StringVector
 end
 
+# https://github.com/JuliaLang/julia/pull/22064
+if !isdefined(Base, Symbol("@__MODULE__"))
+    export @__MODULE__
+    macro __MODULE__()
+        return current_module()
+    end
+    Base.expand(mod::Module, x::ANY) = eval(mod, :(expand($(QuoteNode(x)))))
+    Base.macroexpand(mod::Module, x::ANY) = eval(mod, :(macroexpand($(QuoteNode(x)))))
+    Base.include_string(mod::Module, code::String, fname::String) =
+        eval(mod, :(include_string($code, $fname)))
+    Base.include_string(mod::Module, code::AbstractString, fname::AbstractString="string") =
+        eval(mod, :(include_string($code, $fname)))
+end
+
 # https://github.com/JuliaLang/julia/pull/19784
 if isdefined(Base, :invokelatest)
     import Base.invokelatest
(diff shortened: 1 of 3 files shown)


timholy:
Probably also worth CCing @JobJob just to make sure I didn't break his `invokelatest` stuff.

src/Compat.jl
```diff
@@ -1481,11 +1481,26 @@ else
     using Base: StringVector
 end
 
+# https://github.com/JuliaLang/julia/pull/22064
+if !isdefined(Base, Symbol("@__MODULE__"))
+    export @__MODULE__
+    macro __MODULE__()
+        esc(:(current_module()))
```
vtjnash: should be just `return current_module()` (if it had needed the function call, you also wouldn't have wanted the `esc` either)

src/Compat.jl
```diff
@@ -1481,11 +1481,26 @@ else
...
+    end
+    Base.binding_module(mod::Module, s::Symbol) = Base.binding_module(s)
+    Base.expand(mod::Module, x::ANY) = expand(x)
```
vtjnash: all of these functions should take the form `eval(mod, :(expand($(QuoteNode(x))))`

src/Compat.jl
```diff
@@ -1481,11 +1481,26 @@ else
...
+    Base.include_string(mod::Module, code::AbstractString, fname::AbstractString="string") =
+        include_string(code, fname)
+end
+
 # https://github.com/JuliaLang/julia/pull/19784
 if isdefined(Base, :invokelatest)
     import Base.invokelatest
 else
-    invokelatest(f, args...) = eval(current_module(), Expr(:call, f, map(QuoteNode, args)...))
+    invokelatest(f, args...) = eval(@__MODULE__, Expr(:call, f, map(QuoteNode, args)...))
```
vtjnash: invokelatest was introduced first, so this change is unnecessary

timholy:
Thanks, @vtjnash!

test/runtests.jl
```diff
@@ -1860,7 +1862,8 @@ let foo() = begin
     end
     @test foo() == 2
 end
-@test Compat.invokelatest(current_module) === current_module()
+cm359() = @__MODULE__
+@test Compat.invokelatest(cm359) === @__MODULE__
```
timholy: Will emit a depwarn on 0.7. But perhaps worth it?

JobJob: Just while testing yeah? Anyway, I doubt it will ever make a difference, and if you, stevengj and vtjnash are happy, I think we're probably pretty ok 😄

musm:
Would it be possible to also add support for `https://github.com/JuliaLang/julia/pull/22064/files#diff-4cbc8eb916240b8a07540daa4316d1c5R233`
i.e. `include` accepting a module in the first argument

vtjnash:
`include` doesn't take a module as the first argument. Perhaps you meant `Base.include`, which is not exported anymore, but does?

musm:
Yes sorry I meant `Base.include`

PR #371, opened 2017-06-21
WIP: Dropping 0.4

rofinn:
Initial work at dropping 0.4, but I'm guessing there is other 0.4 specific code that wasn't in a version check (e.g., `!isdefined(...)`). I'm guessing we also need to properly deprecate any Compat calls being removed to point to the appropriate Base method?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.4
     - 0.5
     - 0.6
     - nightly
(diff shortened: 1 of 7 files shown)


stevengj:
In the future, rather than closing a PR and creating a new one, just do a forced push to your branch to replace it with a new version in the same PR.

rofinn:
Thanks, yeah, I wasn't sure if force pushing was appropriate for PRs.

PR #375, opened 2017-07-04
WIP: Added StepRangeLen

### LINKED ISSUE #374: Support StepRangeLen

mauro3:
Fixes #374


timholy:
Maybe...the problem is that StepRangeLen fixes a whole host of issues with FloatRange. It might be better to copy https://github.com/JuliaLang/julia/pull/18777 into Compat...

mauro3:
Briefly scanning that PR suggests that it would be about 500 LOC to be added to Compat.  Is that not excessive?  Alternatively, I could add a warning to the README.md that there might be subtle differences?

mauro3:
Migrating away from `FloatRange`s in my package, I also had to patch some more bits:

```julia
if VERSION < v"0.6.0-dev.2376"
    const StepRangeLen = FloatRange
    @compat const SRangeL{F} = StepRangeLen{F}
else
    @compat const SRangeL{F} = StepRangeLen{F,Base.TwicePrecision{F},Base.TwicePrecision{F}}
end

# this is then ok:
type A{F}
    r::SRangeL{F} # replacing r::FloatRange{F}
end
```

PR #377, opened 2017-07-13
Add atsign-compat support for `CartesianRange`

timholy:
This won't pass on nightly because there's some issue regarding `qr` (CC @andreasnoack) that I don't have time to debug. I have verified that if you comment those tests out, this passes on (my) current master.


andreasnoack:
The problem is the `::Val{x}` -> `Val(x)` change in https://github.com/JuliaLang/julia/pull/22475

timholy:
Since the failure is unrelated, I'll merge.

PR #379, opened 2017-07-14
JuliaLang/julia#21197: accessing Cmd as an array of strings

stevengj:
This backports JuliaLang/julia#21197.


stevengj:
AppVeyor error on nightly seems unrelated:
```
while loading C:\Users\appveyor\.julia\v0.7\Compat\test\runtests.jl, in expression starting on line 142
ERROR: LoadError: MethodError: no method matching qr(::Array{Float64,2}, ::Type{Val{false}})
```

stevengj:
curl error on macos travis

tkelman:
Heads up to @mauro3, releasing this actually broke the tests of SimpleTraits.jl https://github.com/mauro3/SimpleTraits.jl/blob/b3e5395ce2c186b17fbee1f9dc1c9b9ce5b8b3b0/test/base-traits.jl#L33

PR #380, opened 2017-07-15
Add Sys.is<os>

### LINKED ISSUE #378: Support 0.7-style Sys.is<os> rather than is_<os>
I'm not sure how best to do this, since it involves a Base module that as far as I know has been in every version, unlike Iterators. For reference, this was introduced in 0.7.0-DEV.914.

ararslan:
Fixes #378.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -138,11 +138,11 @@ let x = rand(rng, Int64, 3,4)
     @test x == rand(rng, Int64, (3,4))
 end
 
-extrapath = is_windows() ? joinpath(JULIA_HOME,"..","Git","usr","bin")*";" : ""
+extrapath = Compat.Sys.iswindows() ? joinpath(JULIA_HOME,"..","Git","usr","bin")*";" : ""
 @compat withenv("PATH" => extrapath * ENV["PATH"]) do
     cmd1 = pipeline(`echo hello`, `sort`)
     cmd2 = pipeline(`true`, `true`)
-    if is_windows()
+    if Compat.Sys.iswindows()
         try # use busybox-w32
             success(`busybox`)
             cmd1 = pipeline(`busybox echo hello`, `busybox sort`)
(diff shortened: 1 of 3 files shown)


ararslan:
Sorry @yuyichao, I wasn't sufficiently diligent in my initial iteration here. You may want to have another look to ensure this still meets expectations.

ararslan:
Gah, never mind, it's still wrong for 0.5 and 0.6...

ararslan:
Alright, I think things are good now. The failure on nightly is unrelated.

test/runtests.jl
```diff
@@ -1242,6 +1242,14 @@ if VERSION < v"0.5.0-dev+4267"
...
     @test Compat.KERNEL == Sys.KERNEL
+    for os in [:apple, :bsd, :linux, :unix, :windows]
+        from_base = if VERSION >= v"0.7.0-DEV.914"
+            Expr(:., Expr(:., :Base, Base.Meta.quot(:Sys)), Base.Meta.quot(Symbol("is", os)))
+        else # VERSION >= v"0.5.0-dev+4267"
+            Expr(:., :Base, Base.Meta.quot(Symbol("is_", os)))
+        end
+        @eval @test Compat.Sys.$(Symbol("is", os))() == $from_base()
+    end
 end
```
Sacha0: Do I understand correctly that the objective is to support `Sys.is$(os)` from 0.5 on only? If so I will lgtm :).

ararslan: Yes, though 0.4 support will soon be dropped here anyway.

tkelman:
Should have actually read the logs, this is giving `WARNING: imported binding for Sys overwritten in module Compat` on 0.5

tkelman:
0.6 as well

tkelman:
This probably should eval into `Base.Sys` to avoid that warning

ararslan:
I had initially suggested evaling in #378. I can change it to that.

PR #381, opened 2017-07-19
Add Val(x)

ararslan:
Introduced in https://github.com/JuliaLang/julia/pull/22475


tkelman:
needs readme note

ararslan:
Good call, thanks.

PR #383, opened 2017-07-20
Move the include to avoid warning from Sys

ararslan:
Ref https://github.com/JuliaLang/Compat.jl/pull/380#issuecomment-316772204

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -645,22 +645,23 @@ if VERSION < v"0.7.0-DEV.1041"
     chol(J::UniformScaling, args...) = UniformScaling(chol(J.λ, args...))
 end
 
-include("deprecated.jl")
-
 # https://github.com/JuliaLang/julia/pull/21746
 const macros_have_sourceloc = VERSION >= v"0.7-" && length(:(@test).args) == 2
 
 # https://github.com/JuliaLang/julia/pull/22182
 module Sys
+    const KERNEL = Base.Sys.KERNEL
     if VERSION < v"0.7.0-DEV.914"
-        isapple(k::Symbol=Base.Sys.KERNEL)   = k in (:Darwin, :Apple)
-        isbsd(k::Symbol=Base.Sys.KERNEL)     = isapple(k) || k in (:FreeBSD, :OpenBSD, :NetBSD, :DragonFly)
-        islinux(k::Symbol=Base.Sys.KERNEL)   = k == :Linux
-        isunix(k::Symbol=Base.Sys.KERNEL)    = isbsd(k) || islinux(k)
-        iswindows(k::Symbol=Base.Sys.KERNEL) = k in (:Windows, :NT)
+        isapple(k::Symbol=KERNEL)   = k in (:Darwin, :Apple)
+        isbsd(k::Symbol=KERNEL)     = isapple(k) || k in (:FreeBSD, :OpenBSD, :NetBSD, :DragonFly)
+        islinux(k::Symbol=KERNEL)   = k == :Linux
+        isunix(k::Symbol=KERNEL)    = isbsd(k) || islinux(k)
+        iswindows(k::Symbol=KERNEL) = k in (:Windows, :NT)
     else
         import Base.Sys: isapple, isbsd, islinux, isunix, iswindows
     end
 end
+
+include("deprecated.jl")
 
 end # module Compat


src/Compat.jl
```diff
@@ -651,15 +651,13 @@ include("deprecated.jl")
 const macros_have_sourceloc = VERSION >= v"0.7-" && length(:(@test).args) == 2
 
 # https://github.com/JuliaLang/julia/pull/22182
-module Sys
```
yuyichao: Just move this before the include should be fine.

tkelman: which include?

yuyichao:
Either
```diff
diff --git a/src/Compat.jl b/src/Compat.jl
index 5679e6c..63c2a09 100644
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -645,8 +645,6 @@ if VERSION < v"0.7.0-DEV.1041"
     chol(J::UniformScaling, args...) = UniformScaling(chol(J.λ, args...))
 end

-include("deprecated.jl")
-
 # https://github.com/JuliaLang/julia/pull/21746
 const macros_have_sourceloc = VERSION >= v"0.7-" && length(:(@test).args) == 2

@@ -663,4 +661,6 @@ module Sys
     end
 end

+include("deprecated.jl")
+
 end # module Compat
diff --git a/src/deprecated.jl b/src/deprecated.jl
index 36b54ec..3d1263a 100644
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -157,11 +157,11 @@ macro functorize(f)
 end

 if VERSION >= v"0.6.0"
-    Base.@deprecate_binding KERNEL Sys.KERNEL
+    Base.@deprecate_binding KERNEL Base.Sys.KERNEL
     Base.@deprecate_binding UTF8String Core.String
     Base.@deprecate_binding ASCIIString Core.String
 else
-    const KERNEL = Sys.KERNEL
+    const KERNEL = Base.Sys.KERNEL
     const UTF8String = Core.String
     const ASCIIString = Core.String
 end
```

or

```diff
diff --git a/src/Compat.jl b/src/Compat.jl
index 5679e6c..c0c82d4 100644
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -645,22 +645,23 @@ if VERSION < v"0.7.0-DEV.1041"
     chol(J::UniformScaling, args...) = UniformScaling(chol(J.λ, args...))
 end

-include("deprecated.jl")
-
 # https://github.com/JuliaLang/julia/pull/21746
 const macros_have_sourceloc = VERSION >= v"0.7-" && length(:(@test).args) == 2

 # https://github.com/JuliaLang/julia/pull/22182
 module Sys
+    const KERNEL = Base.Sys.KERNEL
     if VERSION < v"0.7.0-DEV.914"
-        isapple(k::Symbol=Base.Sys.KERNEL)   = k in (:Darwin, :Apple)
-        isbsd(k::Symbol=Base.Sys.KERNEL)     = isapple(k) || k in (:FreeBSD, :OpenBSD, :NetBSD, :DragonFly)
-        islinux(k::Symbol=Base.Sys.KERNEL)   = k == :Linux
-        isunix(k::Symbol=Base.Sys.KERNEL)    = isbsd(k) || islinux(k)
-        iswindows(k::Symbol=Base.Sys.KERNEL) = k in (:Windows, :NT)
+        isapple(k::Symbol=KERNEL)   = k in (:Darwin, :Apple)
+        isbsd(k::Symbol=KERNEL)     = isapple(k) || k in (:FreeBSD, :OpenBSD, :NetBSD, :DragonFly)
+        islinux(k::Symbol=KERNEL)   = k == :Linux
+        isunix(k::Symbol=KERNEL)    = isbsd(k) || islinux(k)
+        iswindows(k::Symbol=KERNEL) = k in (:Windows, :NT)
     else
         import Base.Sys: isapple, isbsd, islinux, isunix, iswindows
     end
 end

+include("deprecated.jl")
+
 end # module Compat
```

ararslan:
That avoids the warning that Tony mentioned in https://github.com/JuliaLang/Compat.jl/pull/380#issuecomment-316592478? I don't see why it would, am I missing something?

tkelman:
Ah, guess it has to do with whether or not the binding `Sys` has been resolved to be the Base one yet

ararslan:
Okay, went with @yuyichao's second diff.

tkelman:
If it fixes the warning, works for me

ararslan:
Yes it does. Unfortunate IMO that the name conflict between Base and Compat exists. For example, the thinking behind adding `Sys.iswindows()` was that the `Sys` prefix added context. But with this,
```julia
julia> using Compat.Sys

julia> Sys.iswindows()
WARNING: both Sys and Base export "Sys"; uses of it in module Main must be qualified
ERROR: UndefVarError: Sys not defined
```
You either need `Compat.Sys.iswindows()` or `using Compat.Sys: iswindows`.

ararslan:
@yuyichao I assume you're opposed to `eval`ing into Base.Sys. Can you elaborate on why?

yuyichao:
> `Compat.Sys.iswindows()`

Yes, that's what you should do.

> I assume you're opposed to evaling into Base.Sys.

Strongly.

> Can you elaborate on why?

`eval`ing into a different module is about as bad as you can do for precompilation.

ararslan:
> evaling into a different module is about as bad as you can do for precompilation.

Ah, thanks, good to know. I don't know much about about the intricacies of precompilation.

ararslan:
Travis error is unrelated.

Issue #384, opened 2017-07-22
Add support for at-nospecialize


ararslan:
I started working on this but haven't quite nailed it down yet. Here's my start if anyone else wants to have a go:
```diff
diff --git a/src/Compat.jl b/src/Compat.jl
index c0c82d4..3007fab 100644
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -662,6 +662,27 @@ module Sys
     end
 end

+if VERSION < v"0.7.0-DEV.969"
+    # If any type annotations are provided, we'll just ignore them for now
+    macro nospecialize(ex)
+        if isa(ex, Symbol)                                    # @nospecialize(x)
+            Expr(:(::), quot(ex), :ANY)
+        elseif ex.head == :(=) && isa(ex.args[1], Symbol)     # @nospecialize(x=1)
+            Expr(:(=), Expr(:(::), quot(ex.args[1]), :ANY), quot(ex.args[2]))
+        elseif ex.head == :(=) && ex.args[1].head == :(::)    # @nospecialize(x::T=1)
+            Expr(:(=), Expr(:(::), quot(ex.args[1].args[1]), :ANY), quot(ex.args[1].args[2]))
+        elseif ex.head == :(...) && isa(ex.args[1], Symbol)   # @nospecialize(x...)
+            Expr(:(...), Expr(:(::), quot(ex.args[1]), :ANY))
+        elseif ex.head == :(...) && ex.args[1].head == :(::)  # @nospecialize(x::T...)
+            Expr(:(...), Expr(:(::), quot(ex.args[1].args[1]), :ANY))
+        else
+            error("improper use of @nospecialize")
+        end
+    end
+else
+    import Base: @nospecialize
+end
+
 include("deprecated.jl")

 end # module Compat
```
It basically just takes the `@nospecialize` annotation for a function argument and turns it into an `::ANY` annotation on the argument.

The quoting here isn't working quite right, since, for example, the returned expression from `@nospecialize(x)` is actually `:x::Main.ANY` as the above stands. And we can't use `esc` here instead, since that produces `Main.x::Main.ANY`, which is wrong in the context of specifying a function signature.

PR #372, opened 2017-06-21
Dropping 0.4

rofinn:
I'll leave updating the test cases till after this has been reviewed to minimize breaking existing functionality.

Fixes #361

PR #385, opened 2017-07-23
Drop 0.4 and 0.7 fixes

yuyichao:
* Drop remaining 0.4 support code
* Deprecate remaining 0.4 Compat bindings
* Remove tests for 0.4 Compat
* Do not run deprecated tests on 0.6
* Recover one test that's incorrectly moved to deprecated (StringVector)
* Add compat for at-nospecialize with tests
* Add compat for `read(..., String)` with tests

This covers all the changes since it'll be really annoy to fix the above issues one by one......................

Fix https://github.com/JuliaLang/Compat.jl/issues/384


yuyichao:
Also, there were a lot of 0.4 (and even 0.3!!!) compat code that was not removed previously so I added a rough version for all the rest that do not branch on version check. Also added some for the tests....

test/runtests.jl
```diff
@@ -1917,6 +549,24 @@ if VERSION < v"0.7.0-DEV.880"
...
+# @nospecialize
+# 0.7
+no_specialize(@nospecialize(x)) = sin(1)
+no_specialize(@nospecialize(x::Integer)) = sin(2)
+@test no_specialize(1.0) == sin(1)
+@test no_specialize(1) == sin(2)
+
+@test read(IOBuffer("aaaa"), String) == "aaaa"
+@test contains(read(@__FILE__, String), "read(@__FILE__, String)")
+@test read(`$(Base.julia_cmd()) -e "println(:aaaa)"`, String) == "aaaa\n"
```
tkelman: --startup-file=no

tkelman:
Please don't delete so many tests. If it's not deprecated, leave it.

ararslan:
What tests for non-deprecated functionality are being deleted here?

tkelman:
most of the deletions in test/

yuyichao:
> If it's not deprecated, leave it.

The deleted tests are all unrelated to `Compat`. Keeping them would be like running the `Base` test suite again here.

tkelman:
It's independent coverage of things and makes sure we don't break them.

yuyichao:
> It's independent coverage of things and makes sure we don't break them.

AFAICT they are all covered in base tests (in fact many of them are copied from base) so they are not independent coverage. These are additional runs but if you really want to catch that it's much better to setup a service for it. The test here are to test `Compat.jl`, not `Base`.

I also don't see why these test can help us catching breakage. For old releases, the test here might run more frequently than the base CI, but I don't see why running the tests here can catch anything. They increase CI time and the only reason I've seen a test fail on old version is for unrelated reasons (like binary download stops working). Adding tests won't help those.

For nightly, it'll run strictly less frequent than the base CI so it's pretty useless for catching base breakage. The tests here are also very primitive (they are all testing if the difference between versions are patched by the package and already always assume that the correct behavior is given by base). Examples of this including testing name existing and assuming they have the correct definition, testing against base implementation on old versions to make sure the new name has identical behavior. Tests like this are exactly what should be in the Compat test (it's testing the package) and should be removed when the corresponding Compat def is removed. It also basically means that the only breakage caught by these tests are intentional breakage in base and we'll need to fix the test here, instead of fixing base. Whenever a test has this kind of behavior, it's usually a signal that the test should not be there.

tkelman:
> should be removed when the corresponding Compat def is removed

I disagree with that. If it was once implemented as a feature here, then it should still be tested even if it's now a no-op, as people are still likely using it as if it isn't a no-op.

src/deprecated.jl
```diff
@@ -143,6 +143,15 @@ end
 using .CompatCartesian
 export @ngenerate, @nsplat
 
+function primarytype(@nospecialize(t))
```
tkelman: why is this in this file if it isn't deprecated?

yuyichao: What isn't deprecated? It's an internal function used only in deprecated code.

yuyichao:
> If it was once implemented as a feature here, then it should still be tested even if it's now a no-op, as people are still likely using it as if it isn't a no-op.

It was implemented as a patch on old versions and it was never implemented on 0.5+. It's not changed to no-op but simply removed and people never use it from Compat.

tkelman:
People did get things from Compat on older versions. I don't see the benefit of deleting huge numbers of tests, they're harmless and arguably beneficial to leave in. It's not like they take all that long to run.

yuyichao:
They are not harmless. They do take time to run and they are causing issues due to new changes in Base. I believe there's more harm to leave the in than the benefits since I can remember 4-5 times where an old test has to be disabled on master due to base changes but not a single time where a test breakage here indicates some unexpected base change.

yuyichao:
> People did get things from Compat on older versions.

older versions -> those are unsupported now.

tkelman:
People have been writing code against these features though, as if they come from Compat.

Delete or disable them when they cause problems. They aren't causing a problem right now, so please don't delete them yet.

yuyichao:
> Delete or disable them when they cause problems.

That's exactly what useless tests are.

tkelman:
I disagree. It doesn't hurt anything to run them.

yuyichao:
A test is used to catch issues that cause it to fail. If when the test fail the fix is to remove the test then the test is useless.

And also, when you move one feature from one packge to another, you move the tests with it. We are basically moving the "features" (not really since they are never an Compat feature on 0.5+) from this package to Base so the tests should be moved to base as well (as was done a long time ago when the correspondign base features were added.).

yuyichao:
> It doesn't hurt anything to run them.

Well, not run them. It hurts (waste a lot of time) to maintain them.

tkelman:
It takes less time to incrementally fix the handful that cause problems over time than the substantial review burden right now of deleting hundreds of lines of tests to verify that you aren't deleting anything that shouldn't be. The default with tests is that more is better and it's easier to leave them alone.

yuyichao:
> It takes less time to incrementally fix the handful that cause problems over time than the substantial review burden right now of deleting hundreds of lines of tests to verify that you aren't deleting anything that shouldn't be.

I disagree. I'll certain choose to review a PR like this than to keep maintaining old tests overtime, which I did and it's a waste of time. The waste of time was necessary when we still support the old versions but not when we don't. The loc removed in tests here isn't really more than https://github.com/JuliaLang/Compat.jl/pull/372 (or at least what should have been in it since it misses many 0.3 and 0.4 compat defs). So I don't think it's harder than that to verify that the removal are correct. FWIW, I'm not a huge fan of dropping 0.4 support but I do think that when we do, the code and the tests for them should be remove together and that'll actually make it easier.

> The default with tests is that more is better and it's easier to leave them alone.

That's certainly not true. Bad and useless tests make things worse. Again, tests are there to catch breakage. But if the expectation of the tests are that their failure is always a signal that the test themselves needs to be fixed or removed than they don't make anything better and they make things worse by increasing the maintenance effort, exactly the opposite what they should do.

yuyichao:
I totally agree that this needs to be checked and that's the point of PR review. For checking myself and to make reviewing easier. I've created a gist for all the tests that are removed (not including moved ones) here https://gist.github.com/yuyichao/f7243e25250b54e55bbd5d102489d264. The file should pass on 0.5 and 0.6 without any error or warnings. It should be easy to verify if the file is an accurate list of tests removed and if the few compat defs are needed are tested in current tests. `Sys.*`, `take!(::IOBuffer)` in `runtests.jl`, `*String` aliases in `deprecated.jl`. I do think this can be used in the future as a required reviewing procedure when we remove def and tests for old compat definitions.

tkelman:
I'm still in the process of reviewing this, would like a bit more time if you're not in a hurry.

#372 intentionally didn't remove tests, it removed code under src that was dead not tests that would now unconditionally be checking Base behavior. The Compat test cases are if anything a bit less likely than a normal equivalent amount of code to hit breakage on nightly Julia (since they're things that have changed at least once before in recent memory).

yuyichao:
> The Compat test cases are if anything a bit less likely than a normal equivalent amount of code to hit breakage on nightly Julia (since they're things that have changed at least once before in recent memory).

I really hope it's true and we stop making minor changes just to raise depwarns. Things that have recently (up to ~1yr) been changed twice includes some linear qrfact function signatures, `takebuf_string` on `IOBuffer`, and `is_<os>`s. These are just three cases where the exact thing Compat defined changes in base. And there are more depwarns caused by other changes that happens to be used in the test. So keeping more tests will cause more breackage/depwarn in Compat tests and will need more effort to maintain than necessary.

tkelman:
The things where Base changes twice and Compat has to handle 2-3 different possibilities are actually the ones where we'd need to be the most careful about how we do things here in a minimally breaking way...

yuyichao:
Right, those tests are **NOT** removed. It's just the easiest to remember cases of compat test being broken due to base changes. Their use in other tests needs to be fixed (many of which are for old compat that are now removed in this PR) and their test for old behaviors are conditioned on old julia versions, which are either removed or moved to deprecated.

stevengj:
+1 for removing tests for features that Compat no longer provides.  The point of those tests is to test the *Compat* implementation, not to test the *Base* implementation.

Keeping them just seems like it invites maintenance headaches.

musm:
Any chance this can be merged soon? Looking forward to the `readstring` compat fix here

tkelman:
Still looking over this, should finish over the weekend

musm:
Thanks, no strong rush however.

ararslan:
Bump

yuyichao:
I don't see what's time consuming about reviewing given the gist linked above https://github.com/JuliaLang/Compat.jl/pull/385#issuecomment-317255249 since there's a clear rule of when the code can be removed (i.e. if they can run without Compat.jl being loaded) so merge.

tkelman:
sorry had been looking st other things instead of this

src/Compat.jl
```diff
@@ -145,18 +50,13 @@ function _compat(ex::Expr)
...
-            return ex
-        elseif VERSION < v"0.6.0-dev.2575" #20414
+        if VERSION < v"0.6.0-dev.2575" #20414
             ex = Expr(:curly, map(a -> isexpr(a, :call, 2) && a.args[1] == :(<:) ?
                                   :($TypeVar($(QuoteNode(gensym(:T))), $(a.args[2]), false)) :
                                   isexpr(a, :call, 2) && a.args[1] == :(>:) ?
                                   :($TypeVar($(QuoteNode(gensym(:T))), $(a.args[2]), $Any, false)) : a,
                                   ex.args)...)
         end
-    elseif ex.head === :macrocall
```
tkelman: why isn't this needed any more?

src/Compat.jl
```diff
@@ -240,98 +140,7 @@ macro compat(ex...)
...
-function primarytype(t::ANY)
-    tn = t.name
-    if isdefined(tn, :primary)
-        return tn.primary
-    else
-        return tn.wrapper
-    end
-end
-
-if !isdefined(Base, :Threads)
```
tkelman: this is still needed for a no-threads build, please put it back

src/deprecated.jl
```diff
@@ -160,13 +169,22 @@ end
     Base.@deprecate_binding KERNEL Sys.KERNEL
     Base.@deprecate_binding UTF8String Core.String
     Base.@deprecate_binding ASCIIString Core.String
+    Base.@deprecate_binding unsafe_convert Base.unsafe_convert
+    Base.@deprecate_binding remote_do Base.remote_do
```
tkelman: also remotecall, remotecall_fetch, remotecall_wait

src/deprecated.jl
```diff
@@ -160,13 +169,22 @@ end
...
+    Base.@deprecate_binding Filesystem Base.Filesystem
+    Base.@deprecate_binding AsyncCondition Base.AsyncCondition
+    Base.@deprecate_binding promote_eltype_op Base.promote_eltype_op
+    @eval Base.@deprecate_binding $(Symbol("@irrational")) Base.$(Symbol("@irrational"))
+    @eval Base.@deprecate_binding $(Symbol("@blasfunc")) Base.LinAlg.BLAS.$(Symbol("@blasfunc"))
 else
```
tkelman: missing srand, rand, rand!

test/runtests.jl
```diff
@@ -1,917 +1,13 @@
...
-end
-
-# The constructors for some linear algebra stuff changed to take Val{x}
-# instead of Type{Val{x}}
-const valtrue = VERSION < v"0.7.0-DEV.843" ? Val{true} : Val(true)
-const valfalse = VERSION < v"0.7.0-DEV.843" ? Val{false} : Val(false)
-
-# qr, qrfact, qrfact!
-let A = [1.0 2.0; 3.0 4.0]
-    Q, R = qr(A, valfalse)
```
tkelman: Compat should be backporting the 0.7 version of this

src/Compat.jl
yuyichao: `f` isn't used, It's a piece of code that the last PR failed to remove.

src/Compat.jl
yuyichao: No it's not. The module is always available.

src/deprecated.jl
yuyichao: No those are exported symbols.

src/deprecated.jl
yuyichao: Same as above.

test/runtests.jl
yuyichao: It isn't currently. If it was backported then yes, a similar test should be added in that pr.

tkelman: then better to leave this if it's going to need to be added back

yuyichao: No. We should keep tests that's testing what's in the package, not what could have been in or what was in the package. This procedure makes it much more clear what is being tested and what it's testing.

Also, no one have asked for such a feature so it's not 100% certain if one will be made. If none was made, it will make it much harder, like what I have to do this time, to figure out why the test is being here and at the same time harder for the next version dropping PR to be reviewed. If one was added, explicitly adding it back later also make it much more clear to the reviewer what's actually being tested, without having to dig through the history to figure out if the test wasn't being incorrectly re-proposed.

It's not like the test can't be found anymore, they can be retried easily. It's also not like if the test is kept here it can be directly used, it'll have to be modified to actually test what the PR want.

yuyichao: There was many cases where deprecated API's are not added in Compat.

In any case, I'm just listing both cases and as I mentioned above, even if you are 100% sure the Compat support will be added, without the context in this thread, it will actually take more review effort and won't really help the one writing the PR. For the PR author, it'll be easier to just check the Base tests changed in the Val Base PR https://github.com/JuliaLang/julia/pull/22475, instead of going through the old tests and figure out if they can be modified for this purpose. For reviewer, it'll also be much harder to figure out if the repurpose of the test is correct. It won't be if we haven't drop 0.4 support yet so the reviewer needs to check that whatever this test is doing is already supported by base julia 0.5, doing that check in a batch with a single script containing deleted tests is the whole point why reviewing the deletion here is easier than deleting one at a time as we go.

src/Compat.jl
```diff
@@ -4,126 +4,31 @@ module Compat
...
-    if isexpr(ex, :call)
-        Expr(:call, rewrite_show(ex.args[1]), ex.args[2:end]...)
-    elseif isexpr(ex, :curly)
-        Expr(:curly, rewrite_show(ex.args[1]), ex.args[2:end]...)
-    else
-        :(Base.writemime)
-    end
-end
-
-function rewrite_dict(ex)
```
tkelman: Lint.jl was using this

yuyichao: AFAICT it was just using this as test code. Deprecating this won't even help.

PR #394, opened 2017-08-09
Fix vectorization deprecation macros for Julia 0.7

### LINKED ISSUE #392: dep_vectorize_xarg fails on master
E.g.
```julia
julia> Compat.@dep_vectorize_1arg Float64 log
ERROR: TypeError: in @deprecate, in typeassert, expected Symbol, got Expr
```

ararslan:
Fixes #392


ararslan:
Come to think of it, why isn't Compat using the definition from Base for versions where it's defined?

yuyichao:
No, because that'll make removing a deprecation from Base a breaking change.

PR #396, opened 2017-08-10
Remove compat 0.7 pkg eval logo

musm:
Currently it doesn't exist so it does not render at all when viewing the readme. This removes the logo.


musm:
bump

musm:
Is this important or should I just close this?

PR #399, opened 2017-08-24
Backport reshape and ntuple Val() APIs to 0.5 and 0.6

mbauman:
The functionality is there on both 0.5 and 0.6; we just need to convert Val(N) to Val{N}.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Cmd` elements can be accessed as if the `Cmd` were an array of strings for 0.6 and below ([#21197]).
 
-* `Val(x)` constructs `Val{x}()`. The `reshape` and `ntuple` APIs are extended to support `Val{x}()` arguments on 0.6 and below. ([#22475])
+* `Val(x)` constructs `Val{x}()`. ([#22475])
+
+* The `reshape` and `ntuple` APIs are extended to support `Val{x}()` arguments on 0.6 and below.
 
 * `chol` and `chol!` for `UniformScalings` ([#22633]).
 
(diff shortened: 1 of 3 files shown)


mbauman:
Ok, I also added support for ntuple and updated the readme.

src/Compat.jl
```diff
@@ -461,6 +461,11 @@ end
 if VERSION < v"0.7.0-DEV.843"
     import Base: Val
     (::Type{Val})(x) = (Base.@_pure_meta; Val{x}())
+    # Also add methods for Val(x) that were previously Val{x}
+    import Base: reshape
+    reshape{N}(parent::AbstractArray, ndims::Val{N}) = reshape(parent, Val{N})
```
ararslan: Shouldn't this be `reshape(parent, Val(N))` since `Val{N}` is the type and giving `reshape` the `Val` type is deprecated in favor of giving it the `Val` object?

mbauman: This is adding a method to support `reshape(parent, Val(N))` on 0.6 and 0.5.  Those versions have this functionality, but it's defined in terms of `reshape(parent, Val{N})`.  So we just covert the new API to speak with the old API.

ararslan: Ah, gotcha. Sorry for the noise.

README.md
```diff
@@ -149,7 +149,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Cmd` elements can be accessed as if the `Cmd` were an array of strings for 0.6 and below ([#21197]).
 
-* `Val(x)` constructs `Val{x}()`. ([#22475])
+* `Val(x)` constructs `Val{x}()`. The `reshape` and `ntuple` APIs are extended to support `Val{x}()` arguments on 0.6 and below. ([#22475])
```
tkelman: could you put these on separate bullets?

mbauman:
Thanks Alex!

PR #402, opened 2017-09-29
Add at-isdefined

ararslan:
Note that this is **NOT** functionally equivalent to `@isdefined` on 0.7; that isn't possible on earlier versions. Instead this defines `@isdefined` to wrap `isdefined`, which can only check for global variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,11 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@nospecialize` has been added ([#22666]).
 
+* Julia 0.7 introduced `@isdefined` to replace `isdefined`, which only works for globals
+  ([#22281]). It isn't possible to achieve the 0.7 behavior in earlier versions of Julia,
+  but an `@isdefined` macro is defined here to avoid deprecation warnings when supporting
+  both 0.7 and earlier versions.
+
 ## Other changes
 
 * On versions of Julia that do not contain a Base.Threads module, Compat defines a Threads module containing a no-op `@threads` macro.
@@ -318,6 +323,7 @@ includes this fix. Find the minimum version from there.
 [#21709]: https://github.com/JuliaLang/julia/issues/21709
 [#22064]: https://github.com/JuliaLang/julia/issues/22064
 [#22182]: https://github.com/JuliaLang/julia/issues/22182
+[#22281]: https://github.com/JuliaLang/julia/issues/22281
 [#22350]: https://github.com/JuliaLang/julia/issues/22350
 [#22475]: https://github.com/JuliaLang/julia/issues/22475
 [#22629]: https://github.com/JuliaLang/julia/issues/22629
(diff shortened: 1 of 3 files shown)


ararslan:
CI failures on macOS are unrelated. Is this something we want or would it be too misleading to have it only support a very limited subset of its 0.7 functionality on 0.6?

yuyichao:
The subset where this behaves the same as `@isdefined` can be written as `isdefined(@__MODULE__, ...)` so I don't think we should have this.

ararslan:
Okay.

nalimilan:
I think we should definitely have this because when updating code from 0.6 using `isdefined`, the most direct replacement is `@isdefined`, not `isdefined(@__MODULE__, ...)`. Indeed that's what the deprecation suggests. Otherwise, the deprecation message should be changed.

yuyichao:
No that's unrelated. `@isdefined` is the closest replacement on 0.7. It is not on 0.6 and adding a definition on 0.6 that does not do the same thing at all does not help.

nalimilan:
If `isdefined(@__MODULE__, ...)` is a good replacement, why couldn't we define `@isdefined` to use it on 0.6?

yuyichao:
> couldn't we define `@isdefined` to use it on 0.6?

Because it's a good replacement of `isdefined`. It is **not** a good implementation (replacement) of `@isdefined` so such a definition will do completely different things on 0.6 and 0.7.

yuyichao:
In another word, if you really don't want to type `@__MODULE__` defining `Compat.@isdefined_in_module` to do `isdefined(@__MODULE__, ...)` is certainly fine. The only thing that's wrong is to call it `@isdefined`.

Also, defining `@isdefined` on 0.6 is kind of possible. I don't think it can be done with any performance though (you need to `try catch`). Such a definition is mostly useless due to the performance issue but if the function of it is really useful for someone (i.e. checking isdefined for local var and not global var) it can be added to `Compat` and called `@isdefined`.

yuyichao:
And yet another way you can call it `@isdefined` is to make sure it either return the correct result or raise an error on old versions, it's not as good as something that is always correct but at least it won't lead to bugs/much confusion.

In this case, this can be done (conservatively) if you make sure the macro is only used in global scope. I'm not sure the best way to do this but here's one of the very old tricks I've used before.

```julia
julia> macro assert_global()
           mod = isdefined(Base, Symbol("@__MODULE__")) ? __module__ : current_module()
           @gensym dummy
           quote
               $(esc(dummy)) = nothing
               if !isdefined($mod, $(QuoteNode(dummy)))
                   error("Not in global scope")
               end
           end
       end
@assert_global (macro with 1 method)

julia> @assert_global

julia> let
       @assert_global
       end
ERROR: Not in global scope
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] macro expansion at REPL[0]:7 [inlined]
 [3] top-level scope at REPL[2]:2
```

It's pretty ugly though...........................

PR #403, opened 2017-10-02
Add @compat for standard library imports

TotalVerb:
Alternative to #404.

See https://github.com/JuliaLang/julia/issues/23931.

The `SharedArrays` module is emulated instead of real.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,5 @@
 using Compat
-using Base.Test
+@compat using Test
 
 # Issue #291
 # 0.6
@@ -513,7 +513,7 @@ module Test18839
 
 using Compat
 using Compat.Iterators
-using Base.Test
+@compat using Test
 
 @test collect(take(countfrom(2), 3)) == [2, 3, 4]
 @test collect(take(cycle(5:8), 9)) == [5:8; 5:8; 5]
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -823,6 +824,19 @@ end
 # 0.7
 @test isconcrete(Int)
 
+# 0.7
+module Test23876
+    using Compat
+    @compat importall Test
```
yuyichao: Also `importall` is being deprecated

TotalVerb:
Module aliasing would force loading all the standard library modules as soon as Compatible loads, however.

yuyichao:
They are anyway.

TotalVerb:
Even so, module aliasing here introduces parts of Compat that will eventually need to be deprecated and migrated again, whereas the syntax rewriting can be removed later without deprecation.

test/runtests.jl
TotalVerb: ok, I'll avoid rewriting `importall` then

TotalVerb:
Alternatively, we could consider registering stub modules `Test`, `SharedArrays`, etc. that are restricted on METADATA for versions 0.6 and below?

yuyichao:
> that will eventually need to be deprecated and migrated again

Yes? What's the problem? It doesn't change the fact that syntax rewrite is always the least preferred way to handle version compatibility, especially for a non-syntax change.

> registering stub modules Test, SharedArrays, etc

That can also work.

TotalVerb:
I agree in principle, but it just seems more elegant to be able to do `@compat using Test` instead of `using Compat.Test`. Would someone else like to comment?

TotalVerb:
Replaced by #404.

PR #404, opened 2017-10-02
Add Compat for standard library imports

TotalVerb:
Alternative to #403. Feedback on which approach is preferable would be helpful.

cc @yuyichao

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ Currently, the `@compat` macro supports the following syntaxes:
   `CartesianRange` now has two type parameters, so using them as
   fields in other `struct`s requires manual intervention.
 
+* `using Compat.Test`, `using Compat.SharedArrays`, `using Compat.Mmap`, and `using
+  Compat.DelimitedFiles` are provided on versions older than 0.7, where these are not yet
+  part of the standard library. ([#23931])
+
 ## Module Aliases
 
 * In 0.6, some 0.5 iterator functions have been moved to the `Base.Iterators`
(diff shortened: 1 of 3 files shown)


nalimilan:
I don't have strong feelings about which solution is best, except that I'd rather use the one which FemtoCleaner will be able to remove automatically after dropping support for 0.6 (if it exists).

TotalVerb:
I think both are able to be removed easily by FemtoCleaner.

musm:
Maybe just stick to current convention  as in this PR. We already have to do e.g. `using Compat.Sys.islinux` so this seems consistent.

TotalVerb:
It is not quite the same because `Compat.Sys` need not be deprecated and removed, whereas `Compat.Test` will be...

yuyichao:
There are many that needs deprecation and removed.

TotalVerb:
Which existing things need deprecation?

yuyichao:
Currently active one that needs deprecation includes `MathConstants`, `invokelatest`, `StringVector`.

Also see the currently deprecating list which includes `KERNEL`, `UTF8String`, `ASCIIString`, `unsafe_convert`, `remote_do`, `Filesystem`, `AsyncCondition`, `promote_eltype_op`, `@irrational`, `@blasfunc`.

TotalVerb:
I'm not convinced that this is a module rename and should be handled similarly. `using Test` does something very different than `using Compat.Test` does, because it's loading a package, whereas `using Compat.Test` is just importing the exported symbols of an already-loaded module. Isn't it possible that standard library packages will not be loaded by default in the future?

TotalVerb:
In the spirit that "something is better than nothing", I am merging this PR. It will require a deprecation cycle later, but femtocleaner is perfectly capable of cleaning such things up.

PR #406, opened 2017-10-17
character and string concatenation

chakravala:
this adds backwards compatibility for character with string concatenation https://github.com/JuliaLang/julia/pull/22532


chakravala:
looks like all tests are good

chakravala:
will this get tagged when it is merged? there are multiple packages that need this in order to upgrade to 0.7 while maintaining backwards compatibility. simple fix that will solve a lot of 0.7 package upgrades

PR #408, opened 2017-11-07
Add (sp)diagm pair constructors

ararslan:


fredrikekre:
I think we need something like:
```julia
# 0.7.0-DEV.2116
@static if VERSION < v"0.7.0-DEV.2116"
    Base.spdiagm(kv::Pair...) = spdiagm(last.(kv), first.(kv))
end

# 0.7.0-DEV.2161
@static if VERSION < v"0.7.0-DEV.2161"
    function Base.diagm(kv::Pair...)
        T = promote_type(map(x -> eltype(x.second), kv)...)
        n = mapreduce(x -> length(x.second) + abs(x.first), max, kv)
        A = zeros(T, n, n)
        for p in kv
            inds = diagind(A, p.first)
            for (i, val) in enumerate(p.second)
                A[inds[i]] += val
            end
        end
        return A
    end
end
```

ararslan:
Ah, looks like the tests will need to be adjusted. I guess some sparse/dense equality stuff is 0.7 only?

fredrikekre:
> Ah, looks like the tests will need to be adjusted. I guess some sparse/dense equality stuff is 0.7 only?

No, the difference is that `spdiagm` always return a square matrix on 0.7, which `diagm` has always done. I can fix up this PR later.

ararslan:
Looks like the Travis failures are now that their macOS images are having issues. Thanks for your help here!

fredrikekre:
failures are https://github.com/JuliaLang/julia/issues/24658

PR #416, opened 2017-11-25
Add function-first finalizers

### LINKED ISSUE #414: Finalizers take function as first argument
[Finalizers take function as first argument](https://github.com/JuliaLang/julia/pull/24605/files)

ararslan:
Fixes #414

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Uninitialized` and `uninitialized` with corresponding `Array` constructors ([#24652]).
 
+* `finalizer` accepts the finalizer to run as the first argument and the object to be finalized
+  as the second ([#24605])
+
 ## Renaming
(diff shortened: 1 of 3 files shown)


Nosferican:
#414 for tracking.

fredrikekre:
```
julia> Test.detect_ambiguities(Base)
Skipping Base.<|
1-element Array{Tuple{Method,Method},1}:
 (finalizer(f::Function, o) in Compat at /home/fredrik/.julia/v0.6/Compat/src/Compat.jl:830, finalizer(o::T, f::Ptr{Void}) where T in Base at base.jl:133)
```

fredrikekre:
We should probably add a test like this:
https://github.com/JuliaCollections/DataStructures.jl/blob/71e6001be48e311307596887cc687f7b329e7101/test/runtests.jl#L6
to `Compat` as well.

PR #418, opened 2017-11-28
Base64

Nosferican:
Addresses Julia [0.7.0-DEV.2338](https://github.com/JuliaLang/julia/commit/a8f8df6cc6d5bdadbadbc2dfbdec67f547470e4b).

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -769,6 +769,11 @@ end
     end
 end
 
+# 0.7.0-DEV.2338
+if VERSION >= v"0.7.0-DEV.2338"
+    import Base64
+end
+
 if VERSION < v"0.7.0-DEV.2377"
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -769,6 +769,11 @@ end
     end
 end
 
+# 0.7.0-DEV.2338
+if VERSION >= v"0.7.0-DEV.2338"
+    import Base64
+end
```
fredrikekre: I think we need
```julia
@static if VERSION >= v"0.7.0-DEV.2338"
    import Base64
else
    import Base.Base64
end
```
such that `Compat.Base64` is available on older versions too.

Nosferican: Sounds good. I will make the change.

PR #419, opened 2017-11-28
Cleanup Julia 0.7 deprecations

omus:
Eliminates almost all the deprecations within Compat on Julia 0.7. Changes should be compatible with Julia 0.5 and 0.6 as well.

PR #421, opened 2017-11-30
update deprecated array constructor

fredrikekre:
These where introduced in #419 but just got deprecated on master (https://github.com/JuliaLang/julia/pull/24781)


fredrikekre:
Actually, this changes what was intended to test #330 so we should just upper bound the test here instead I think and revert the change in #419. Does that make sense @Sacha0 ?

Sacha0:
> Actually, this changes what was intended to test #330 so we should just upper bound the test here instead I think and revert the change in #419. Does that make sense `@Sacha0`?

Yes, that proposal makes sense on this end :).

PR #422, opened 2017-12-01
Revert function first finalizers due to ambiguities

ararslan:
cc also @KristofferC

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,8 +192,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Uninitialized` and `uninitialized` with corresponding `Array` constructors ([#24652]).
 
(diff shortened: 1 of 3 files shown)


iamed2:
How about a `@compat` recipe that flips the call to `finalizer` on old versions?

ararslan:
I don't know that I'll have time to look into that myself but if someone wants to take that on, it seems like a reasonable idea.

rofinn:
I'll give it a shot. I'm pretty sure we'll want it for Memento and Dispatcher at least.

fredrikekre:
> I'll give it a shot. I'm pretty sure we'll want it for Memento and Dispatcher at least.

Ah, sorry, I already had the fix, should have mentioned it

PR #423, opened 2017-12-02
add KeySet

fredrikekre:


fredrikekre:
ref https://github.com/JuliaCollections/DataStructures.jl/pull/348

fredrikekre:
This is a bit iffy. Probably better for packages that need this to solve it locally. I guess this type is not used that much anyway.

andyferris:
> Probably better for packages that need this to solve it locally. I guess this type is not used that much anyway.

I find this hard to judge (I've had little experience with `Compat` anyway).

fredrikekre:
In all registered packages, apart from in DataStructures, this: https://github.com/BenLauwens/ResumableFunctions.jl/blob/ec55bcf9bc1cd24ba84b28743e10953795f1d90d/src/transforms.jl#L22 is the only place where `KeyIterator` is used. Since this PR does not offer the complete functionality of `KeySet` I think it is better to do local fixes for this, and not provide it from `Compat`.

PR #427, opened 2017-12-05
IOContext(::IO, ::Pair, ::Pair...)

fredrikekre:


src/Compat.jl
```diff
@@ -807,6 +807,11 @@ end
     (::Type{Array{T}}){T}(s::UniformScaling, m::Integer, n::Integer) = Matrix{T}(s, m, n)
 end
 
+# https://github.com/JuliaLang/julia/pull/23271
+@static if VERSION < v"0.7.0-DEV.1472"
+    Base.IOContext(io::IO, arg::Pair, args::Pair...) = IOContext(io; Dict(arg, args...)...)
```
rfourquet: I think `IOContext(io::IO, arg::Pair)` was already available. So if performance is a concern, it may (I'm not sure) be faster to avoid the creation of a `Dict`, like:
```
 Base.IOContext(io::IO, arg1::Pair, arg2::Pair, args::Pair...) = IOContext(IOContext(io, arg1), arg2, args...)
```

fredrikekre: Done.

src/Compat.jl
```diff
@@ -807,6 +807,13 @@ end
     (::Type{Array{T}}){T}(s::UniformScaling, m::Integer, n::Integer) = Matrix{T}(s, m, n)
 end
 
+# https://github.com/JuliaLang/julia/pull/23271
+@static if VERSION < v"0.7.0-DEV.1472"
+    Base.IOContext(io::IO, arg1::Pair, arg2::Pair, args::Pair...) = IOContext(IOContext(io, arg1), arg2, args...)
+    # needed for ambiguity resolution
+    Base.IOContext(io::IOContext, arg1::Pair, arg2::Pair) = IOContext(IOContext(io, arg1), arg2)
```
rfourquet: Good catch!

PR #431, opened 2017-12-14
`Complex<2N>` to `ComplexF<N>` rename

martinholters:


martinholters:
Failures on nightly are unrelated (due to Unicode moval to stdlib); I have verified locally that tests pass after commenting out the few tests that rely on the unicode stuff.

PR #432, opened 2017-12-15
Add Compat.Unicode for standard library Unicode

omus:
Changed in https://github.com/JuliaLang/julia/pull/25021. At the moment I disabled the OffsetArray tests as they cause the tests to fail on Julia 0.7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ Currently, the `@compat` macro supports the following syntaxes:
   `CartesianRange` now has two type parameters, so using them as
   fields in other `struct`s requires manual intervention.
 
+## Module Aliases
+
+* In 0.6, some 0.5 iterator functions have been moved to the `Base.Iterators`
+  module. Code can be written to work on both 0.5 and 0.6 by `import`ing or
+  `using` the `Compat.Iterators` module instead. ([#18839])
+
 * `using Compat.Test`, `using Compat.SharedArrays`, `using Compat.Mmap`, and `using
   Compat.DelimitedFiles` are provided on versions older than 0.7, where these are not yet
   part of the standard library. ([#23931])
@@ -89,11 +95,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `using Compat.Dates` is provided on versions older than 0.7, where this library is not
   yet a part of the standard library. ([#24459])
 
-## Module Aliases
-
-* In 0.6, some 0.5 iterator functions have been moved to the `Base.Iterators`
-  module. Code can be written to work on both 0.5 and 0.6 by `import`ing or
-  `using` the `Compat.Iterators` module instead. ([#18839])
+* `using Compat.Unicode` is provided on versions older than 0.7, where this library is not
+  yet a part of the standard library. ([#25021])
 
 ## New functions, macros, and methods
 
@@ -236,8 +239,6 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `IntSet` is now `BitSet` ([#24282])
 
-* `strwidth` and `charwidth` are now merged into `textwidth` ([#23667]).
-
 * `Complex32`, `Complex64`, and `Complex128` are now `ComplexF16`, `ComplexF32`, and
   `ComplexF64`, respectively ([#24647]).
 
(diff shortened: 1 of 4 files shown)


src/Compat.jl
```diff
@@ -752,13 +752,6 @@ end
     export BitSet
 end
 
-# 0.7.0-DEV.1930
-@static if !isdefined(Base, :textwidth)
-    textwidth(c::Char) = charwidth(c)
-    textwidth(c::AbstractString) = strwidth(c)
-    export textwidth
-end
-
```
omus: I'm not sure what the procedure with deprecating a Compat supported change is. Since the rule for Compat.jl is to use the latest syntax it seemed reasonable to move `textwidth` into the `Unicode` module. Do I need to deprecate this syntax?

test/runtests.jl
```diff
@@ -628,11 +628,11 @@ using Compat: StringVector
...
-using OffsetArrays
-a = OffsetArray(1:3, -1:1)
-b = Compat.collect(a)
-@test indices(b) === (Base.OneTo(3),)
-@test b == [1,2,3]
+# using OffsetArrays
+# a = OffsetArray(1:3, -1:1)
+# b = Compat.collect(a)
+# @test indices(b) === (Base.OneTo(3),)
+# @test b == [1,2,3]
```
omus: Generates the error:
```
ERROR: LoadError: size not supported for arrays with indices (-1:1,); see http://docs.julialang.org/en/latest/devdocs/offset-arrays/
```
I haven't looked into this.

README.md
```diff
@@ -89,11 +95,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `using Compat.Dates` is provided on versions older than 0.7, where this library is not
   yet a part of the standard library. ([#24459])
 
-## Module Aliases
-
-* In 0.6, some 0.5 iterator functions have been moved to the `Base.Iterators`
-  module. Code can be written to work on both 0.5 and 0.6 by `import`ing or
-  `using` the `Compat.Iterators` module instead. ([#18839])
+* `using Compat.Unicode` is provided on versions older than 0.7, where this library is not
+  yet a part of the standard library. ([#25021])
```
omus: The diff is messed up here. I moved all of the `using X` notes into the "Module Aliases" section.

src/Compat.jl
```diff
@@ -874,6 +867,43 @@ end
     export ComplexF64
 end
 
+# 0.7.0-DEV.2915
+module Unicode
+    export normalize, graphemes, isassigned, textwidth, isvalid,
```
omus: I should remove `normalize` and `isassigned` here. https://github.com/JuliaLang/julia/pull/25079

src/Compat.jl
martinholters: I would prefer a deprecation over a breaking change. It is quite common for packages to require a fairly recent Compat version, so if others need to set an upper bound, that could be rather unfortunate.

But I have no idea want the standard procedure is.

src/Compat.jl
```diff
@@ -874,6 +867,43 @@ end
...
+                if isspace(c)
+                    print(b, c)
+                    startword = true
+                else
+                    print(b, startword ? titlecase(c) : c)
+                    startword = false
+                end
+            end
+            return String(take!(b))
+        end
```
omus: Failure on 0.6 due to `titlecase` being exported by Base in `6f8e94e3ff` (https://github.com/JuliaLang/julia/pull/19469)

src/Compat.jl
omus: I added a deprecation for this so the transition should be non-breaking.

omus:
Needs #435 and OffsetArrays to pass on 0.7

src/deprecated.jl
```diff
@@ -188,3 +188,7 @@ else
     import Base.@irrational
     import Base.LinAlg.BLAS.@blasfunc
 end
+
+if VERSION < v"0.7.0-DEV.2915"
+    Base.@deprecate textwidth Compat.Unicode.textwidth
```
nalimilan: I don't understand the justification for these lines. They are annoying as they print deprecation warnings *on Julia 0.6*. These can appear for packages which worked perfectly well before, just because Compat has been updated.

omus: Since `textwidth` needed to be moved inside `Compat.Unicode` module the only two real choices were to either make a alias at the top-level or to add a deprecation. I ended up going for the deprecation since Compat is about supporting "the _latest syntax_ in a backwards-compatible way".

omus: I personally haven't seen this scenario where previous Compat code was required to be moved. There may be a standard procedure on how to handle this for which I am unaware.

nalimilan: OK, I see the logic. However I think it's bad to generate tons of deprecation warnings on Julia 0.6 for packages which worked perfectly fine until now. For example, DataFrames is very painful to use right now because of this.

I think the best approach is to keep exporting this function on Julia < 0.7 since it's supported by Julia 0.6, but not on Julia 0.7. We don't need to deprecate the function ourselves since Julia takes care of that already. See https://github.com/JuliaLang/Compat.jl/pull/437.

omus: Sounds good to me. We should think about making a CONTRIBUTING.md to Compat.jl to clarify these kind of details.

PR #435, opened 2017-12-20
A bunch of stuff for 0.7

ararslan:


ararslan:
Nightly tests will need #432 in order to pass here.

src/Compat.jl
```diff
@@ -953,6 +982,67 @@ module Unicode
...
+@static if !isdefined(Base, :AbstractDict)
+    const AbstractDict = Associative
+    export AbstractDict
+end
+
+# 0.7.0-DEV.2978
+@static if !isdefined(Base, :axes)
+    const axes = Base.indices
+    export axes
+end
```
fredrikekre: This should  have 
```julia
else
    import Base.axes
end
```
such that you can overload `Compat.axes` on 0.6 too.

fredrikekre: Nevermind... I confused myself.

tknopp:
@ararslan: This seem to have broken `ImageAxes` package (see https://discourse.julialang.org/t/issues-with-axisarrays-on-0-6/8064). Not sure how and where to fix.

ararslan:
I assume it's similarly broken on 0.7 then, since Base exports `axes` on 0.7. I'm not familiar with ImageAxes, but if it wouldn't be type piracy then I recommend extending `axes` from Compat.

omus:
If we didn't export `axes` from Compat then the change wouldn't impact other packages. That is kind of a work around however.

ararslan:
> If we didn't export `axes` from Compat then the change wouldn't impact other packages

Indeed, see #442.

tknopp:
Many thanks @ararslan. For 0.7 we now need discussion on the meaning of this function. Either base or AxisArrays need to rename it. Since there are now so many threads out there: Where should be initiate that discussion.

tknopp:
maybe here https://github.com/JuliaLang/julia/pull/25057
is the best discussion place

PR #436, opened 2017-12-21
Temporarily disable the OffsetArrays tests for 0.7

ararslan:
The functionality being tested is only provided in Compat for compatibility with Julia 0.5, which is near end of life, and OffsetArrays currently has some issues on 0.7 that cause unrelated test failures in this package.


ararslan:
The IterativeEigensolvers failure that only happens on Linux with Julia 0.7 is rather bizarre.

ararslan:
Ah, the Linux nightly binaries are a few days old, before the IterativeEigensolvers rename. The logic for that piece wasn't great so I've improved it in a separate commit.

PR #412, opened 2017-11-20
Add constructors for `Matrix{T}`, `Array{T}`, and `SparseMatrixCSC{T}` from `UniformScaling`

martinholters:
I have omitted the variants without explicit element type due to the change of the default element type of `I`.

PR #438, opened 2017-12-27
Add Matrix(I, dims) constructor

### LINKED ISSUE #468: Add `Matrix(I, n, m)` method
In Julia v0.7 you don't need to specify the eltype of the matrix - it is inferred from the `UniformScaling{T}` input. Compat has only defined the `Matrix{T}(::UniformScaling, n, m)` method.

```
# Julia v0.6.2
julia> using Compat

julia> Matrix(I, 4, 4)
ERROR: MethodError: no method matching Array{T,2} where T(::UniformScaling{Int64}, ::Int64, ::Int64)
Closest candidates are:
  Array{T,2} where T(::Compat.Uninitialized, ::Any...) at string:5
  Array{T,2} where T(::Integer, ::Integer) at sysimg.jl:105

julia> Matrix{Float64}(I, 4, 4)
4×4 Array{Float64,2}:
 1.0  0.0  0.0  0.0
 0.0  1.0  0.0  0.0
 0.0  0.0  1.0  0.0
 0.0  0.0  0.0  1.0
```

ararslan:
fix  #468

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * Constructors for `Matrix{T}`, `Array{T}`, and `SparseMatrixCSC{T}` from `UniformScaling` ([#24372], [#24657])
 
+* Constructor for `Matrix` from `UniformScaling` ([#24372], [#24657]).
+
 * `Uninitialized` and `uninitialized` with corresponding `Array` constructors ([#24652]).
 
 * `BitArray` constructors for `uninitialized` ([#24785]).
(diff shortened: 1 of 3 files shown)


fredrikekre:
Note that these where intentionally excluded in https://github.com/JuliaLang/Compat.jl/pull/412

ararslan:
Why the revival? Don't the concerns in #412 still apply?

fredrikekre:
> Why the revival? Don't the concerns in #412 still apply?

Meh, it is not the fault of the `Matrix `constructor that the element type of `I` has changed, that is documented as a breaking change separately :)

ararslan:
Works for me :)

martinholters:
Just so it be said: I didn't include them in #412 because I was unsure about them, not because I was opposed to them.

PR #439, opened 2017-12-28
Moved BINDIR definition into '__init__'.

rofinn:


rofinn:
Is there an alternative that would allow Compat to be used in system images.

src/Compat.jl
```diff
@@ -585,10 +585,11 @@ module Sys
...
 
-    @static if VERSION < v"0.7.0-DEV.3073"
-        const BINDIR = JULIA_HOME
-    else
-        const BINDIR = Base.Sys.BINDIR
+    # https://github.com/JuliaLang/julia/pull/25102
+    # NOTE: This needs to be in an __init__ because JULIA_HOME is not
+    # defined when building system images.
+    function __init__()
+        global const BINDIR = VERSION < v"0.7.0-DEV.3073" ? JULIA_HOME : Base.Sys.BINDIR
```
omus: Would changing this to no longer be `const` be acceptable?

stevengj: Is the `__init__` required in 0.7?

rofinn: I don't know because my attempts to build a system image on macos and linux both produce segfaults.

omus: After reviewing https://github.com/JuliaLang/julia/pull/25102 I think that the `__init__` is also required on 0.7 as well as I believe that the variable is set at runtime as part of `jl_resolve_sysimg_location`.

stevengj: `isconst(Sys, :BINDIR)` is `false`, so `Compat.BINDIR` should not be `const` anyway.

iamed2:
@yuyichao is the current version acceptable?

PR #440, opened 2017-12-29
Support RangeGeneratorInt rename

malmaud:


malmaud:
Good to merge?

ararslan:
Needs an entry in the README

src/Compat.jl
```diff
@@ -1051,6 +1051,17 @@ else
     import Base: notnothing
 end
 
+module Random
+
+    # Cope with renaming of RangeGeneratorInt to SamplerRangeInt
+    @static if !isdefined(Base.Random, :SamplerRangeInt)
+        SamplerRangeInt = Base.Random.RangeGeneratorInt
```
rfourquet: Don't you want a `const` here?

test/runtests.jl
```diff
@@ -1063,6 +1063,8 @@ end
 @test Compat.notnothing(1) == 1
 @test_throws ArgumentError Compat.notnothing(nothing)
 
+@test isa(Base.Random.RangeGenerator(1:3), Compat.Random.SamplerRangeInt)
```
rfourquet: `Base.Random.RangeGenerator` has been kept as an alias for compatibility, but eventually it will go; but I guess this test is fine for now. Also, `Random` will move to stdlib.

rfourquet:
This change looks good, do you have a link to an example where it would be used? It's possible that for `MersenneTwister`, you would get generally better efficiency with using `SamplerRangeFast` (I'm considering making this the default always in `Base`). Also, as this PR is to support a newly introduced name, I wonder if this is this best stategy, as depending on the RNG and the number of times you want to generate a random number, the `Random.Sampler` constructor would choose between `SamplerRangeInt` and `SamplerRangeFast`. Cf also https://github.com/JuliaStats/Distributions.jl/issues/697.

malmaud:
This was designed for https://github.com/JuliaStats/Distributions.jl/pull/700/files. Happy to implement or defer to your implementation of a better approach; I just wanted to get Distributions.jl working again.

rfourquet:
I think we can go with your solution for now, and see if a better solution exists when things have stabilized after 1.0 is out.

quinnj:
needs a rebase

tlnagy:
Any update on this?

martinholters:
Updated on top of #601. The test of course fails on Julia past 0.7 as `Base.Random` is gone, so needs updating. Also, on 0.7 it fails because `RangeGenerator` there gives a `SamplerRangeFast` instead of a `SamplerRangeInt`.

martinholters:
Closing as outdated since we've dropped support for Julia prior to 1.0.

PR #443, opened 2017-12-30
Added Compat.AbstractDateTime

rofinn:


src/Compat.jl
```diff
@@ -769,6 +769,12 @@ else
     import Dates
 end
 
+if VERSION < v"0.7.0-DEV.3216"
+    const AbstractDateTime = Dates.TimeType
```
ararslan: Should this (and below) be `Compat.Dates` rather than just `Dates`?

ararslan:
Cool, thanks! Could you add a test and an entry in the README?

test/runtests.jl
```diff
@@ -1063,6 +1063,10 @@ end
 @test Compat.notnothing(1) == 1
 @test_throws ArgumentError Compat.notnothing(nothing)
 
+# 0.7.0-DEV.3216
+@test Compat.AbstractDateTime === (isdefined(Compat.Dates, :AbstractDateTime) ? Compat.Dates.AbstractDateTime : Compat.Dates.TimeType)
+@test Compat.AbstractDateTime <: Compat.Dates.TimeType
```
omus: Probably should have a test which is `Compat.Dates.DateTime <: Compat.AbstractDateTime`

src/Compat.jl
```diff
@@ -769,6 +769,12 @@ else
     import Dates
 end
 
+if VERSION < v"0.7.0-DEV.3216"
+    const AbstractDateTime = Compat.Dates.TimeType
+else
+    const AbstractDateTime = Compat.Dates.AbstractDateTime
+end
+
```
omus: Probably should export this

src/Compat.jl
```diff
@@ -769,6 +769,12 @@ else
     import Dates
 end
 
+if VERSION < v"0.7.0-DEV.3216"
+    const AbstractDateTime = Compat.Dates.TimeType
```
omus: Seems slightly dangerous to use with dispatch. You could accidentally be given a `Date` or `Time`. I'm not sure what a better solution would be ATM

omus: Possibly just having a `const AbstractDateTime = Union{DateTime, ZonedDateTime}` in TimeZones.jl would be a better solution. Doing this in Compat would make TimeZones a requirement in Compat which is probably worse.

src/Compat.jl
rofinn: That seems like an odd choice given that it isn't exported in base.

src/Compat.jl
rofinn: That could also work for our use cases, but I figured it's better to write code that is a bit too general than too restrictive. IFAICT, the worst case scenario is that some passes in a `Date` or `Time` and the code errors at a later point... which I guess would make the traceback a little bit more complicated.

src/Compat.jl
omus: You are right here. Don't export this since it isn't exported in Base

src/Compat.jl
omus: A possible alternative solution is to implement `@compat AbstractDateTime` which would re-write this as a `Union{DateTime,ZonedDateTime}`. That would probably be a bit too annoying though.

omus:
@rofinn I'm ok if we go ahead with this change. I definitely don't like how overly general the fallback type is but it should probably be ok as a temporary compatibility change.

omus:
Any other comments? I think this approach may be the best we can do for now. I'll merge tomorrow if there are no objections.

PR #446, opened 2018-01-06
Add CartesianIndices and LinearIndices

nalimilan:
Needed in particular to get linear indices once `find` returns cartesian indices (https://github.com/JuliaLang/julia/pull/24774).

@timholy I couldn't find a simple way of reusing `CartesianRange` from Base since it had a very different structure on 0.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.5
     - 0.6
     - nightly
 notifications:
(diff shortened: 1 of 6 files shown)


timholy:
This seems great, thanks!

Looks like there may be a missing `size` method (at least).

nalimilan:
Woops. Yet I thought I had run it on Julia 0.6... I've copied more code, this time it should work.

I've also added a commit to drop support for Julia 0.5, since official support has just ended and tests fail.

nalimilan:
OK to merge?

timholy:
:+1: The only question would be about dropping 0.5 support, but I'm a fan.

ararslan:
+1 from me too for dropping 0.5. In general if package authors are trying to support 0.5 and 0.7 at the same time, they're signing up for _way_ more work than it's worth.

PR #453, opened 2018-01-12
Add @info msg

EricForgy:


src/Compat.jl
```diff
@@ -1218,6 +1218,12 @@ end
     end
 end
 
+@static if !isdefined(Base, Symbol("@info"))
+    macro info(msg)
+        return :( info( $(esc(msg)) ) )
```
c42f: `@info` has a much richer syntax than just the message in 0.7.

You will need `macro info(msg, exs...)` at the very least.  You'd presumably then want to interpret the values in `exs...` as key value pairs in a similar way to the 0.7 version of `@info` and format them into the message eagerly.  (Minimum effort version would be to just discard `exs...` entirely, but that's not great.)

fredrikekre:
I think this is useful. People just want a simple replacement, and use the new syntax without depending on another package. I agree that we should also add `@warn`, `@error` and `@debug` though, and accept more input args. I think it is totally fine to just discard the extra metadata though, that functionality does not exist in 0.6, and people will understand that. We just have to document that extra arguments to those macros are discarded.

omus:
> For what it's worth, I feel this is a bit simplistic and I'm not convinced we should merge this stub when a fully functional version exists in `MicroLogging`. My concern is this `@info` will provide the same syntax with rather different semantics.

One of the issues I find with using MicroLogging.jl for Julia 0.6 compat is that the output doesn't conform to the old Julia 0.6 output or the new Julia 0.7 output. Personally I'd rather have a simplistic `@info` in Compat which keeps the Julia 0.6 output the same.

c42f:
> the output doesn't conform

I want to update `MicroLogging` to conform aesthetically with 0.7, and also fix a few minor API difference which have crept in.  Haven't found time to do it yet though.

fredrikekre:
Hmmm... I tried to add the changes but something went wrong... See #458. Sorry!

PR #455, opened 2018-01-13
Add size(::LinearIndices)

nalimilan:
Fixes indexing `LinearIndices` with an array. This is useful in particular to convert cartesian indices that `find` now returns to linear indices.

See https://github.com/JuliaLang/julia/pull/25541 (CI won't pass on 0.7 without it).

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1376,6 +1376,7 @@ end
     # AbstractArray implementation
     Base.IndexStyle(::Type{LinearIndices{N,R}}) where {N,R} = IndexCartesian()
     Compat.axes(iter::LinearIndices{N,R}) where {N,R} = iter.indices
+    Base.size(iter::LinearIndices{N,R}) where {N,R} = length.(iter.indices)
     @inline function Base.getindex(iter::LinearIndices{N,R}, I::Vararg{Int, N}) where {N,R}
         dims = length.(iter.indices)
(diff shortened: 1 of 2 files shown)


nalimilan:
CI failed on nightlies because Julia is still too old (except on Windows 64 bit for some reason).

ararslan:
> Julia is still too old (except on Windows 64 bit for some reason).

Windows is even older. It doesn't contain some of the breaking changes.

nalimilan:
64-bit Windows is relatively recent: `Commit d569a2923c* (2018-01-14 21:52 UTC)`.

martinholters:
> CI failed on nightlies because Julia is still too old

Isn't the purpose of Compat to make things work even if the used Julia is still too old?

nalimilan:
Yes, here the problem is that it's both too new and too old. Too new so that we use `Base.LinearIndices` rather than defining it in Compat, and too old because `Base.LinearIndices` doesn't support `size`. Adding that method would be type piracy, so I'm not sure it's worth doing this to support early 0.7 versions.

martinholters:
I think type piracy is ok for Compat---often unavoidable even---but not supporting a range of -DEV versions might be just as ok.

nalimilan:
I've added a commit to fix intermediate 0.7 releases so that CI passes. Should be good to go now.

PR #457, opened 2018-01-13
WIP: add find(first|next|prev|last) variants with `nothing` sentinel

timholy:
This will allow `Compat.findnext` to use the new `nothing` sentinel.

I'm frankly not sure what to do about several of the aspects of this PR, particularly the string-related ones. I'm also not sure we need the SparseArray changes in Compat? Advice appreciated (or help from anyone who wants to take it from here).

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1223,6 +1223,9 @@ end
     end
 end
 
+if VERSION < v"0.7.0-DEV.3399"
+    include("find.jl")
+end
 
 include("deprecated.jl")
 
--- a/src/find.jl
+++ b/src/find.jl
@@ -1,0 +1,897 @@
+const ByteArray = Union{Array{Int8,1}, Array{UInt8,1}}
+
+## from array.jl
+
+"""
+    findnext(A, i::Integer)
+
+Find the next linear index >= `i` of a `true` element of `A`, or `nothing` if not found.
+
+# Examples
+```jldoctest
+julia> A = [false false; true false]
+2×2 Array{Bool,2}:
+ false  false
+  true  false
+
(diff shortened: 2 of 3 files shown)


fredrikekre:
> I'm also not sure we need the SparseArray changes in Compat

Those where added just recently and are not there on 0.6, should be fine to use the slower fallbacks on 0.6.

fredrikekre:
AFAICT all of those returned 0 previously? Could the implementations here not simply be
```
function findnext(A, start::Integer)
    i = Base.findnext(A, start)
    if i == 0
        return nothing
    end
    return i
end
```

timholy:
That would be a little dangerous for the day when I want to do this:
```julia
a = OffsetArray([false, true, false], -1:1)
Compat.findfirst(a)
```
and expect it to work properly.

nalimilan:
Wow! Thanks for doing this radical backporting work!

However, to be honest, I'm hesitant about whether it's a good idea to copy all of these functions. @fredrikekre's proposal is appealing for its simplicity. All this duplicated code will have to be kept in sync with Base, which is tedious (I have already a few new PRs open) and which we will probably forget to do. Are you really concerned about people using the brand new `OffsetArrays` support on Julia 0.6? By the time users will have discovered it exists, we'll have dropped support for that release, and the error is relatively easy to catch anyway.

garrison:
Have these methods stabilized enough in Base that they could be brought in here?  I don't have a strong opinion on whether they should be copied or the simpler approach should be taken, but either way it would be great to come to a decision soon so porting can continue.

nalimilan:
Actually this has been covered by #484 and #513.

PR #458, opened 2018-01-16
Add Compat.at-error, Compat.at-warn, Compat.at-info and Compat.at-debug

fredrikekre:
replacement for #453 which I managed to mess up....

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,17 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `@nospecialize` has been added ([#22666]).
 
+* The logging macros `@error`, `@warn`, `@info` and `@debug` can be used as
+  `Compat.@error`, `Compat.@warn`, `Compat.@info` and `Compat.@debug` on Julia 0.6 ([#24490]).
+  Note that the behavior do not mirror the logging macros in Julia 0.7, instead on Julia 0.6:
+  - Messages are printed to `STDERR` (like `info` and `warn` on Julia 0.6) and not to a
+    dedicated logging stream.
+  - The loglevel can not be controlled, but `Compat.@debug` messages can be turned on/off
+    by calling `Compat.enable_debug(true/false)`.
+  - Extra metadata sent to the macros are ignored.
+
+  As an alternative, see the MicroLogging.jl package for a logging interface similar to the one in Julia 0.7.
+
 ## Other changes
 
 * On versions of Julia that do not contain a Base.Threads module, Compat defines a Threads module containing a no-op `@threads` macro.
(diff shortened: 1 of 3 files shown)


fredrikekre:
Added a reference to MicroLogging.

src/Compat.jl
```diff
@@ -1229,6 +1229,50 @@ end
     end
 end
 
+@static if !isdefined(Base, Symbol("@info"))
+    macro info(msg, args...)
+        return :(info($(esc(msg)), prefix = "Info: "))
```
c42f: If you want 0.6 equivalent formatting, perhaps skip the `prefix` for `@info` and `@warn`?

src/Compat.jl
```diff
@@ -1229,6 +1229,50 @@ end
...
+    macro debug(msg, args...)
+        return :(debug($(esc(msg))))
+    end
+else
+    @eval const $(Symbol("@debug")) = Base.$(Symbol("@debug"))
+end
+@static if !isdefined(Base, Symbol("@error"))
+    function _error(msg)
+        buf = IOBuffer()
+        iob = Base.redirect(IOContext(buf, STDERR), Base.log_info_to, :error)
```
c42f: Use `Base.log_error_to`, perhaps?

README.md
```diff
@@ -301,6 +301,15 @@ Currently, the `@compat` macro supports the following syntaxes:
...
 * `@nospecialize` has been added ([#22666]).
 
+* The logging macros `@error`, `@warn`, `@info` and `@debug` can be used as
+  `Compat.@error`, `Compat.@warn`, `Compat.@info` and `Compat.@debug` on Julia 0.6 ([#24490]).
+  Note that the behavior do not mirror the logging macros in Julia 0.7, instead on Julia 0.6:
+  - Messages are printed to `STDERR` (like `info` and `warn` on Julia 0.6) and not to a dedicated logging stream
+  - The loglevel can not be controlled
+  - Extra metadata sent to the macros are ignored
+
+  As an alternative, see the MicroLogging.jl package for a logging interface similar to the one in Julia 0.7.
```
c42f: This is accurate. While I plan to update `MicroLogging`, it's not possible to provide exact equivalence in a package.

src/Compat.jl
```diff
@@ -1229,6 +1229,50 @@ end
...
+    function debug(msg)
+        buf = IOBuffer()
+        iob = Base.redirect(IOContext(buf, STDERR), Base.log_info_to, :debug)
+        print_with_color(:blue, iob, "Debug: "; bold = true)
+        Base.println_with_color(:blue, iob, chomp(string(msg)))
+        print(STDERR, String(take!(buf)))
+        return
+    end
+    macro debug(msg, args...)
+        return :(debug($(esc(msg))))
```
c42f: Debug messages are invisible by default in 0.7, so I think having them visible here by default might end up being hopelessly verbose.

Maybe a simplistic global flag which disables debug by default might be a good idea?

src/Compat.jl
fredrikekre: I did it like this to be consistent with 0.7, and also to differentiate logging errors from actual errors, which prints `ERROR: ` in all caps red.

src/Compat.jl
fredrikekre: Good idea, we can have `const DEBUG = Ref(false)` in Compat, such that people can turn it on.

PR #461, opened 2018-01-18
Add parentmodule

ararslan:


ararslan:
Looks like this will need #463 in order to pass tests on Travis. (The Windows nightly is apparently pretty old.)

ararslan:
Hmm, 0.7 is failing here saying that `Diagonal` moved to LinearAlgebra. Which is weird given that there's a `using Compat.LinearAlgebra` at the top...

fredrikekre:
Doesn't happen on master, but I get that failure locally when checking out this PR, but after rebasing on master everything works.

ararslan:
Cool, thanks for investigating!

PR #463, opened 2018-01-19
Compat.LinearAlgebra

fredrikekre:


src/Compat.jl
```diff
@@ -796,6 +796,12 @@ else
     import IterativeEigensolvers
 end
 
+@static if VERSION < v"0.7.0-DEV.3449"
+    const LinearAlgebra = Base.LinAlg
```
ararslan: Does this work with `using`/`import` statements?

fredrikekre: Yes :)

PR #465, opened 2018-01-20
Compat for Libdl moving to stdlib

quinnj:


tkelman:
this needs to be in the readme, and tested

PR #466, opened 2018-01-21
Compat for findall

quinnj:


tkelman:
needs to be in readme and tested

PR #470, opened 2018-01-22
Add argmin and argmax

ararslan:


martinholters:
Too bad I'm a bit late, but I think this is not quite correct because for ndims>1, `indmin` returns a linear index in Julia 0.6 (more specifically before 0.7.0-DEV.1660), while `argmin` returns a `CartesianIndex`. So we should define `argmin` as a wrapper around `indmin`, putting the result into a `CartesianIndex`, if `VERSION < v"0.7.0-DEV.1660"`.

fredrikekre:
Good catch, lets make sure we fix that before making a new Compat release.

ararslan:
Yes very good catch, @martinholters. Would you be willing to submit a PR for that?

martinholters:
Yes, but might need a day or two.

ararslan:
Sure thing. I'll see if I can get around to it sooner and I'll ping you if I beat you to it, but I might not be able to.

PR #472, opened 2018-01-24
Fix `argmin`/`argmax` for non-vectors

martinholters:
Ref. https://github.com/JuliaLang/Compat.jl/pull/470#issuecomment-359798659.

I don't know whether there is a completely generic way of replicating the functionality on Julia 0.6, but I guess I have covered the most common cases.

CC @ararslan

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1311,8 +1311,17 @@ if !isdefined(Base, :findall)
 end
 
 @static if !isdefined(Base, :argmin)
-    const argmin = indmin
-    const argmax = indmax
+    if VERSION >= v"0.7.0-DEV.1660" # indmin/indmax return key
+        const argmin = indmin
+        const argmax = indmax
+    else
+        argmin(x::AbstractArray) = CartesianIndex(ind2sub(x, indmin(x)))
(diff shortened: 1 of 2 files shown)


src/Compat.jl
```diff
@@ -1311,8 +1311,17 @@ if !isdefined(Base, :findall)
 end
 
 @static if !isdefined(Base, :argmin)
-    const argmin = indmin
-    const argmax = indmax
+    if VERSION >= v"0.7.0-DEV.1660" # indmin/indmax return key
+        const argmin = indmin
+        const argmax = indmax
+    else
+        argmin(x::AbstractArray) = CartesianIndex(ind2sub(x, indmin(x)))
```
nalimilan: Probably better use `CartesianIndices` instead of `ind2sub`, given that the latter is deprecated? Otherwise, looks good.

martinholters: This branch is only taken for `VERSION < v"0.7.0-DEV.1660"`, but the deprecation was in `0.7.0-DEV.3025` IIUC.

nalimilan: OK, cool.

PR #473, opened 2018-01-24
Compat for Serialization stdlib

omus:


omus:
Ack, I do need tests! I'm not looking forward to being put back in the Travis queue...

omus:
Tests added. Feel free to squash.

omus:
Failures look unrelated:
```
Error During Test at /home/travis/.julia/v0.7/Compat/test/runtests.jl:1079
  Test threw an exception of type MethodError
  Expression: Compat.IteratorSize(v) == Base.HasShape()
  MethodError: no method matching Base.HasShape()
  Stacktrace:
   [1] top-level scope at /home/travis/.julia/v0.7/Compat/test/runtests.jl:1079
   [2] include at ./boot.jl:292 [inlined]
   [3] include_relative(::Module, ::String) at ./loading.jl:1012
   [4] include(::Module, ::String) at ./sysimg.jl:26
   [5] process_options(::Base.JLOptions) at ./client.jl:334
   [6] _start() at ./client.jl:386
ERROR: LoadError: There was an error during testing
in expression starting at /home/travis/.julia/v0.7/Compat/test/runtests.jl:1079
```

omus:
@ararslan I think we're ready here.

fredrikekre:
Out of curiousity, what was wrong with the first version:
```
if VERSION < v"0.7.0-DEV.3476"
    const Serialization = Base.Serializer
else
    import Serialization
end
```

omus:
If I remember correctly since the module name is different you wouldn't be able to do:
```julia
import Compat: Serialization
using Serialization
```
You can see this on Julia 0.6 with this example:
```julia
julia> const Serialization = Base.Serializer
Base.Serializer

julia> using Serialization
ERROR: UndefVarError: Serialization not defined
```

fredrikekre:
But `using Compat.Serialization` would work, right? That is the way (most) of the other stdlibs in Compat work. The problem with the approach in this PR is that only the names explicitly imported into the `Compat.Serialization` module can be used.

omus:
> But `using Compat.Serialization` would work, right?

Yes, it does work. I originally used `const` implementation but unfortunately that approach did not have `Serialization` defined which is the primary way we test for module compatibility. For example:

```julia
julia> VERSION
v"0.6.2"

julia> module Compat
       const Serialization = Base.Serializer
       end
Compat

julia> using Compat.Serialization

julia> isdefined(:Serialization)
false
```

> The problem with the approach in this PR is that only the names explicitly imported into the `Compat.Serialization` module can be used.

I understand this concern and it is also a concern for me. That said some of the internals of `Serialization` have changed from `Base.Serializer` which means that there are some features available in 0.7 which are not available earlier (`AbstractSerializer`). Only exporting the shared features means that using what Compat provides is consistent for different Julia versions. 

Looking at the code again I just noticed that `SerializationState` is deprecated so we'll need to do an update like:

```julia
@eval module Serialization
    import Base.Serializer: serialize, deserialize, SerializationState
    const Serializer = SerializationState
    export serialize, deserialize, Serializer
end
```

(provided that the `SerializationState` deprecation was just a rename)

PR #474, opened 2018-01-24
codeunits and ncodeunits

stevengj:
Added these functions from JuliaLang/julia#25241, which are especially useful for `Vector{UInt8}(string)` deprecations.

(See also JuliaIO/JSON.jl#234)


stevengj:
Looks like an unrelated Pkg error on nightly MacOS.

stevengj:
Rebased to see if CI is green now.

stevengj:
Yay, CI green.

PR #476, opened 2018-01-26
Handle `HasShape` change in tests

martinholters:
On newer Julia, `HasShape` has to be instantiated with a type parameter giving the dimensionality.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1076,7 +1076,7 @@ end
 
 # 0.7.0-DEV.3309
 let v = [1, 2, 3]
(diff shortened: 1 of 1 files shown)


martinholters:
Simplified the construct here to an `isa`, circumventing the problem.

> I guess there's no way to handle that in Compat?

We could have `@compat HasShape{N}()` translate to `HasShape()` on older Julia, but obviously, we can't support dispatching on `::HasShape{N}`. But I'm not sure it's worth it and I'd like to keep this PR slim to fix the test failures on Julia master as soon as possible.

PR #478, opened 2018-01-30
deprecate isconcrete and use isconcretetype

vchuravy:


vchuravy:
#477 has this as well. (Using Compat in anger...)

PR #480, opened 2018-02-01
Add `lastindex`

martinholters:
Where not provided by Julia, it is just defined as an alias for `endof`, hence does not support `lastindex(A, n)`, as `endof` doesn't.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `gc` and `gc_enable` are now `GC.gc` and `GC.enable`, respectively ([#25616]).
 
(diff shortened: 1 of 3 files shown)


martinholters:
@timholy, you're exactly the right one to merge this, as I did this after seeing
```
┌ Warning: `endof(a)` is deprecated, use `lastindex(a)` instead.
│   caller = @showprogress(::LineNumberNode, ::Module, ::Vararg{Any,N} where N) at ProgressMeter.jl:417
└ @ ProgressMeter ProgressMeter.jl:417
```

timholy:
Let's tag a new version, I want `lastindex` for Revise too.

PR #481, opened 2018-02-01
Add printstyled

ararslan:


ararslan:
Looks like there's an issue with the test with 0.6 on Windows. `print_with_color` didn't have its own test in 0.6, so I'm not sure what the best backwards-compatible test would be.

test/runtests.jl
```diff
@@ -1297,4 +1297,10 @@ end
 @test lastindex(zeros(4)) == 4
 @test lastindex(zeros(4,4)) == 16
 
+# 0.7.0-DEV.3585
+let buf = IOBuffer()
+    printstyled(IOContext(buf, :color=>true), "foo", color=:red)
```
stevengj: This won't work on 0.6, because the `:color` attribute is new.  You have to `eval(Base, :(have_color = true))` IIRC.

ararslan: Why does it work on Linux and macOS in 0.6 though?

PR #483, opened 2018-02-03
Add bytesavailable

ararslan:
Ask and ye shall receive


stevengj:
(I just ran into this for JuliaLang/IJulia.jl#623

PR #484, opened 2018-02-03
Add occursin and find* methods replacing search and findin

nalimilan:
Add `OccursIn` predicate which replaces `findin` and all specific `findfirst`/`findnext`/`findlast`/`findprev` methods for specific types. These all call `search` or `findin`, but return `nothing` instead of `0` where appropriate. Note that there is an inconsistency with generic `find*` methods which already existed in Julia 0.6 and which continue to return `0`: separate `Compat.find*` methods will have to be introduced for them to avoid conflicts with Base (https://github.com/JuliaLang/Compat.jl/pull/457).

Also fix `in(::CartesianIndex, ::CartesianIndices)` bug uncovered by importing `in`.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1261,7 +1261,7 @@ end
 
 @static if VERSION < v"0.7.0-DEV.3025"
     import Base: convert, ndims, getindex, size, length, eltype,
-                 start, next, done, first, last
+                 start, next, done, first, last, in, tail
     export CartesianIndices, LinearIndices
 
     struct CartesianIndices{N,R<:NTuple{N,AbstractUnitRange{Int}}} <: AbstractArray{CartesianIndex{N},N}
@@ -1433,6 +1433,7 @@ else
     @eval const $(Symbol("@error")) = Base.$(Symbol("@error"))
 end
 
+# 0.7.0-DEV.3415
 if !isdefined(Base, :findall)
     const findall = find
     export findall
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -1509,6 +1510,59 @@ end
...
+    const occursin = OccursIn
+    export occursin
+
+    zero2nothing(x) = x == 0 ? nothing : x
+
+    findnext(r::Regex, s::AbstractString, idx::Integer) = search(s, r, idx)
+    findfirst(r::Regex, s::AbstractString) = search(s, r)
+    findnext(c::EqualTo{Char}, s::AbstractString, i::Integer) = zero2nothing(search(s, c.x, i))
+    findfirst(c::EqualTo{Char}, s::AbstractString) = zero2nothing(search(s, c.x))
+    findnext(b::EqualTo{<:Union{Int8,UInt8}}, a::Vector{<:Union{Int8,UInt8}}, i::Integer) =
```
nalimilan: I'm not sure it's actually a good idea to return `nothing` rather than `0` from these functions. That means that changing the type of the element or of the collection would return a different value when there's no match, which makes it hard to write generic code. So I guess we should return `0` for all methods, and provide `Compat.find*` methods which return `nothing`.

The same discussion applies to `findnext(c::EqualTo{Char}, s::AbstractString, i::Integer)`, but there its less clear since the method is less generic (less risk that types would change).

stevengj: I would suggest providing *only* the `Compat.find*` functions that return `nothing`.   No need to provide new functions that return 0.

nalimilan: Well, several methods return ranges, so we can override the Base functions for them: it sounds too bad to force people to litter their code with `Compat.` when we could avoid it. Only methods returning scalars are problematic.

Also, I've realized that `equalto` and `occursin` (from Compat) work with the functions from Base, they are just going to be slow. So we may as well provide the fast methods since we provide these types.

I've pushed a commit adding unexported functions which return `nothing`. Methods are still added to Base functions, and they return zero. That sounds simpler both in terms of implementation and in terms of description of the behavior (see additions to README).

PR #488, opened 2018-02-07
add Pkg3 Project.toml file

StefanKarpinski:


StefanKarpinski:
Passed on 0.6, no more broken on master than it was before this.

PR #491, opened 2018-02-14
Add GC.at-preserve

ararslan:
Ref discussion in https://github.com/JuliaLang/julia/pull/25616

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,7 +346,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `isabstract` and `isleaftype` are now `isabstracttype` and `isconcretetype`, respectively
   ([#23666], [#25496]).
 
-* `gc` and `gc_enable` are now `GC.gc` and `GC.enable`, respectively ([#25616]).
+* `gc` and `gc_enable` are now `GC.gc` and `GC.enable`, respectively. `Base.@gc_preserve`
+  is now `GC.@preserve` for versions where this is defined; `GC.@preserve` is a no-op on
+  earlier versions ([#25616]).
 
 * `endof` is now `lastindex` ([#25458]). (Note that `lastindex(A, n)` is not supported.)
 
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1486,8 +1486,8 @@ end
         using Base: gc
         const enable = Base.gc_enable
         @static if !isdefined(Base, Symbol("@gc_preserve"))
-            macro preserve(x)
-                esc(x)
+            macro preserve(args...)
+                esc(args[end])
             end
         else
             @eval const $(Symbol("@preserve")) = Base.$(Symbol("@gc_preserve"))
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1485,6 +1485,13 @@ end
     @eval module GC
         using Base: gc
         const enable = Base.gc_enable
+        @static if !isdefined(Base, Symbol("@gc_preserve"))
+            macro preserve(x)
+                esc(x)
+            end
+        else
+            @eval const $(Symbol("@preserve")) = Base.$(Symbol("@gc_preserve"))
+        end
     end
     export GC
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1319,7 +1319,7 @@ end
 @test !GC.enable(true)
 @test GC.enable(true)
 let x = 1
-    @test GC.@preserve(x) == 1
+    @test GC.@preserve(x, 2) == 2
 end
 
 @test eltype(Base.Multimedia.displays) <: AbstractDisplay
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1318,6 +1318,9 @@ end
 @test !GC.enable(false)
 @test !GC.enable(true)
 @test GC.enable(true)
+let x = 1
+    @test GC.@preserve(x) == 1
+end
 
 @test eltype(Base.Multimedia.displays) <: AbstractDisplay
 


src/Compat.jl
```diff
@@ -1485,6 +1485,13 @@ end
     @eval module GC
         using Base: gc
         const enable = Base.gc_enable
+        @static if !isdefined(Base, Symbol("@gc_preserve"))
+            macro preserve(x)
```
stevengj: seems like it should be `preserve(args...)` and return `esc(args[end])`, since the syntax is `GC.@preserve x1 x2 ... xn expr`

test/runtests.jl
```diff
@@ -1318,6 +1318,9 @@ end
 @test !GC.enable(false)
 @test !GC.enable(true)
 @test GC.enable(true)
+let x = 1
+    @test GC.@preserve(x) == 1
```
stevengj: Also add `@test GC.@preserve(x, 2) == 2` to test for the problem I mentioned above.

src/Compat.jl
ararslan: Oh right, whoops. Good catch.

yuyichao:
The optimization is very limited on 0.6 so the semantic should only be needed in rare cases. Unfortunately, I'm pretty sure dead code elimination is an optimization that 0.6 can do so there won't be any guarantee that the ccall's will do anything.

README.md
```diff
@@ -346,7 +346,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `isabstract` and `isleaftype` are now `isabstracttype` and `isconcretetype`, respectively
   ([#23666], [#25496]).
 
-* `gc` and `gc_enable` are now `GC.gc` and `GC.enable`, respectively ([#25616]).
+* `gc` and `gc_enable` are now `GC.gc` and `GC.enable`, respectively. `Base.@gc_preserve`
+  is now `GC.@preserve` for versions where this is defined; `GC.@preserve` is a no-op on
+  earlier versions ([#25616]).
```
omus: Needs the full link at the bottom of the page.

stevengj: (Just run the NEWS-update.jl script in the julia repo.)

omus: Good tip. I was unaware of this. Would be good if Compat had these kind of things mentioned in a CONTRIBUTING.md

stevengj:
Bump.   The argument for having this as a no-op on 0.6 is persuasive to me — in most cases, you only need this in 0.7, and not having this in Compat forces one to re-implement Compat in one's own code.

yuyichao:
>  forces one to re-implement Compat in one's own code.

Right. That's the whole point. There are actually correct way to do this most of the time on 0.6, just that it can't be made into a macro.

yuyichao:
In another word, defining this will encourage people to write wrong code on 0.6.
Not that I particularly care about 0.6 support, but you should either support it correctly or drop it, rather than pretending to support it but actually break it.

stevengj:
In practice, people just pass pointers to C functions in 0.6. Not defining this macro encourages them to do the same thing in 0.7 (since otherwise compatibility is too hard), even though it is more dangerous there. 

We want to encourage people to transition to doing the right thing.

yuyichao:
That'll be exactly why we shouldn't do this since if you are just passing pointers to `ccall`, the correct way on both 0.6 and 0.7 is to define the correct `unsafe_convert` and `cconvert`.

stevengj:
This macro exists because there are cases where that is not practical...

yuyichao:
Not really. The macro exists because there are cases where that is impossible and the unsafe code is not using ccall at all. The macro is of course nontheless useful for the general case even if the usecase is already supported by the ccall guarantee but it does not mean that it's what you should use if you want to claim you support 0.6.

stevengj:
(There seem to be at least 7 cases in base where `GC.@preserve` is used with a `ccall`.)

Prior to the introduction of this macro, base was simply doing nothing in most cases (e.g. your own commit https://github.com/JuliaLang/julia/commit/15ca594696b5823a9f7b7c32dad6b75b0699e320).  Although this was theoretically unsafe, in practice it seemed to work fine in 0.6 because of the less aggressive garbage collector, no?  

I don't see why you are so dead set on preventing user code from doing the same thing that Base did: "live dangerously" in 0.6 (in theory), and use `GC.@preserve` in 0.7.

yuyichao:
Most of the changes there were simply using `@gc_preserve` because it minimize the code changes and because the code is 0.7 only, if the code you are dealing with satisfies the same requirement, go ahead and use the new macro.

> doing the same thing that Base did: "live dangerously" in 0.6 (in theory if not in practice)

Those code are writen a long time ago and still exists in base. I'm by all mean **NOT** fine with them, never. And they are by all mean not safe on 0.6. There are just too many of them to be taken care of and that crashes caused by those are very hard to reproduce. For code that is already broken and if the user want to fix it properly it'll certainly be fine but that fix had better be a real fix on all supported versions.

vtjnash:
I think we can define:
```julia
@noinline preserve(args...) = nothing
macro preserve(args...)
    eargs = map(esc, args)
    return quote
        let v
        preserve($(eargs[2:end]...))
        v = $(eargs[end])
        preserve($(eargs[2:end]...))
        v
        end
    end
end
```
It's not quite right, of course, but should be almost enough to defeat the v0.6- gc placement algorithm. It wasn't actually possible to do this correctly before Keno gave us `GC.@preserve` in v0.7 anyways, so this is pretty much what you would have needed to write to maybe get the right behavior (as long as `args[end]` doesn't contain control flow to jump outside of the block – esp. `return`).

yuyichao:
That'll be the same as the `jl_breakpoint` one above but I believe dead code elimination is one of the few things that 0.6 can actually do.

ararslan:
If someone wants to come up with a proper way to do this on 0.6 then I encourage them to submit a PR, but I'm not familiar enough with the differences to do it myself, so I'll just close this.

oschulz:
I realize this has been closed, but speaking from a user perspective: IMHO, even a macro that does nothing on v0.6 would be better than no solution. Otherwise, people trying to get their packages ready for v0.7 (without dropping v0.6 support, since large parts of the Julia ecosystem are simply not ready for v0.7 yet) will either create exactly such a macro in their packages - or they will simply ignore the issue, leaving their code riddled with time bombs that'll go off on v0.7 at some point.

yuyichao:
That's super expensive but it should indeed be ok on 0.6. It will have different scope though.....

oschulz:
Would it be acceptable though? Even if there's a performance impact on v0.6, it would enable packages to do the "right thing" right now, and be prepared for 0.7. I would hope that in many cases, the code inside a `@gc_preserve` will do significant work, so that the cost of the `try`/`finally` (or a similar solution) would not be prohibitive (does the `try`/`finally` actually help preventing GC?).

If so, what would be a good generic "do something with that's cheap, but safe from dead code elimination" operation?

yuyichao:
Right, performance won't be a problem, scope difference is a little borderline.

oschulz:
But simply
```
macro gc_preserve(args...)
        $(args[end])
        # do something with with args[1:end-1] that's cheap but safe from dead code elimination 
end
```
(without `try`/`catch` and different scope) won't work?

Regarding the "do something with that's cheap, but safe from dead code elimination"  operation - would either of these (or something similar) do?

```
    @assert UInt(pointer_from_objref(obj)) > 0

    @assert object_id(obj) > 0
```

Or is the v0.6 compiler smart enough to realize that the condition will always be true?

vtjnash:
`object_id` is very expensive. The no-op `@noinline` function I posted above would be sufficient

vtjnash:
I think `pointer_from_objref` is probably also sufficient (the assert, etc. is not needed, as we don't mark this pure)

yuyichao:
> (without try/catch and different scope) won't work?

No it won't.

oschulz:
Darn ... but how can GC collect something that's used afterwards in a `@noinline` function or similar? I don't doubt your word on it, I'm just curious.

yuyichao:
The point is that if those are eliminated it doesn't matter what you put there.

oschulz:
I guess what I don't understand is how the Julia compiler is able to eliminate code if it cannot predict the outcome (e.g. with my `@assert`).

yuyichao:
Like basically all optimizations, this will of course only happen if the compiler is able to predict the outcome.....

oschulz:
E.g.:

function foo(obj)
    UInt(pointer_from_objref(obj)) != 0
    nothing
end

@code_llvm foo("Hello, World!")

function bar(obj)
    @assert UInt(pointer_from_objref(obj)) != 0
    nothing
end

@code_llvm bar("Hello, World!")
```

On Julia v0.6 (and v0.7, too), the comparison is eliminated for `foo()`, but not for `bar()`.

**Comment by oschulz** [comment]
While @assert is not free, I'd hope it would be cheap enough to be acceptable in a `@gc_preserve` compatibility macro.

**Comment by yuyichao** [comment]
As jameson said, `@assert` is not useful here. It's also now what DCE is applied in.

```
function bar(obj)
    return
    @assert UInt(pointer_from_objref(obj)) != 0
    nothing
end

function bar2(obj)
    throw(2)
    @assert UInt(pointer_from_objref(obj)) != 0
    nothing
end
```

oschulz:
Argh, I see. So there's basically no way around `try`/`catch`. Thanks for explaining, though!

oschulz:
Speaking from a purely practical point of view though, I do believe we need some compatibility macro to make people start using GC-preservation in their code at this stage. So the question is, for a future PR here, what would you judge the best solution?

* The safe implementation, with `try`/`catch`, accepting the different scoping on v0.6.
* Some solution with an `@assert` or `@noinline` that would work in some, but not all situations - but that in combination with the typical behaviour of the v0.6 GC might just be good enough until v0.7.
* An empty macro, as proposed above, because it wouldn't be worse than what people usually do right now.

oschulz:
Sorry for bothering you about the DCE, @yuyichao - should have realized that it's the whole guard code construct that might get eliminated.

yuyichao:
The basic requirement that everything in Compat should have is that the definition with the right name should do the right thing. We do accept exception but in those cases the Compat definition should throw an error. In another word, the definition must do the same thing or error.

Note that this is very different from what a package may actually want. For this particular example, I do believe most of the package will not hit most of the corner cases, but I do believe all of the corner cases mentioned in this thread will be hit by some packages. That's what make it sometimes a better solution for the package to just define it's own version that satisfy only it's need if it won't be affected by any special cases.

The different in scope is actually a major difference but I said it's a borderline case since that case is actually detectable in certain cense. If the `Compat.@gc_preserve` for 0.6 is written so that it only accept `while`, `for`, `let` or `try` blocks as the last argument this will fully satisfy the requirement of doing the right thing or error.

> because it wouldn't be worse than what people usually do right now.

This is the argument I don't buy because there is actually well defined way to do this correctly on 0.6, just that those solutions can't be made into a `@gc_preserve` macro. Most of the cases that do simple C-interop should just use `cconvert` and `unsafe_convert`, on both 0.6 and 0.7 as mentioned above. I'm pretty sure I've also mentioned this in another thread on this topic that the most reliable way to do this on 0.6 is to use a noninline function since the compiler will always make sure the argument is rooted inside the function, at least on 0.6.

oschulz:
> If the Compat.@gc_preserve for 0.6 is written so that it only accept while, for, let or try blocks as the last argument this will fully satisfy the requirement of doing the right thing or error.

I guess that would cover most use cases. And I do understand the need to be rigorous about correctness in Compat.

> Most of the cases that do simple C-interop [...]

Yes, I saw that - I was actually concerned about non C-related use cases here. But I have to admit those are not so common.

> argument is rooted inside the function, at least on 0.6.

So an argument (of a noinline-function) will not be GC'ed during the function on 0.6? Thanks, I had been wondering about that (my question on Discourse).

PR #492, opened 2018-02-16
Markdown stdlib

fredrikekre:


ararslan:
Gotta love the CI situation. Red across the board, all unrelated.

README.md
```diff
@@ -141,6 +141,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `using Compat.UUIDs` is provided on versions older than 0.7, where this library is
   not yet part of the standard library ([#25819]).
 
+* `using Compat.Markdown` is provided on versions older than 0.7, where this library is
+  not yet part of the standard library ([#25738]).
```
omus: Needs the full link added at the bottom of the page. Currently is a broken link.

ararslan: We need a script here that does what `news-update.jl` or whatever it's called does in Base.

omus: That would be good. It would also be good to add this as a CI check.

fredrikekre: ran the script in #493

PR #493, opened 2018-02-16
names with kwargs

fredrikekre:


src/Compat.jl
```diff
@@ -1629,6 +1629,11 @@ else
     findall(b::OccursIn, a::Number) = a in b.x ? [1] : Vector{Int}()
 end
 
+# https://github.com/JuliaLang/julia/pull/25647
+@static if VERSION < v"0.7.0-DEV.3526"
+    Base.names(m; all=true, imported=true) = names(m, all, imported)
```
ararslan: Is it kosher to overwrite Base methods like this? We could provide an unexported `Compat.names` with the appropriate signatures.

fredrikekre: Hmm.... right, this overwrites the one arg function? Lets just do the unexported thing then.

src/Compat.jl
```diff
@@ -1643,6 +1643,13 @@ else
     findall(b::OccursIn, a::Number) = a in b.x ? [1] : Vector{Int}()
 end
 
+# https://github.com/JuliaLang/julia/pull/25647
+@static if VERSION < v"0.7.0-DEV.3526"
+    names(m; all=true, imported=true) = names(m, all, imported)
```
ararslan: The method that it calls might need a `Base.` qualification.

README.md
```diff
@@ -274,6 +274,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `firstindex` to obtain the first index of an iterable ([#25458]).
 
+* `Compat.names` supporting keyword arguments for `all` and `imported` ([#25647]).
```
omus: Do you need to add the full link at the bottom of the page?

src/Compat.jl
```diff
@@ -1643,6 +1643,13 @@ else
     findall(b::OccursIn, a::Number) = a in b.x ? [1] : Vector{Int}()
 end
 
+# https://github.com/JuliaLang/julia/pull/25647
+@static if VERSION < v"0.7.0-DEV.3526"
+    names(m; all=true, imported=true) = Base.names(m, all, imported)
```
mortenpi: Should this not be `names(m; all=false, imported=false)` to be [in line with 0.7](https://github.com/JuliaLang/julia/blob/16620b62443f0ffd3aa5fa58174b5d66a28cc646/base/reflection.jl#L86-L99)?

This behaves differently on 0.6 and 0.7 right now:

```julia
import Compat
module Foo
    import Base: copy
end
Compat.names(Foo)
```

```julia-repl
julia-0.7> Compat.names(Foo)
1-element Array{Symbol,1}:
 :Foo

julia-0.6> Compat.names(Foo)
4-element Array{Symbol,1}:
 Symbol("#eval")
 :Foo
 :copy
 :eval
```

stevengj: Fixed by #505.

PR #497, opened 2018-02-22
rename reprmime and mimewritable

stevengj:
This PR implements the renames `reprmime` -> `repr` (JuliaLang/julia#25990) and `mimewritable` -> `showable` (JuliaLang/julia#26089).

JuliaLang/julia#25990 also moved `stringmime` into the `Base64` package, but `Compat.Base64.stringmime` should work as-is.

(In 0.7, `repr` also supports a new `context` keyword, but that feature is not implemented by this PR.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Display` is now `AbstractDisplay` ([#24831]).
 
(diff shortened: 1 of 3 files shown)


stevengj:
Unrelated download error on Windows. Should be good to merge.

PR #498, opened 2018-02-23
add devnull, stdin, stdout and stderr

fredrikekre:


fredrikekre:
Hmm... This will not work becuse of functions changing `Base.stdout` but `Compat.stdout` will always point to the old one.

stevengj:
Yes, it needs to be done with `@compat` to rewrite at the source level, I think.

yuyichao:
Not necessarily.

```julia
julia> module A
       a = 1
       end
Main.A

julia> module B
       import ..A.a
       end
Main.B

julia> B.a
1

julia> A.a
1

julia> A.eval(:(a = 2))
2

julia> B.a
2

julia> A.a
2
```

yuyichao:
Actually nvm, I guess this doesn't work for renaming...

PR #499, opened 2018-02-23
stdin/stdout/stderr and devnull

stevengj:
Updated version of #498 to use ~~`@compat`~~ `_redirect_*` for JuliaLang/julia#25959.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -4,6 +4,11 @@ module Compat
 
 include("compatmacro.jl")
 
+@static if !isdefined(Base, :devnull) #25959
+     export devnull
+     const devnull = DevNull
+end
+
 @static if !isdefined(Base, Symbol("@nospecialize"))
     # 0.7
     macro nospecialize(arg)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -17,6 +17,7 @@ include("compatmacro.jl")
             function Base.$rf(stream::IO)
                 invoke(Base.$rf, Tuple{Any}, stream)
                 global $f = $F
+                nothing
             end
         end
     end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -15,9 +15,9 @@ include("compatmacro.jl")
             # overload internal _redirect_std* functions
             # so that they change Compat.std*
             function Base.$rf(stream::IO)
-                invoke(Base.$rf, Tuple{Any}, stream)
+                ret = invoke(Base.$rf, Tuple{Any}, stream)
                 global $f = $F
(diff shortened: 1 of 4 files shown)


stevengj:
CI failure on Windows is unrelated download error.

src/Compat.jl
```diff
@@ -2,7 +2,12 @@ __precompile__()
 
 module Compat
 
-using Base.Meta
+include("compatmacro.jl")
+
+@static if !isdefined(Base, :devnull) #25959
+     export devnull
```
ararslan: 5 space indent?

ararslan:
@mbauman suggested that we override `Base._redirect_*` here, which seems like it could be a cleaner solution.

stevengj:
Updated to use `_redirect_*`, which should work for anything that doesn't explicitly mutate `STDOUT` (like IJulia does — it will just need a `VERSION` check).

I kept the change moving `@compat` to a separate file, since it still seems cleaner to separate this out, but the macro is unchanged.

stevengj:
Usual unrelated download error on AppVeyor

martinholters:
Any idea what's up here?
```julia
julia> VERSION
v"0.6.0"

julia> using Compat: stderr

julia> print(STDERR, "\r")

julia> print(stderr, "\r")

signal (11): Segmentation fault
while loading no file, in expression starting on line 0
uv_write2 at deps/srccache/libuv-52d72a52cc7ccd570929990f010ed16e2ec604c8/src/unix/stream.c:1389
uv_write at deps/srccache/libuv-52d72a52cc7ccd570929990f010ed16e2ec604c8/src/unix/stream.c:1475
jl_uv_write at src/jl_uv.c:414
uv_write at ./stream.jl:798
unsafe_write at ./stream.jl:832
print at ./strings/io.jl:122
unknown function (ip: 0x7f7c5b6e9556)
jl_call_fptr_internal at src/julia_internal.h:339 [inlined]
jl_call_method_internal at src/julia_internal.h:358 [inlined]
jl_apply_generic at src/gf.c:1933
do_call at src/interpreter.c:75
```
(Yea, I'll give 0.6.2 a try...)

martinholters:
Confirmed on 0.6.2. Also note:
```julia
julia> using Compat: stderr

julia> stderr === STDERR
false
```
Should the assignment happen in `__init()__`?

stevengj:
I guess so?

stevengj:
Since this is an urgent bugfix, I pushed it directly: e39939a8f172ae6503c99a7a87b35b3d074229b3

PR #501, opened 2018-02-24
Add IOBuffer keyword support

TotalVerb:
cc @ararslan

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1304,7 +1304,7 @@ enable_debug(x::Bool) = DEBUG[] = x
 @static if !isdefined(Base, Symbol("@debug"))
     function debug(msg)
         DEBUG[] || return
-        buf = IOBuffer()
+        buf = Base.IOBuffer()
         iob = Base.redirect(IOContext(buf, STDERR), Base.log_info_to, :debug)
         print_with_color(:blue, iob, "Debug: "; bold = true)
         Base.println_with_color(:blue, iob, chomp(string(msg)))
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -1559,6 +1559,37 @@ else
...
+    function IOBuffer(
+            data::Union{AbstractVector{UInt8},Nothing}=nothing;
+            read::Union{Bool,Nothing}=nothing,
+            write::Union{Bool,Nothing}=nothing,
+            truncate::Union{Bool,Nothing}=nothing,
+            maxsize::Integer=typemax(Int),
+            sizehint::Union{Integer,Nothing}=nothing)
+        write    === nothing && (write    = false)
+        read     === nothing && (read     = !write)
+        truncate === nothing && (truncate = false)
```
stevengj: When `data === nothing` these should default to `read=write=true`.

See https://github.com/JuliaLang/julia/blob/79c7bdd9ecf20bcfb685a0abb74b028f72ccf70c/base/iobuffer.jl#L105-L106

src/Compat.jl
```diff
@@ -1559,6 +1559,37 @@ else
...
+        if maxsize < 0
+            throw(ArgumentError("negative maxsize: $(maxsize)"))
+        end
+        if sizehint !== nothing
+            sizehint!(data, sizehint)
```
stevengj: This is wrong is `data === nothing`

src/Compat.jl
```diff
@@ -1559,6 +1559,37 @@ else
...
+        end
+        buf = if data !== nothing
+            Base.IOBuffer(data, read, write, Int(maxsize))
+        else
+            size = maxsize == typemax(Int) ? 32 : Int(maxsize)
```
stevengj: Use `sizehint` here if it is available: https://github.com/JuliaLang/julia/blob/79c7bdd9ecf20bcfb685a0abb74b028f72ccf70c/base/iobuffer.jl#L111

stevengj:
Needs a PR to fix the above issues.

Issue #502, opened 2018-02-26
strange behavior from IOBuffer


ExpandingMan:
The `Compat` version of `IOBuffer` does not seem to be working properly on 0.6.  If I do
```julia
b = IOBuffer(zeros(UInt8, 4), write=true)
write(b, 'a')

take!(b)
```
On the last line on 0.7 I get simply `[0x61]` (expected behavior).  On 0.6 however I get the entire buffer: `[0x61,0x00,0x00,0x00]`.

The problem is even worse if you do `IOBuffer(maxsize=4, write=true)` in which case in 0.6 all the values of an uninitialized array are returned.

Perhaps [this line](https://github.com/JuliaLang/Compat.jl/blob/master/src/Compat.jl#L264) is calling the wrong function?

stevengj:
I think #501 is buggy, as I mentioned at the end of that PR...

stevengj:
I'm preparing a PR to fix IOBuffer, including this issue.

ExpandingMan:
This was resolved by #504.

PR #504, opened 2018-02-27
Compat.IOBuffer bugfixes

stevengj:
Fixes some bugs in #501: incorrect handling of `sizehint` when `data===nothing` and incorrect logic for the `truncate` flag (#502).

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1567,29 +1567,49 @@ end
 
 # https://github.com/JuliaLang/julia/pull/25872
 if VERSION < v"0.7.0-DEV.3734"
+    if isdefined(Base, :open_flags)
+        import Base.open_flags
+    else
+        # copied from Base:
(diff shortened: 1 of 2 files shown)


src/Compat.jl
```diff
@@ -1567,29 +1567,49 @@ end
...
+            append   === nothing && (append   = false)
+            return (read, write, create, truncate, append)
+        end
+    end
     function IOBuffer(
             data::Union{AbstractVector{UInt8},Nothing}=nothing;
-            read::Union{Bool,Nothing}=nothing,
-            write::Union{Bool,Nothing}=nothing,
-            truncate::Union{Bool,Nothing}=nothing,
+            read::Union{Bool,Nothing}=data===nothing?true:nothing,
```
ararslan: 0.7 deprecated the omission of spaces around `?` and `:`, so this causes deprecation warnings. Fixed in #508.

PR #508, opened 2018-03-01
Add spaces in ternaries

ararslan:
Omitting the spaces in ternaries was deprecated in 0.7 so `using Compat` emits deprecation warnings now on master.


omus:
Always be spacing

PR #511, opened 2018-03-05
Add range with keyword arguments

ararslan:
Ref https://github.com/JuliaLang/julia/pull/25896

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.IOBuffer` supporting keyword arguments ([#25873]).
 
-* `Compat.range` supporting keyword arguments ([25896]).
+* `Compat.range` supporting keyword arguments ([#25896]).
 
 ## Renaming
 
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.IOBuffer` supporting keyword arguments ([#25873]).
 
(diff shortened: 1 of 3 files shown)


ararslan:
Hm, the test failure is confusing.

src/Compat.jl
```diff
@@ -1620,6 +1620,35 @@ if VERSION < v"0.7.0-DEV.3734"
...
+        if !(have_stop || have_length)
+            throw(ArgumentError("At least one of `length` or `stop` must be specified"))
+        elseif have_step && have_stop && have_length
+            throw(ArgumentError("Too many arguments specified; try passing only one of `stop` or `length`"))
+        elseif start === nothing
+            throw(ArgumentError("Can't start a range at `nothing`"))
+        end
+
+        if have_stop && !have_length
+            return have_step ? start:step:stop : start:stop
```
stevengj: Need parens here.   Should be `have_step ? (start:step:stop) : (start:stop)` but actually parses as `have_step ? start : ((step:stop:start):stop)`

README.md
```diff
@@ -278,6 +278,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.IOBuffer` supporting keyword arguments ([#25873]).
 
+* `Compat.range` supporting keyword arguments ([25896]).
```
stevengj: Needs `#` in issue link.

src/Compat.jl
ararslan: image

PR #513, opened 2018-03-08
add find(first|last|next|prev) methods that return nothing

fredrikekre:


src/Compat.jl
```diff
@@ -1498,6 +1498,7 @@ else
     export occursin
 
     zero2nothing(x::Integer) = x == 0 ? nothing : x
+    zero2nothing(x::AbstractUnitRange{<:Integer}) = x == 0:-1 ? nothing : x
```
nalimilan: Maybe use `isempty` just in case?

test/runtests.jl
```diff
@@ -1378,18 +1378,14 @@ for (f1, f2, i) in ((Compat.findfirst, Compat.findnext, 1),
         @test f2(occursin(chars), "bx", i) == f1(occursin(chars), "bx") == nothing
     end
 end
-for (f1, f2, i) in ((findfirst, findnext, 1),
```
nalimilan: I think we should keep testing the unprefix methods as Compat provides a few ones (which do not return `nothing`).

README.md
```diff
@@ -355,6 +355,10 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `Compat.findfirst`, `Compat.findnext`, `Compat.findlast` and `Compat.findprev`,
   return `nothing` when no match is found (rather than `0`) as on Julia 0.7 ([#24673]).
 
+* `Compat.findfirst`, `Compat.findnext`, `Compat.findlast` and `Compat.findprev`,
+  return `nothing` when no match is found (rather than the empty range `0:-1`)
```
stevengj: If you are searching for a particular element they return `0` on 0.6, not `0:-1`.

src/Compat.jl
stevengj: I would use `ifelse(iszero(first(x)), nothing, x)`

README.md
fredrikekre: Methods returning 0 were added in #484, see the bullet just above. Perhaps I can collapse them to the same bullet.

stevengj: Oh, I see.  Yes, please combine the bullets.

test/runtests.jl
fredrikekre: But they return nothing on 0.7. In fact this block of code does not pass on master.

src/Compat.jl
nalimilan: IMHO that would be harder to read for no real advantage. Also I think the compiler is smart enough to avoid branches were it's faster even if you use `?` (at least that's what Jameson noted somewhere).

test/runtests.jl
nalimilan: Ah, OK. Should probably have a version-specific test then, as long as we provide these methods via `Base.findfirst`, etc.

fredrikekre: What is the point in providing Compat methods which do not behave like 0.7 though?

src/Compat.jl
stevengj: Fine, no big deal either way.

test/runtests.jl
fredrikekre: I guess none of
https://github.com/JuliaLang/Compat.jl/blob/fd3b92ad49bda4c7c02b1494b3a79ef69c30a367/src/Compat.jl#L1508-L1540
should extend the Base functions. I think we should deprecate those and require `Compat.find*` such that people have the 0.7 behaviour w.r.t `nothing`?

nalimilan: IIRC these were added before they returned `nothing` on 0.7. Removing them would be problematic because it would break packages that use the new forms (i.e. those which have been ported before they returned `nothing`) even on 0.6. So I'd say it's simple for everybody to keep them.

fredrikekre: They don't behave the same on 0.6 and 0.7, thats the problem. So packages can't possibly rely on these methods and expect it to work the same on both versions.
0.6:
```julia
julia> using Compat

julia> findfirst(r"a", "b")
0:-1
```
and on 0.7:
```julia
julia> findfirst(r"a", "b")
nothing
```

nalimilan: They can't currently, but a few weeks ago the behavior was the same on 0.6 and 0.7, so some packages may have started using these methods. If any package does that, removing these functions in the next Compat release will break it on Julia 0.6, which would be really bad. If you really want to remove them, you can deprecate them (in Compat), but I don't think it's worth the pain. We can live with these until we drop 0.6 support.

Also as I noted at https://github.com/JuliaLang/Compat.jl/pull/484#discussion_r165821701, some of these methods are optimizations for 0.7 syntax that does work on 0.6 (using `equalto` and `occursin` which are exported from Compat), but are very slow compared to their standard 0.6 and 0.7 counterparts. Without these, people following deprecation messages would end up making their code much slower on 0.6.

fredrikekre: I don't get it, now we basically just provide another 0.6 interface to those functions. So the options here are 1. deprecate those to `Compat.find*` methods or 2. add the same methods but unexported that return `nothing`. I definitely think we should go with 2. It is very confusing to provide `findfirst(...)` and `Compat.findfirst()` with the same signatures, where only one of them have 0.7 behaviour.

nalimilan: Doesn't 2. already work?

fredrikekre: Yea you are right, it does, but I still think it is confusing. The Base extended methods are only useful on 0.6, at which point you might use the 0.6 syntax (e.g. with `search` and `rsearch` etc.)

nalimilan: Yes, I'm not saying they are really useful, but that deprecating them is more trouble than it's worth.

fredrikekre: Fair enough.

nalimilan:
Woops, sorry, my `isempty` suggestion was actually incorrect since e.g. `findfirst(r"", "a") == 1:0` even on 0.7. I've reverted the change directly.

PR #517, opened 2018-03-16
Fixes for JuliaLang/julia/pull/26436.

fredrikekre:
- Add one argument "curried" forms of `isequal`, `==` and `in`.

- Keep the following definitions to not break Compat on julia v0.6:
   - `const EqualTo{T} = Fix2{typeof(isequal),T}`
   - `equalto(x) = isequal(x)`
   - `const OccursIn{T} = Fix2{typeof(in),T}`
   - `occursin(x) = in(x)`

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -868,27 +868,39 @@ else
     import Serialization
 end
 
-# 0.7.0-DEV.1993
-@static if !isdefined(Base, :EqualTo)
-    if VERSION >= v"0.6.0"
-        include_string(@__MODULE__, """
-            struct EqualTo{T} <: Function
-                x::T
-
-                EqualTo(x::T) where {T} = new{T}(x)
-            end
-        """)
-    else
-        include_string(@__MODULE__, """
-            immutable EqualTo{T} <: Function
-                x::T
-            end
-        """)
+@static if VERSION < v"0.7.0-DEV.4592"
+    struct Fix2{F,T} <: Function
+        f::F
+        x::T
+        Fix2(f::F, x::T) where {F,T} = new{F,T}(f, x)
+        Fix2(f::Type{F}, x::T) where {F,T} = new{F,T}(f, x)
     end
-    (f::EqualTo)(y) = isequal(f.x, y)
-    const equalto = EqualTo
-    export equalto
+    (f::Fix2)(y) = f.f(y, f.x)
+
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -868,28 +868,27 @@ else
...
+
+    Base.isequal(x) = Fix2(isequal, x)
+    Base.:(==)(x) = Fix2(==, x)
+    Base.in(x) = Fix2(in, x)
+end
+# # keep these definitions to be non breaking for 0.6 usage
+@static if VERSION < v"0.7.0-DEV.1993"
+    const EqualTo{T} = Fix2{typeof(isequal),T}
     export equalto
+    equalto(x) = isequal(x)
```
nalimilan: That won't work AFAICT, you need to keep the old code for `EqualTo`. It would also be nice to have a test to cover this. Same for `occursin`.

src/Compat.jl
```diff
@@ -1486,14 +1485,10 @@ end
...
 else
-    struct OccursIn{T} <: Function
-        x::T
-
-        OccursIn(x::T) where {T} = new{T}(x)
-    end
-    (f::OccursIn)(y) = y in f.x
-    const occursin = OccursIn
+    # keep these definitions to be non breaking for 0.6 usage
+    const OccursIn{T} = Fix2{typeof(in),T}
```
nalimilan: Like `equalto`, should probably only be defined on 0.6 and older 0.7 versions, but print deprecations on recent 0.7 versions? Else people won't notice it's gone.

nalimilan:
Forget my previous comments, I had completely misread the VERSION checks. Looks fine to me (though I haven't followed https://github.com/JuliaLang/julia/pull/26436 very closely). However, the OS X failure seems to be real, and might indicate that the code doesn't work with 0.7 versions before https://github.com/JuliaLang/julia/pull/26436 was merged.

fredrikekre:
Passes locally now for v0.6, master and master just before JuliaLang/julia/pull/26436

fredrikekre:
> Looks good if you also confirm `equalto` and `occursin` continue to work on 0.6 for packages which had already been ported to them.

Added tests for this.

garrison:
It would be nice if `Compat.Fix2` existed regardless of julia version.  Right now I have to do `Fix2 = isdefined(Compat, :Fix2) ? Compat.Fix2 : Base.Fix2`.

PR #518, opened 2018-03-19
Add functions with `dims` keyword argument

martinholters:
Add (unexported) `Compat.accumulate`, `Compat.accumulate!`,
`Compat.all`, `Compat.any`, `Compat.cor`, `Compat.cov`,
`Compat.cumprod`, `Compat.cumprod!`, `Compat.cumsum`, `Compat.cumsum!`,
`Compat.findmax`, `Compat.findmin`, `Compat.mapreduce`,
`Compat.maximum`, `Compat.mean`, `Compat.median`, `Compat.minimum`,
`Compat.prod`, `Compat.reduce`, `Compat.sort`, `Compat.std`,
`Compat.sum`, `Compat.var`, and `Compat.varm` with `dims` keyword
argument.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -513,7 +513,7 @@ if VERSION < v"0.7.0-DEV.755"
     # This is a hack to only add keyword signature that won't work on all julia versions.
     # However, since we really only need to support a few (0.5, 0.6 and early 0.7) versions
     # this should be good enough.
-    let Tf = typeof(cov), Tkw = Core.Core.kwftype(Tf)
+    let Tf = typeof(Base.cov), Tkw = Core.Core.kwftype(Tf)
         @eval begin
             @inline function _get_corrected(kws)
                 corrected = true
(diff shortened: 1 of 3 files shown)


martinholters:
If we wait for https://github.com/JuliaLang/julia/pull/26521, I can put the commented tests behind `VERSION` conditionals, exempting only the version range where Julia was broken. (Of course, we can also merge this now and fix up the tests later.)

fredrikekre:
I added the version check. I didn't verify this, but you implied in the julia issue that it was broken by https://github.com/JuliaLang/julia/pull/25989?

src/Compat.jl
```diff
@@ -1690,6 +1690,72 @@ if VERSION < v"0.7.0-DEV.4585"
...
+        dims===nothing ? Base.cor(a, b) : Base.cor(a, b, dims)
+    @eval mapreduce(f, op, a::AbstractArray; dims=nothing) =
+        dims===nothing ? Base.mapreduce(f, op, a) : Base.mapreducedim(f, op, a, dims)
+    @eval mapreduce(f, op, v0, a::AbstractArray; dims=nothing) =
+        dims===nothing ? Base.mapreduce(f, op, v0, a) : Base.mapreducedim(f, op, a, dims, v0)
+    @eval reduce(op, a::AbstractArray; dims=nothing) =
+        dims===nothing ? Base.reduce(op, a) : Base.reducedim(op, a, dims)
+    @eval reduce(op, v0, a::AbstractArray; dims=nothing) =
+        dims===nothing ? Base.reduce(op, v0, a) : Base.reducedim(op, a, dims, v0)
+    @eval accumulate!(op, out, a; dims=nothing) =
```
fredrikekre: I guess these `@eval` could be removed? Only the ones in the loops over function symbols is needed AFAICT.

martinholters:
> I added the version check.

Thanks, but I've actually bisected the origin to an earlier change; the correct lower bound would have to be v"0.7.0-DEV.3262". I'll update accordingly.

src/Compat.jl
martinholters: Good catch. I'll update.

martinholters: Um, no, that gives
```
ERROR: LoadError: error compiling anonymous: error in method definition: function Base.accumulate! must be explicitly imported to be extended
```
There is no other use `accumulate!` though, so why doesn't it just create a new binding to a new function? BUT if I move the definition to its own `if` block, it does work without the `@eval`. Strange...

fredrikekre: Yea I tried the same, lets just leave it as is I guess.

PR #521, opened 2018-03-21
Make Compat.Fix2 always exist

garrison:
Following https://github.com/JuliaLang/Compat.jl/pull/517#issuecomment-374675232

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -888,6 +888,8 @@ end
     else
         Base.in(x) = Fix2(in, x)
     end
+else
+    import Base: Fix2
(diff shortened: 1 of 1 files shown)


fredrikekre:
Just out of curiosity, what is the usecase for this?

garrison:
Implementing `find*` methods for [UniqueVectors](https://github.com/garrison/UniqueVectors.jl/blob/master/src/UniqueVectors.jl).  See also: [here](https://github.com/JuliaLang/julia/pull/23812#issuecomment-331637483).

PR #522, opened 2018-03-21
Add `selectdim`

martinholters:


martinholters:
Anyone care to take a look here? The code (but not the tests) is copied from Base, so I'm rather confident it's ok, but a second pair of eyes surely wouldn't hurt...

PR #529, opened 2018-04-09
Fix tests for current Julia nightly

martinholters:
The first commit replaces calls to `CartesianIndices(1:3, 1:2)` with `CartesianIndices((Base.OneTo(3), Base.OneTo(2)))`. The former now gives an error when used for indexing, e.g.:
```julia
julia> zeros(3, 2)[CartesianIndices((1:3, 1:2))]
ERROR: MethodError: no method matching similar(::Array{Float64,2}, ::Type{Float64}, ::Tuple{UnitRange{Int64},UnitRange{Int64}})
```
Is that supposed to happen, @mbauman?

The second commit removes the test`findfirst` on `IOBuffer` as Julia removed support for it in https://github.com/JuliaLang/julia/pull/26600. Should we also add a deprecated method to `findfirst`, using the definition of the deprecated `search`, i.e. https://github.com/JuliaLang/julia/blob/278a2893b8b02d6063ee4b1717c201c3431cb042/base/deprecated.jl#L1565-L1571 ?

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1633,13 +1633,11 @@ else
 end
 
 # compatibiltiy with https://github.com/JuliaLang/julia/pull/26156
-if VERSION < v"0.7.0-DEV.4062" || VERSION >= v"0.7.0-DEV.4804"
-    trunc(x, digits; base = base) = trunc(x, digits = digits, base = base)
-    floor(x, digits; base = base) = floor(x, digits = digits, base = base)
-    ceil(x, digits; base = base) = ceil(x, digits = digits, base = base)
-    round(x, digits; base = base) = round(x, digits = digits, base = base)
-    signif(x, digits; base = base) = round(x, sigdigits = digits, base = base)
-end
+trunc(x, digits; base = base) = trunc(x, digits = digits, base = base)
+floor(x, digits; base = base) = floor(x, digits = digits, base = base)
+ceil(x, digits; base = base) = ceil(x, digits = digits, base = base)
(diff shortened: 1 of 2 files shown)


martinholters:
Looks like I should have checked on _current_ current Julia nightly... Will look into the remaining failures.

Meanwhile CC @nalimilan for the `findfirst` thing.

nalimilan:
Thanks. I don't think we should add a deprecation to `findfirst` given that we only define the method on Julia 0.6. If we deprecated it there packages would start printing annoying warnings even for users who don't care about 0.7.

Regarding `CartesianIndices`, why not use `CartesianIndices((1:3, 1:3))` instead of `OneTo`?

martinholters:
> Regarding `CartesianIndices`, why not use `CartesianIndices((1:3, 1:3))` instead of `OneTo`?

https://github.com/JuliaLang/julia/pull/26682#issuecomment-379762632

martinholters:
The remaining failure is not just a problem with the tests. So I propose we merge this as is, which will enable a future PR fixing `round` and friends to actually see its tests run and get a green CI.

martinholters:
> I'd rather wait until we have a clearer view of what we want to do, or disable/remove the `map(i -> l[i], c)` test

Yep, will do. But note that it's also the `l[c]` that fails.

martinholters:
Ah, test failures on the out-of-data Julia nightlies are due to a mistake in #530. I'll add the fix to this PR to verify I get CI green...

martinholters:
Good to go?

PR #533, opened 2018-04-16
RFC: Set version in Project.toml to 0.64.0

martinholters:
Ref. https://github.com/JuliaLang/Pkg3.jl/issues/248

The idea is to (squash and) merge this PR just before tagging the next release and then tag that commit.

We don't need to start doing this now, but we need to agree on a workflow eventually. So I'm also opening this to start a discussion: Should we go with a version-bumping PR for every release? Or should those with the power to do so just push to master and tag the release? Or what other alternatives do we have?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
+version = "v0.64.0"
 
 [deps]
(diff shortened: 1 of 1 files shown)


martinholters:
No opinions on this, anyone?

martinholters:
If no-one is against this, I'll merge tomorrow and tag a release.

KristofferC:
Sorry, I hadn't seen this. :+1: here.

PR #534, opened 2018-04-20
Add `qr` with `pivot` as `Val` instance and `full` keyword

martinholters:
Test pass locally on 0.7 when commenting out the (unrelatedly) failing tests (see #529).


martinholters:
Barring objections I'll merge tomorrow.

PR #537, opened 2018-04-26
Set default base for round to 10

jmkuhn:


jmkuhn:
The same issue exists for `floor`, `ceil`, `trunc`.  I'll take a look at those now.

jmkuhn:
The AppVeyor build failure is fixed in Julia version 0.7.0-DEV.4942.

jmkuhn:
Fixes #567.

src/Compat.jl
```diff
@@ -1594,10 +1594,10 @@ end
...
 # https://github.com/JuliaLang/julia/pull/26670
 @static if VERSION < v"0.7.0-DEV.4062"
-    trunc(x; digits = digits, base = base) = Base.trunc(x, digits, base)
-    floor(x; digits = digits, base = base) = Base.floor(x, digits, base)
-    ceil(x; digits = digits, base = base) = Base.ceil(x, digits, base)
-    function round(x; digits = nothing, sigdigits = nothing, base = base)
+    trunc(x; digits = 0, base = 10) = Base.trunc(x, digits, base)
+    floor(x; digits = 0, base = 10) = Base.floor(x, digits, base)
+    ceil(x; digits = 0, base = 10) = Base.ceil(x, digits, base)
+    function round(x; digits = nothing, sigdigits = nothing, base = 10)
```
omus: Maybe these should be constants?

martinholters:
Should probably also test cases where `base` is given but `digits` is not, and where neither is given.

jmkuhn:
I'm looking at the Irrational issue now.

jmkuhn:
```
julia> Compat.round(1.2, RoundNearest)
ERROR: MethodError: no method matching round(::Float64, ::RoundingMode{:Nearest}, ::Int64)
Closest candidates are:
  round(::Real, ::Integer, ::Integer) at floatfuncs.jl:152
  round(::Float64) at float.jl:352
  round(::Complex{#s45} where #s45<:AbstractFloat, ::RoundingMode{MR}, ::RoundingMode{MI}) where {MR, MI} at complex.jl:885
```
Need to add additional methods to `Compat.round`.

martinholters:
This has gone a bit stale, so I'll close/reopen to rerun CI. If that passes, should be good to go.

martinholters:
Test failures are because on 0.7, we have:
```julia
julia> round(pi, base=10)
ERROR: MethodError: no method matching round(::Irrational{:π}, ::RoundingMode{:Nearest})
Closest candidates are:
  round(::Irrational, ::RoundingMode) at irrationals.jl:137
  round(::Real, ::RoundingMode; digits, sigdigits, base) at floatfuncs.jl:127
  round(::Number, ::Any) at deprecated.jl:53
```
(Already noted above; I had assumed this had been fixed in the meantime.) Ironically:
```julia
julia> round(pi, RoundingMode{:Nearest}())
3.0
```
The method error above is explicitly thrown in floatfuncs.jl:
```julia
# if we hit this method, it means that no `round(x, r)` method is defined
_round(x::Real, r::RoundingMode, digits::Nothing, sigdigits::Nothing, base) = throw(MethodError(round, (x,r)))
```
That's a genuine issue in base. To get this PR going, I'd propose to just replace `pi` with `3.14159`.

martinholters:
This has gone a bit stale, so I'll close/reopen to rerun CI. If that passes, should be good to go.

(Yes, I've said this before.)

Issue #540, opened 2018-05-07
Wrong `IteratorSize` for `pairs(::IndexCartesian, ...)`


martinholters:
On Julia master:
```julia
julia> Base.IteratorSize(pairs(IndexCartesian(), zeros(2,2)))
Base.HasShape{2}()
```
On Julia 0.6 with Compat:
```julia
julia> Compat.IteratorSize(Compat.pairs(IndexCartesian(), zeros(2,2)))
Base.HasLength()
```

We implicitly test for the latter behavior, making our tests fail on 0.7. We probably want to do
```diff
diff --git a/test/runtests.jl b/test/runtests.jl
index 4201f5fe..05bbf174 100644
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -778,9 +778,9 @@ let
     A14 = [11 13; 12 14]
     R = CartesianIndices(Compat.axes(A14))
     @test [a for (a,b) in pairs(IndexLinear(),    A14)] == [1,2,3,4]
-    @test [a for (a,b) in pairs(IndexCartesian(), A14)] == vec(collect(R))
+    @test [a for (a,b) in pairs(IndexCartesian(), A14)] == collect(R)
     @test [b for (a,b) in pairs(IndexLinear(),    A14)] == [11,12,13,14]
-    @test [b for (a,b) in pairs(IndexCartesian(), A14)] == [11,12,13,14]
+    @test [b for (a,b) in pairs(IndexCartesian(), A14)] == [11 13; 12 14]
 end
 
 # Val(x)
```
But that requires fixing the result of `pairs`, which currently relies on `IndexValue` on 0.6. Should we replace the whole implementation https://github.com/JuliaLang/Compat.jl/blob/9ae12616a525b6d331884555269f0d538b99e3b0/src/Compat.jl#L574-L614 with the current one from Base?

nalimilan:
I'm afraid we cannot change the existing implementation on 0.6 or it could break packages that rely on it. So maybe just make the test conditional on the Julia version? Anyway it's a relatively obscure method.

BTW, https://github.com/JuliaLang/julia/pull/27038 will change this test again.

martinholters:
I really don't know enough about the details to make a call on how to proceed here, but...

> So maybe just make the test conditional on the Julia version?

[...]

PR #543, opened 2018-05-15
Export isnumeric

jmkuhn:
`isnumeric` is now exported from Base. https://github.com/JuliaLang/julia/pull/25479

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,6 +413,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `cfunction` is now `@cfunction` ([#26486]).
 
+* `Unicode.isnumeric` is now available as `isnumeric` ([#25479]).
+
 ## New macros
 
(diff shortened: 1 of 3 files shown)


jmkuhn:
Build errors due to Issue #540.

README.md
```diff
@@ -413,6 +413,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `cfunction` is now `@cfunction` ([#26486]).
 
+* `Unicode.isnumeric` is now available as `isnumeric` ([#25479]).
```
nalimilan: This description isn't correct: it's `Base.Unicode.isnumeric`. `Unicode.isnumeric` doesn't exist. I think `isnumeric` should just be mentioned with `isletter`, with something like "`isnumber` is now `isnumeric`".

PR #545, opened 2018-05-15
Add Sockets module.

rofinn:
~Currently, based off #544 for testing purposes.~


src/Compat.jl
```diff
@@ -2,6 +2,23 @@ __precompile__()
 
 module Compat
 
+# https://github.com/JuliaLang/julia/pull/25935
+if VERSION < v"0.7.0-DEV.4442"
+    @eval module Sockets
+        import Base:
+            @ip_str, IPAddr, IPv4, IPv6, UDPSocket, TCPSocket,
+            accept, connect, getaddrinfo, getipaddr, getsockname, listen,
```
ararslan: Looks like you're missing `bind`.

rofinn: Weird. Is there a reason it wasn't deprecated in base? https://github.com/JuliaLang/julia/blob/180111f53155e99fa71bcb461ee562152307720a/base/sysimg.jl#L895

ararslan: Indeed. Apparently `Base.bind` is also used for `Channel`s.

rofinn: Should I still import and export it in the fake module?

ararslan: Might as well, I guess?

rofinn:
Thank you.

PR #546, opened 2018-05-15
Add rmul!

### LINKED ISSUE #496: Add rmul!
In my particular case I need it as `scale!(A::Array, b::Number)` is deprecated in favour of `rmul!(A,b)`.

ararslan:
Fixes #496.

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.qr` takes `pivot` as a `Val` _instance_ and keyword argument `full` ([#22475], [#24279]).
 
+* `Compat.rmul!` provides a subset of the functionality of `LinearAlgebra.rmul!` for
+  use with Julia 0.6 ([#25701], [#25812]).
+
 ## Renaming
 
 * `Display` is now `AbstractDisplay` ([#24831]).
@@ -607,10 +610,12 @@ includes this fix. Find the minimum version from there.
 [#25647]: https://github.com/JuliaLang/julia/issues/25647
 [#25654]: https://github.com/JuliaLang/julia/issues/25654
 [#25662]: https://github.com/JuliaLang/julia/issues/25662
+[#25701]: https://github.com/JuliaLang/julia/issues/25701
 [#25705]: https://github.com/JuliaLang/julia/issues/25705
 [#25706]: https://github.com/JuliaLang/julia/issues/25706
 [#25738]: https://github.com/JuliaLang/julia/issues/25738
 [#25780]: https://github.com/JuliaLang/julia/issues/25780
+[#25812]: https://github.com/JuliaLang/julia/issues/25812
 [#25819]: https://github.com/JuliaLang/julia/issues/25819
 [#25873]: https://github.com/JuliaLang/julia/issues/25873
 [#25896]: https://github.com/JuliaLang/julia/issues/25896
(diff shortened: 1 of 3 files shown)


fredrikekre:
This is more involved than just `const rmul! = Base.scale!`, see https://github.com/JuliaLang/julia/pull/25701 (and https://github.com/JuliaLang/julia/pull/25812).

README.md
```diff
@@ -400,6 +400,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `LinSpace` is now `LinRange` ([#25896]).
 
+* `scale!` is now `rmul!` ([#25571]).
```
martinholters: Most importantly `scale!(::AbstractArray, ::Number)` is, but In general, it can be `rmul!` or `lmul!`.

src/Compat.jl
```diff
@@ -1842,6 +1842,13 @@ else
     using LinearAlgebra: qr
 end
 
+if VERSION < v"0.7.0-DEV.3449"
+    const rmul! = Base.scale!
```
martinholters: Let's rather define a new function with only those methods where `scale!` does the right thing. Maybe only  `rmul!(::AbstractArray, ::Number)` for now.

src/Compat.jl
```diff
@@ -1842,6 +1842,13 @@ else
...
+    export rmul!
```
martinholters: Needs to be in the `LinearAlgebra` module, so that `using Compat.LinearAlgebra` (without `using Compat`) makes it available.

README.md
martinholters: Why `#25571`? Shouldn't that rather be `#25701` and `#25812`?

ararslan:
> it would also be nice if this lived in the LinearAlgebra module

Agreed. If you have any suggestions for how to structure that sensibly, do let me know; I wasn't able to think of a good solution.

src/Compat.jl
```diff
@@ -1842,6 +1842,13 @@ else
     using LinearAlgebra: qr
 end
 
+if VERSION < v"0.7.0-DEV.3449"
```
martinholters: Wrong version? Should be `0.7.0-DEV.3563`, but also needs a definition until `0.7.0-DEV.3665` in terms of `mul1!`.

martinholters:
> If you have any suggestions for how to structure that sensibly, do let me know

I guess `eval`ing the new definition into `LinearAlgebra` might be ok; it's not that much worse than the type-piracy that Compat commits all over the place. Otherwise, I'd favor simply not exporting `rmul!`, so that users need to write `Compat.rmul!`.

martinholters:
Or something along those lines might work:
```julia
module _LinearAlgebra
    using LinearAlgebra # or Base.LinAlg prior to 0.7.0-DEV.3449
    for n in names(LinearAlgebra)
        @eval export $n
    end
    # more definitions like rmul! as needed
end
const LinearAlgebra = _LinearAlgebra
```

rofinn:
FWIW, I'd also be fine with not exporting `rmul!`.

ararslan:
Yeah that seems like the way to go. `eval`ing into a Base module seems pretty un-legit, which is why I didn't do that to begin with.

ararslan:
This is needed for progress elsewhere so I'm just going to go ahead and merge and tag.

PR #548, opened 2018-05-18
Bump Project.toml version to 0.65.0

ararslan:


ararslan:
Looks like AppVeyor is having trouble installing Julia on x86-64; failures are timeouts.

PR #549, opened 2018-05-21
Add fetch(::Task)

ararslan:
cc @iamed2


ararslan:
Test failures addressed in #550.

PR #550, opened 2018-05-22
RFC: Remove obsolete custom Iterators module

martinholters:
Our tests have started failing because
```julia
julia> @test collect(rest(1:10, 5)) == [5,6,7,8,9,10]
Test Failed at REPL[13]:1
  Expression: collect(rest(1:10, 5)) == [5, 6, 7, 8, 9, 10]
   Evaluated: [6, 7, 8, 9, 10] == [5, 6, 7, 8, 9, 10]
```
the state for iterating over a range has changed its meaning. (Arguably, it's a bad idea to depend on the meaning of iteration state like that.)

As on all supported Julia versions, we just load `Base.Iterators`, there is no point in defining our own module, and then I hope we can just remove all those tests instead of developing elaborate alternatives.

There is a twist: We've been `using Base.Iterators` (if present), meaning that e.g. `Compat.zip(...)` would have worked, but only since Julia 0.6, where `Iterators.zip(...)` without any `Compat` worked just as well. Anyone actually aiming at compatibility with Julia 0.5 had to use `Compat.Iterators`, so I guess switching from `using` to `import` here should not break anyones code. At some point, we should also deprecate that binding, but I'm not sure whether that point is now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,10 +82,6 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## Module Aliases
 
-* In 0.6, some 0.5 iterator functions have been moved to the `Base.Iterators`
-  module. Code can be written to work on both 0.5 and 0.6 by `import`ing or
-  `using` the `Compat.Iterators` module instead. ([#18839])
-
 * `using Compat.Test`, `using Compat.SharedArrays`, `using Compat.Mmap`, and `using
   Compat.DelimitedFiles` are provided on versions older than 0.7, where these are not yet
   part of the standard library. ([#23931])
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -321,24 +321,7 @@ end # module TypeUtils
 include("arraymacros.jl")
 
 # julia #18839
-if VERSION < v"0.6.0-dev.1024"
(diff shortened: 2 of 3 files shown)


ararslan:
This is an odd failure:
```
Test Failed at /home/travis/.julia/v0.7/Compat/test/runtests.jl:676
  Expression: Dates.Day(30) == Dates.Month(1)
    Expected: MethodError
  No exception thrown
ERROR: LoadError: There was an error during testing
in expression starting at /home/travis/.julia/v0.7/Compat/test/runtests.jl:676
```
Oddly enough it seems not to be affecting macOS 0.7, only Linux and Windows...

martinholters:
These should probably give `false` to match current Base behavior:
https://github.com/JuliaLang/Compat.jl/blob/c5527e9b237dff76f2f4f2564f7fe83176ad4dbc/src/Compat.jl#L449-L450
Making an error a non-error should be non-breaking, so I don't see any harm in changing that. Will open a PR shortly...

martinholters:
Ah no, looking again, the `Dates` comparison stuff is only relevant prior to 0.6, so I'll just remove it completely.

PR #551, opened 2018-05-22
update Compat for rename of Pkg to OldPkg

KristofferC:
also temporarily disable the Project file while things are in flux

Somehow I managed to make the #547 branch close itself... -.-.

@StefanKarpinski, @martinholters


martinholters:
LGTM by itself (shame we have to disable the Project.toml), but do we still plan to add something as `Compat.Pkg` in the (near) future? That would be Pkg3 on 0.7 and ... `Base.Pkg` on 0.6?

So the idea would be to use
* `Compat.OldPkg` if one really needs the old Pkg2 functionality/API, even if that may be of limited value on 0.7
* `Compat.Pkg` if one needs a way of accessing very basic packaging functionality (`add`, `dir`), as is often needed e.g. for CI, compatible with 0.6, 0.7, and onward.

Doesn't sound too bad to me. But I assume the second use to be more common, so maybe we should skip the deprecation?

And of course, if we agree on this as the way forward, README and tests need adjustment.

KristofferC:
Yes, that makes sense to me. So if this PR is ok for you, I intend to merge https://github.com/JuliaLang/julia/pull/27095 when I fixed CI, then merge this and tag a new Compat, then update the registry and we should be able to release an alpha for 0.7.

martinholters:
This PR is ok with me, modulo test and README update.

But should we keep `Compat.Pkg` undeprecated or not? (Yes, suddenly switching `Compat.Pkg` from Pkg2 to Pkg3 might be breaking. But a deprecation that tells people to use `Compat.OldPkg` instead of `Compat.Pkg` when in the majority of cases, they should soon be using `Compat.Pkg` again soon does not sound too useful.) Or should we make `Compat.Pkg` be Pkg3 (when available) immediately?

KristofferC:
> But should we keep Compat.Pkg undeprecated or not?

Ok, let's not then. The PR is then as simple as this.

martinholters:
Sometimes, simple really is best. :smile: And we're surely not doing anything wrong here.

Also, this agrees with (what I perceived) we came to in #547:
* Being able to access Pkg2 on 0.7 will be of little use, probably no need to support that.
* We just let `Compat.Pkg` become Pkg3 and wait for the complaint to come in.

Bottom line: LGTM. Also, no need to wait for JuliaLang/julia#27095 to merge this, or?

KristofferC:
No, this should be good to go then. Thanks for your feedback, greatly appreciated!

martinholters:
We probably should wait a little more to give others a chance to chime in? Having a PR open at least 24 hours if not terribly urgent is a nice rule. OTOH, #547 had been open for a while, so :man_shrugging:

PR #554, opened 2018-05-23
Remove Dates comparison code which has been in Base since 0.6

martinholters:
The behavior of `==(x::Dates.FixedPeriod, y::Dates.OtherPeriod)` has been changed in https://github.com/JuliaLang/julia/pull/27165, making our tests fail. As the relevant code is no longer needed on 0.6 or later, I figured we could just delete the tests as well.

Note: This leaves the difference for `==(x::Dates.FixedPeriod, y::Dates.OtherPeriod)` between 0.6 and 0.7. If we need a Compat for that, I'd suggest opening a separate PR.


fredrikekre:
> Compat requires Julia 0.6 at a minimum, so we can probably just delete all code that's in a if VERSION < v"0.6.0-dev..." block.

Perhaps I should revive #449

martinholters:
Is it worth the effort? Not that I'm against it; I like getting rid of old, unused stuff, but we could just as well keep removing them once they cause trouble.

PR #555, opened 2018-05-24
Don't test that `Pkg` exports `add`

martinholters:
Pkg3 still defines `add`, but doesn't export it anymore.

Minor follow-up to #551.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1275,7 +1275,7 @@ module TestPkg
     using Compat.Pkg
     using Compat.Test
     @test isdefined(@__MODULE__, :Pkg)
-    @test isdefined(@__MODULE__, :add)
+    @test isdefined(Compat.Pkg, :add)
 end
 
 module TestInteractiveUtils


martinholters:
Remaining test failures are due to https://github.com/JuliaLang/julia/pull/27212. What do we do with those? CC @Sacha0

fredrikekre:
What is the failure?

martinholters:
```
ERROR: LoadError: WARNING: Base.qr is deprecated: it has been moved to the standard library package `LinearAlgebra`.
Add `using LinearAlgebra` to your imports.
 in module Base
MethodError: no method matching qr(::Array{Int64,2}, ::Val{false}; full=false)
Closest candidates are:
  qr(::AbstractArray{T,2}, ::Any) where T at usr/share/julia/stdlib/v0.7/LinearAlgebra/src/qr.jl:335 got unsupported keyword argument "full"
  qr(::AbstractArray{T,2}) where T at usr/share/julia/stdlib/v0.7/LinearAlgebra/src/qr.jl:340 got unsupported keyword argument "full"
Stacktrace:
 [1] top-level scope at Compat/test/runtests.jl:1677
```
It's the diappearance of the `full` keyword.

fredrikekre:
Perhaps we can just version that test then.

martinholters:
Sure, we could do that to get the tests pass. But providing this `qr` method on 0.6 only is of little use. It won't help with the migration to 0.7 at all. Ideally, we should probably deprecate the `full` keyword to ... and this is where it gets tricky. I'll open an issue for the discussion, it doesn't really belong here.

PR #558, opened 2018-05-28
RFC: Neither advertise `full` keyword of `qr` nor test it on 0.7

### LINKED ISSUE #556: Fate of qr
From our README:

* `Compat.qr` takes `pivot` as a `Val` _instance_ and keyword argument `full` ([#22475], [#24279]).

And we have these tests:

https://github.com/JuliaLang/Compat.jl/blob/2a822219a157b7fe0f45c6549fe4020c342ac29f/test/runtests.jl#L1676-L1695

But in JuliaLang/julia#27212 the `LinearAlgebra` API has been changed, `qr` doesn't take a `full` (`thin` in 0.6) keyword anymore, making these tests fail. Instead, both the "full" and the "thin" representation can be obtained from the returned object.

Ideally, we provide a `qr` that does the same and deprecate the current method to that. But I have no idea whether that is feasible.

martinholters:
The `thin` keyword was replaced by `full`, but later abandoned completely. To keep code working on Julia 0.6 that had adapted to `full`, the interface is retained on Julia 0.6, but neither tested on 0.7 (where it doesn't work) nor advertised in the README.

Closes #556.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `fetch` for `Task`s ([#25940]).
 
-* `Compat.qr` takes `pivot` as a `Val` _instance_ and keyword argument `full` ([#22475], [#24279]).
+* `Compat.qr` takes `pivot` as a `Val` _instance_ ([#22475]).
 
 * `Compat.rmul!` provides a subset of the functionality of `LinearAlgebra.rmul!` for
   use with Julia 0.6 ([#25701], [#25812]).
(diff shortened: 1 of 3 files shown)


martinholters:
CC @Sacha0, @fredrikekre - you had something like this in mind?

martinholters:
Just noticed that some comments in the code would be helpful in the future, will update momentarily...

martinholters:
I'm not expecting tests to pass on the latest commit due to #558, but it's an NFC, so I'll merge this later today unless objections are voiced.

PR #559, opened 2018-05-29
Neither advertise the functions moved to StatsBase nor test them on 0.7

martinholters:
These functions call through to Base functions, which is rendered impossible after their moval to StatsBase (JuliaLang/julia#27152). As StatsBase is using Compat, Compat cannot use StatsBase. In the future, these functions should print deprecation warnings also on 0.6, telling people how to get the desired functionality, once that is settled (see discussion in https://github.com/JuliaStats/StatsBase.jl/pull/379).

Alternatively, we could duplicate the code now moved to StatsBase in Compat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,11 +287,11 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.mv` and `Compat.cp` with `force` keyword argument ([#26069]).
 
-* `Compat.accumulate`, `Compat.accumulate!`, `Compat.all`, `Compat.any`, `Compat.cor`,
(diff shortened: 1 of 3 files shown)


martinholters:
BTW, what works right now (well, with sufficiently recent StatsBase) is this e.g. pattern:
```
if VERSION < v"0.7.0-DEV.5238"
    y = Compat.varm(...)
else
    y = StatsBase.varm(...)
end
```
I don't think that is something we should be advertising, though.

martinholters:
Known unrelated OSX failure.

nalimilan:
Maybe we should still run the tests on Julia 0.6? Packages can already rely on these functions and we really don't want to break them there.

martinholters:
Batting objections, I'll merge later today.

PR #561, opened 2018-05-30
Restore Project.toml with updated Pkg uuid and bump version to 0.66.0

martinholters:
Note: As this bumps the version, the merge commit should also be tagged as a release. I don't see why we shouldn't do so, soon, just want to be clear about it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "v0.65.0"
+version = "v0.66.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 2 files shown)


martinholters:
`ERROR: Pkg.Types.CommandError("Cannot add package with the same name or uuid as the project")`

Travis just continues, but AV fails. I guess the `build_script` needs some adjustment...

martinholters:
... or could we make `Pkg.clone` more tolerant here, given that its raison d'être is to ease transition?

martinholters:
Yes, fixing the Travis script is a good idea, of course; too bad we cannot centralize that for AppVeyor. But eventually, people will have to adjust the AV scripts, so forcing them to do so when adding a Project.toml might be ok.

martinholters:
CI back to all-green :tada: 

My plan would be to merge #560 first, then this on top to have a reasonably "deprecation clean" release.

Note: This should be squashed upon merge.

PR #562, opened 2018-05-31
Add something

nalimilan:
Needed to fix many deprecations with 0.7-alpha.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `isbits(t::Type)` is now `isbitstype(t)` ([#26850]).
 
+* `something` to get the first argument different from `nothing`, unwrapping those
+  of the `Some` type ([#27258]).
+
 ## Renaming
 
 * `Display` is now `AbstractDisplay` ([#24831]).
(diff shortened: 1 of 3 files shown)


nalimilan:
I don't think we can do anything to the existing `coalesce`, as changing it would break packages even on 0.6. Anyway it doesn't really hurt: on 0.6 `coalesce` doesn't exist in Base anyway so defining it doesn't hurt, and we don't define it on 0.7.

martinholters:
But having `Compat.coalesce` do different things on 0.6 and 0.7 also feels very wrong. Especially as more packages might begin to be ported with 0.7-alpha out which would benefit from having consistency here.

nalimilan:
> But having Compat.coalesce do different things on 0.6 and 0.7 also feels very wrong. Especially as more packages might begin to be ported with 0.7-alpha out which would benefit from having consistency here.

Yes, but I don't see any alternative which wouldn't imply printing suddenly loads of deprecation warnings on 0.6. The inconsistency in `coalesce` is probably not that bad, given that most uses of it in packages should probably switch to `something` as soon as possible (e.g. for the search/find deprecations). `coalesce` is mostly useful when working with data with `missing`s, i.e. in scripts rather than packages.

martinholters:
All test failures here should be fixed by #563, so barring objection, I'll merge this later today.

PR #563, opened 2018-06-01
Fixes for 0.7-alpha

martinholters:
This deals with the latest changes in Base in the conservative way: stop testing here on newer Julia, but keep them around on Julia 0.6. This affects `chol(::UniformScaling)` which will keep working on 0.6 but gives a deprecation warning on 0.7 with a replacement also suitable for 0.7. So no harm here, people can adjust their code and we may or may not deprecate/remove that `chol` definition eventually.

More troublesome is `coalesce`, which has changed its meaning. Following the brief discussion in #562, I've kept its old meaning here, but only on older Julia, and therefore also `VERSION`-gated the tests. Fortunately, it is not mentioned in the README, so hopefully, its uses are rare. Summary of main differences:

expression| 0.6 (with Compat) | 0.7
--|--|--
`coalesce(nothing, Some(1))` | `1` | `nothing`
`coalesce(Some(1), nothing)` | `1` | `Some(1)`

In 0.7, `coalesce` is for working with `missing`, which doesn't exist in 0.6 (and there is no Compat for it, either), so people should switch to using `something` to be added in #562.
I'm open to deprecating `coalesce` in Compat in favor of `something`. But I'd like to keep this PR focused on fixing the existing test failures, and let a potentially long discussion about a better way forward happen somewhere else.

That said, I've sneaked in a tiny breaking change to the promotion rules of `Some` which I hope not to have any significant impact on existing code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,8 +211,6 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * The `reshape` and `ntuple` APIs are extended to support `Val{x}()` arguments on 0.6 and below.
 
-* `chol` and `chol!` for `UniformScalings` ([#22633]).
-
 * `logdet` for `Number`s ([#22629]).
 
 * `fieldcount` is equivalent to `nfields` for Julia versions 0.6 and below and is used to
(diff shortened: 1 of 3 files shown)


martinholters:
CC @ararslan regarding the `coalesce` thing, as you did the initial implementation here

PR #568, opened 2018-06-06
bump version to 0.67.2

### LINKED ISSUE #565: scoping/escaping bug in @cfunction
The `Compat.@cfunction` macro (#553) seems to have an escaping or scoping bug of some kind that prevents it from being used in a function.  For example:
```jl
julia> using Compat

julia> foo(x) = x+1
foo (generic function with 1 method)

julia> bar() = @cfunction(foo, Int, (Int,))
bar (generic function with 1 method)

julia> bar()
ERROR: UndefVarError: foo not defined
Stacktrace:
 [1] bar() at ./REPL[3]:1

julia> baz() = cfunction(foo, Int, Tuple{Int})
baz (generic function with 1 method)

julia> baz()
Ptr{Void} @0x0000000121a79f50
```
cc @ararslan .

stevengj:
(Not 0.67.1 because I accidentally tagged that without updating the TOML file.)   This fixes #565, which I need to update ZMQ.jl.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "v0.67.0"
+version = "v0.67.2"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"


stevengj:
Now that #543 is merged then we need to bump the minor version.

PR #572, opened 2018-06-10
Rename split function keyword argument from keep to keepempty

### LINKED ISSUE #571: split function changed keyword argument name from keep to keepempty
https://github.com/JuliaLang/julia/pull/26647

thanks to https://github.com/jipolanco/WriteVTK.jl/commit/ea046493c7787ae651b5f629455aaf7bdaa000b1

appleparan:
fix #571

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `isupper`, `islower`, `ucfirst` and `lcfirst` are now `isuppercase`, `islowercase`,
   `uppercasefirst` and `lowercasefirst` ([#26442]).
 
+* `Compat.split` and `Compat.rsplit` accept `keepempty` keyword argument
+  if `splitter` is given as second argument ([#26634])
+
 * `isalpha` is now `isletter` ([#27077]).
 
 * `cfunction` is now `@cfunction` ([#26486]).
@@ -649,6 +652,7 @@ includes this fix. Find the minimum version from there.
 [#26442]: https://github.com/JuliaLang/julia/issues/26442
 [#26486]: https://github.com/JuliaLang/julia/issues/26486
 [#26559]: https://github.com/JuliaLang/julia/issues/26559
+[#26634]: https://github.com/JuliaLang/julia/issues/26634
 [#26660]: https://github.com/JuliaLang/julia/issues/26660
 [#26670]: https://github.com/JuliaLang/julia/issues/26670
 [#26850]: https://github.com/JuliaLang/julia/issues/26850
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1992,6 +1992,14 @@ end
     Base.atan(x::Real, y::Real) = atan2(x, y)
 end
 
+# https://github.com/JuliaLang/julia/pull/26647
+@static if VERSION < v"0.7.0-DEV.4724"
+    rsplit(s::AbstractString, splitter; limit::Integer=0, keepempty::Bool=false) =
+        Base.rsplit(s, splitter; limit=limit, keep=keepempty)
+    split(s::AbstractString, splitter; limit::Integer=0, keepempty::Bool=false) =
+        Base.split(s, splitter; limit=limit, keep=keepempty)
+end
+
 include("deprecated.jl")
 
 end # module Compat
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1882,4 +1882,33 @@ end
 @test atan(-1.0, -2.0) ≈ atan(0.5) - π
 @test atan(big"-1.0", big"-2.0") ≈ atan(big"0.5") - π
 
+# 0.7.0-DEV.4724
+let
+    @test Compat.split("", ','  ; keepempty=false) == []
+    @test Compat.split(",", ',' ; keepempty=false) == []
+    @test Compat.split(",,", ','; keepempty=false) == []
+    @test Compat.rsplit("", ','  ; keepempty=false) == []
+    @test Compat.rsplit(",", ',' ; keepempty=false) == []
+    @test Compat.rsplit(",,", ','; keepempty=false) == []
+
+    str = "a.:.ba..:..cba.:.:.dcba.:."
+    @test Compat.split(str, ".:."; keepempty=false) == ["a","ba.",".cba",":.dcba"]
+    @test Compat.split(str, ".:."; keepempty=true) == ["a","ba.",".cba",":.dcba",""]
+    @test Compat.split(str, ".:."; limit=3, keepempty=false) == ["a","ba.",".cba.:.:.dcba.:."]
+    @test Compat.split(str, ".:."; limit=3, keepempty=true) == ["a","ba.",".cba.:.:.dcba.:."]
+    @test Compat.rsplit(str, ".:."; keepempty=false) == ["a","ba.",".cba.:","dcba"]
+    @test Compat.rsplit(str, ".:."; keepempty=true) == ["a","ba.",".cba.:","dcba",""]
+    @test Compat.rsplit(str, ".:."; limit=3, keepempty=false) == ["a.:.ba.",".cba.:","dcba"]
+    @test Compat.rsplit(str, ".:."; limit=3, keepempty=true) == ["a.:.ba..:..cba.:","dcba",""]
+
+    @test Compat.split(str, r"\.(:\.)+"; keepempty=false) == ["a","ba.",".cba","dcba"]
+    @test Compat.split(str, r"\.(:\.)+"; keepempty=true) == ["a","ba.",".cba","dcba",""]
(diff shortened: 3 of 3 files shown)


src/Compat.jl
```diff
@@ -1933,6 +1933,18 @@ if VERSION < v"0.7.0-DEV.5278"
     export something
 end
 
+# https://github.com/JuliaLang/julia/pull/26647
+@static if VERSION < v"0.7.0-DEV.4724"
+    rsplit(s::AbstractString; limit::Integer=0, keepempty::Bool=false) =
+        Base.rsplit(s)
```
martinholters: Shouldn't this have `; limit=limit, keep=keepempty`, too?

src/Compat.jl
```diff
@@ -1933,6 +1933,18 @@ if VERSION < v"0.7.0-DEV.5278"
...
+    rsplit(s::AbstractString, splitter; limit::Integer=0, keepempty::Bool=false) =
+        Base.rsplit(s, splitter; limit=limit, keep=keepempty)
+    split(s::AbstractString; limit::Integer=0, keepempty::Bool=false) =
+        Base.split(s)
```
martinholters: Same here.

test/runtests.jl
```diff
@@ -1778,4 +1778,20 @@ let sep = Compat.Sys.iswindows() ? ';' : ':'
...
+    @test Compat.split(",,", ','; keepempty=false) == []
+    @test Compat.rsplit("", ','  ; keepempty=false) == []
+    @test Compat.rsplit(",", ',' ; keepempty=false) == []
+    @test Compat.rsplit(",,", ','; keepempty=false) == []
+
+    str = "a.:.ba..:..cba.:.:.dcba.:."
+    @test Compat.split(str, ".:."; keepempty=false) == ["a","ba.",".cba",":.dcba"]
+    @test Compat.split(str, r"\.(:\.)+"; keepempty=false) == ["a","ba.",".cba","dcba"]
+    @test Compat.split(str, r"\.+:\.+"; keepempty=false) == ["a","ba","cba",":.dcba"]
+    @test Compat.rsplit(str, ".:."; keepempty=false) == ["a","ba.",".cba.:","dcba"]
```
martinholters: Would be nice to test `keepempty=true`, as well as setting a `limit`, both for the case with and without a `splitter` given.

src/Compat.jl
appleparan: At first I did it. I rebased it so I can't show the results, but it throws `MethodError`

```
Test threw an exception of type MethodError
  Expression: Compat.split("a b c"; keepempty=true) == ["a", "b", "c"]
  MethodError: no method matching split(::String; limit=0, keep=true)
  Closest candidates are:
    split(::AbstractString) at strings/util.jl:302 got unsupported keyword arguments "limit", "keep"
    split(::T<:AbstractString, ::Any; limit, keep) where T<:AbstractString at strings/util.jl:277
    split(::T<:SubString, ::Any; limit, keep) where T<:SubString at strings/util.jl:253
  Stacktrace:
   [1] (::Compat.#kw##split)(::Array{Any,1}, ::Compat.#split, ::String) at ./<missing>:0
   [2] include_from_node1(::String) at ./loading.jl:576
   [3] include(::String) at ./sysimg.jl:14
   [4] process_options(::Base.JLOptions) at ./client.jl:305
   [5] _start() at ./client.jl:371
```

https://travis-ci.org/appleparan/Compat.jl/jobs/390385912

test/runtests.jl
appleparan: Ok. I just copied tests from original test from Julia repo.

src/Compat.jl
martinholters: Well... either call `Base.split(s, Base._default_delims)` or not define this method at all. But silently doing the wrong thing for `limit != 0` or `keepempty = false` seems pretty bad.

appleparan: yes, you are right. I fixed it.

test/runtests.jl
```diff
@@ -1778,4 +1778,33 @@ let sep = Compat.Sys.iswindows() ? ';' : ':'
...
+
+    @test Compat.split(str, r"\.(:\.)+"; keepempty=false) == ["a","ba.",".cba","dcba"]
+    @test Compat.split(str, r"\.(:\.)+"; keepempty=true) == ["a","ba.",".cba","dcba",""]
+    @test Compat.split(str, r"\.(:\.)+"; limit=3, keepempty=false) == ["a","ba.",".cba.:.:.dcba.:."]
+    @test Compat.split(str, r"\.(:\.)+"; limit=3, keepempty=true) == ["a","ba.",".cba.:.:.dcba.:."]
+    @test Compat.split(str, r"\.+:\.+"; keepempty=false) == ["a","ba","cba",":.dcba"]
+    @test Compat.split(str, r"\.+:\.+"; keepempty=true) == ["a","ba","cba",":.dcba",""]
+    @test Compat.split(str, r"\.+:\.+"; limit=3, keepempty=false) == ["a","ba","cba.:.:.dcba.:."]
+    @test Compat.split(str, r"\.+:\.+"; limit=3, keepempty=true) == ["a","ba","cba.:.:.dcba.:."]
+end
```
martinholters: This nicely covers the methods with `splitter`, but doesn't test the ones without.

stevengj:
Ping @appleparan — needs a rebase and test coverage for methods that don't pass `splitter`.

test/runtests.jl
appleparan: I decided not to include non-`splitter` version, because `split/rsplit` doesn't have `keep` keyword argument on 0.6. It doesn't expected to use it.

appleparan:
I decided not to include non-`splitter` version, because `split/rsplit` doesn't have keep keyword argument on 0.6. It doesn't expected to be used it.

also, I rebased all of my commit.

appleparan:
OMG. I did a mistake on rebase sorry. I will fix it

stevengj:
@appleparan, I think the concern of @martinholters was that you have no tests for the one-argument `split` function (which defaults to splitting on whitespace).

appleparan:
@stevengj Yes, it was. 

By the way, in Julia v0.6.3, `split` function without `splitter` just use default delimiter and keyword arguments, `limit` and `keep`. Moreover, `rsplit` doesn't have non-`splitter` version (commented).
```
split(str::AbstractString) = split(str, _default_delims; limit=0, keep=false)
#rsplit(str::AbstractString) = rsplit(str, _default_delims, 0, false)
```
https://github.com/JuliaLang/julia/blob/d55cadc350d426a95fd967121ba77494d08364c8/base/strings/util.jl#L302

In my first version, I forced to use keyword argument `keepempty` whatever the argument was (with `splitter` or without `splitter`). That makes tests fail because original `split` function without `splitter` didn't accept `limit` and `keep` explicitly. @martinholters pointed this and gave me two choices, one for not include non-`splitter` ver. and another for force to use default argument. I decided to follow the first one.  

**Summary**: I did not include non-`splitter` version and didn't write tests for them.

README.md
```diff
@@ -415,6 +415,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `isupper`, `islower`, `ucfirst` and `lcfirst` are now `isuppercase`, `islowercase`,
   `uppercasefirst` and `lowercasefirst` ([#26442]).
 
+* `keep` keyword argument in `split` and `rsplit` is now `keepempty` ([#26647])
```
martinholters: This should point out that you have to use `Compat.split` and `Compat.rsplit`.

martinholters:
Yes, this doesn't provide the complete functionality of 0.7, but it does provide the new API for the old functionality, which should suffice for packages to upgrade.

appleparan:
I will rebase it.

martinholters:
CI failures look unrelated, and it has passed on two of the 0.7 runs, so this should be good to go. Barring objections, I'll merge this later today.

stevengj:
I don't understand why you can't define
```jl
split(s::AbstractString; limit::Integer=0, keepempty::Bool=false) =
        Base.split(s, @static(isdefined(Base,:_default_delims) ? Base._default_delims : isspace); limit=limit, keep=keepempty)
```
on 0.6

appleparan:
On 0.6, `split` without `splitter` isn't defined to accept keyword argument. It just accept one positional argument only, `s::AbstractString`.
```
split(str::AbstractString) = split(str, _default_delims; limit=0, keep=false)
```
If `Compat` supports keyword argument `split` without `splitter` even 0.6 doesn't, I thought it's nonsense. That's why I didn't define even if I could.

stevengj:
> It feels like out of objective of this package.

The whole point of this package is to allow you to use the 0.7 API in 0.6.

appleparan:
I misunderstood. You are right. I will add additional code and tests

PR #573, opened 2018-06-12
thisind, 3-arg length/nextind/prevind, codeunit(s)

stevengj:
Compat for JuliaLang/julia#23805, JuliaLang/julia#24414, and JuliaLang/julia#24999.


stevengj:
(Note that the implementation here is simply copied from Julia master.   The only change was to replace a StringIndexError, which doesn't exist in 0.6, with a BoundsError in `thisind`.)

PR #575, opened 2018-06-14
Add `Compat.readuntil`

carlobaldassi:
This is to allow for the `keep` keyword argument, which was added in https://github.com/JuliaLang/julia/pull/25646 (while also setting it to `false` by default - this was a breaking change).

The code is basically copy-pasted from current Base with a few back-porting changes.
The tests are also taken from the Julia repo, basically as-is.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -283,6 +283,18 @@ end
     # chomp parameter preserved for compatibility with earliear Compat versions
     readline(s::IO=STDIN; chomp::Bool=true, keep::Bool=!chomp) = Base.readline(s; chomp=!keep)
     eachline(s; keep::Bool=false) = Base.eachline(s; chomp=!keep)
+
+    stripdelim(s, d::Union{Char,UInt8}) = s[end] == Char(d) ? s[1:prevind(s,lastindex(s))] : s
+    stripdelim(s, d::AbstractString) = endswith(s, d) ? s[1:prevind(s,lastindex(s),length(d))] : s
+    function readuntil(f, d; keep::Bool = false)
+        s = Base.readuntil(f, d)
+        if keep || isempty(s)
+            return s
+        else
+            return stripdelim(s, d)
+        end
+    end
+    readuntil(f, d::Vector{T}; keep::Bool = false) where {T<:Union{UInt8,Char}} = convert(Vector{T}, readuntil(f, String(d), keep=keep))
 end
 
 # https://github.com/JuliaLang/julia/pull/18727
--- a/README.md
+++ b/README.md
@@ -173,6 +173,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.eachline` with `keep` keyword argument ([#25646])
 
+* `Compat.readuntil` with `keep` keyword argument ([#25646])
+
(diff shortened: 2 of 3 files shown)


stevengj:
Wouldn't it be simpler to just define something like:
```jl
stripdelim(s, d::Union{Char,UInt8}) = s[end] == Char(d) ? s[1:prevind(s,lastindex(s))] : s
stripdelim(s, d::AbstractString) = endswith(s, d) ? s[1:prevind(s,lastindex(s),length(d)] : s
function readuntil(f, d; keep::Bool = false)
    s = Base.readuntil(f, d)
    if keep || isempty(s)
        return s
    else
        return stripdelim(s, d)
    end
end
```
This is a bit slower, of course, but optimizing speed on 0.6 is not a big priority for Compat.

carlobaldassi:
Ok I used the slower simpler version, keeping the tests from Base.

src/Compat.jl
```diff
@@ -341,6 +341,25 @@ end
     # chomp parameter preserved for compatibility with earliear Compat versions
     readline(s::IO=STDIN; chomp::Bool=true, keep::Bool=!chomp) = Base.readline(s; chomp=!keep)
     eachline(s; keep::Bool=false) = Base.eachline(s; chomp=!keep)
+
+    stripdelim(s, d::Union{Char,UInt8}) = s[end] == Char(d) ? s[1:prevind(s,lastindex(s))] : s
+    function stripdelim(s, d::AbstractString)
+        !endswith(s, d) && return s
+        i = lastindex(s)
+        for k = 1:length(d)
+            i = prevind(s, i)
```
stevengj: You can use `prevind(s, i, n)` as I originally suggested in my comment, since this was added to Compat in #573.

src/Compat.jl
```diff
@@ -341,6 +341,25 @@ end
...
+    end
+    function readuntil(f, d; keep::Bool = false)
+        s = Base.readuntil(f, d)
+        if keep || isempty(s)
+            return s
+        else
+            return stripdelim(s, d)
+        end
+    end
+    readuntil(f, d::T; keep::Bool = false) where {T<:AbstractVector} = convert(T, readuntil(f, String(d), keep=keep))
```
stevengj: `String(d)` seems wrong here.  Maybe `String(reinterpret(UInt8, d))`?

src/Compat.jl
carlobaldassi: Ah, I didn't have the latest master. Fixed, thanks.

src/Compat.jl
stevengj: The `convert(T, ` also seems wrong, since `String` doesn't have a `convert` method for an arbitrary array type

in general, it looks like this will fail for `T != UInt8` here?

stevengj: Maybe we can just support `d::Vector{UInt8}` in Compat, since I'm guessing few people need other vector types.

carlobaldassi:
I have updated the code now (sorry for the delay, broken laptop...).
@stevengj I have restricted support to `Vector{UInt8}` and `Vector{Char}`, which seem to be the two relevant cases.

PR #579, opened 2018-06-18
Drop `retry`

martinholters:
This is not needed anymore on Julia 0.6 or later and the test threw a deprecation warning.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -91,64 +91,6 @@ end
         eval(mod, :(include_string($code, $fname)))
 end
 
-if VERSION < v"0.6.0-dev.2042"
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -91,64 +91,6 @@ end
...
-                    end
-                end
-                (delay, state) = next(delays, state)
-                sleep(delay)
-            end
-        end
-    end
-else
-    import Base.ExponentialBackOff
-    import Base.retry
```
martinholters: These are exported by `Base` anyway, so `Compat.retry` and `Compat.ExponentialBackOff` will continue to work.

PR #582, opened 2018-06-21
single-argument permutedims(x)

stevengj:
For JuliaLang/julia#24839, cc @andyferris.


stevengj:
Looks like some unrelated Pkg/uuid problem is causing nightly to fail?
```
ERROR: The following package names could not be resolved:
 * Compat (not found in project or manifest)
Please specify by known `name=uuid`.
```

PR #586, opened 2018-07-04
required keyword arguments working

timkittel:
Following up on the comment by @Keno in this PR JuliaLang/METADATA.jl#15489 I added the functionality to understand required keyword arguments to `@compat`. Basically that means
```Julia
@compat f(;x) = [...]
```
is equivalent to 
```Julia
f(;x=error("RequiredKeywordArgumentError: `x` is a required keyword argument, please provide `x = ...`.")) = [...]
```

As this is my first submission to a JuliaLang rep, could you check that the style etc. matches what you would  expect?

### DIFF
--- a/src/compatmacro.jl
+++ b/src/compatmacro.jl
@@ -25,12 +25,25 @@ else
     new_style_typealias(ex) = false
 end
 
+"Provides the error message if no required keyword argument is given."
+rka_error_message(sym) = string("RequiredKeywordArgumentError: `", sym, "` is a required keyword argument, please provide `", sym, " = ...`.")
+
+"Convert a functions symbol argument to the corresponding required keyword argument."
+function symbol2kw(sym::Symbol)
+    Expr(:kw, sym, Expr(:call, error, rka_error_message(sym)))
+end
+symbol2kw(arg) = arg
+
 function _compat(ex::Expr)
     if ex.head === :call
         f = ex.args[1]
         if VERSION < v"0.6.0-dev.826" && length(ex.args) == 3 && # julia#18510
                 istopsymbol(withincurly(ex.args[1]), :Base, :Nullable)
             ex = Expr(:call, f, ex.args[2], Expr(:call, :(Compat._Nullable_field2), ex.args[3]))
+        end
+        if length(ex.args) > 1 && isexpr(ex.args[2], :parameters)
+            params = ex.args[2]
+            params.args = map(symbol2kw, params.args)
         end
     elseif ex.head === :curly
         f = ex.args[1]
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1856,4 +1856,35 @@ let
     @test Compat.split(str, r"\.+:\.+"; limit=3, keepempty=true) == ["a","ba","cba.:.:.dcba.:."]
 end
 
+# test required keyword arguments
+@compat func1() = 1
+@test func1() == 1 # using the function works
+@compat func2(x) = x
+@test func2(3) == 3 # using the function works
+@compat func3(;y) = y
+@test func3(y=2) == 2 # using the function works
+try
+    func3()
+catch e
+    @test isa(e, ErrorException)
+    @test e.msg == Compat.rka_error_message(:y)
+end
+@compat func4(x; z) = x*z
+@test func4(2,z=3) == 6 # using the function works
+try
+    func4(2)
+catch e
+    @test isa(e, ErrorException)
+    @test e.msg == Compat.rka_error_message(:z)
+end
+@compat func5(;x=1, y) = x*y
+@test func5(y=3) == 3
+@test func5(y=3, x=2) == 6
+try
+    func5(x=2)
+catch e
+    @test isa(e, ErrorException)
+    @test e.msg == Compat.rka_error_message(:y)
+end
+
 nothing


KristofferC:
Restarting CI.

fredrikekre:
I suggest that you add a
```
struct UndefKeywordError <: Exception
    kw
end
Base.showerror(io::IO, e::UndefKeywordError) = print(io, "keyword argument $(e.kw) not assigned")
```
to make it look the same as on Julia 0.7.

test/runtests.jl
```diff
@@ -1856,4 +1856,35 @@ let
...
 end
 
+# test required keyword arguments
+@compat func1() = 1
+@test func1() == 1 # using the function works
+@compat func2(x) = x
+@test func2(3) == 3 # using the function works
+@compat func3(;y) = y
+@test func3(y=2) == 2 # using the function works
+try
```
fredrikekre: Instead of try-catch you can use `@test_throws`, for example
```
julia> f(;y) = y
f (generic function with 1 method)

julia> @test_throws UndefKeywordError f()
Test Passed
      Thrown: UndefKeywordError
```

timkittel:
thanks @fredrikekre , I will add that

timkittel:
The errors seem to be due to the `export UndefKeywordError` line. Should I simply not export it (which would make it a different usage than in 0.7) or should I add a condition (see below)?
```Julia
if !isdefined(:UndefKeywordError)
    struct UndefKeywordError <: Exception
        kw
    end
    Base.showerror(io::IO, e::UndefKeywordError) = print(io, "keyword argument $(e.kw) not assigned")
    export UndefKeywordError
end
```
?

fredrikekre:
Yes, you can use `!isdefined(Base, :UndefKeywordError)`, and also in the macro implementation, since this should just pass through `@compat` unchanged on julia 0.7

timkittel:
Thanks, seems to work now!

src/compatmacro.jl
```diff
@@ -25,13 +25,31 @@ else
     new_style_typealias(ex) = false
 end
 
+if !isdefined(Base, :UndefKeywordError)
+    struct UndefKeywordError <: Exception
+        kw
+    end
+    Base.showerror(io::IO, e::UndefKeywordError) = print(io, "keyword argument $(e.kw) not assigned")
```
fredrikekre: Sorry, this should also print the exception type, so 
```julia
print(io, "UndefKeywordError: keyword argument $(e.kw) not assigned")
```
as it does on Julia v0.7:
```
julia> f(;y) = y
f (generic function with 1 method)

julia> f()
ERROR: UndefKeywordError: keyword argument y not assigned
```
otherwise LGTM

timkittel:
True, corrected that.

stevengj:
Needs a README item.

timkittel:
@stevengj: absolutely. I added a description in the README, is it okai like this?

README.md
```diff
@@ -79,6 +79,8 @@ Currently, the `@compat` macro supports the following syntaxes:
   `foo(::CartesianRange{CartesianIndex{N}})` ([#20974]). Note that
   `CartesianRange` now has two type parameters, so using them as
   fields in other `struct`s requires manual intervention.
+  
+* required keyword arguments: e.g. `@compat foo(;x, y)` makes `x` and `y` required keyword arguments. When calling `foo` an error is thrown if `x` and `y` are not explictly provided
```
stevengj: Some fixes:
```
* Required keyword arguments ([#25830]). For example, `@compat foo(; x, y)` makes `x` and `y` required keyword arguments: when calling `foo`, an error is thrown if `x` and `y` are not explicitly provided.
```

stevengj: (The link to JuliaLang/julia#25830 is added if you run `julia /path/to/julia/doc/NEWS-update.jl README.jl`)

timkittel: I just pushed the update README.md.

timkittel:
@stevengj I responded to your comment and updated the README.md a few days ago. Do you agree with the result?

src/compatmacro.jl
```diff
@@ -25,13 +25,31 @@ else
...
+    end
+    Base.showerror(io::IO, e::UndefKeywordError) = print(io, "UndefKeywordError: keyword argument $(e.kw) not assigned")
+    export UndefKeywordError
+end
+
+"Convert a functions symbol argument to the corresponding required keyword argument."
+function symbol2kw(sym::Symbol)
+    Expr(:kw, sym, Expr(:call, throw, UndefKeywordError(sym)))
+end
+symbol2kw(arg) = arg
```
stevengj: This doesn't handle the case where the argument has a type.  It should probably be:
```jl
arg2kw(arg, sym) = Expr(:kw, sym, Expr(:call, throw, UndefKeywordError(sym)))
symbol2kw(arg::Symbol) = arg2kw(arg, arg)
symbol2kw(arg::Expr) = isexpr(arg, :(::)) ? arg2kw(arg, arg.args[1]) : arg
symbol2kw(arg) = arg
```

stevengj: (See https://discourse.julialang.org/t/make-function-arguments-positional-or-keyword/2984/35)

timkittel:
Seems reasonable, should I update the PR? Can I even or should this be a new PR?

PR #587, opened 2018-07-05
Set Project version to something > 0.69.0

ararslan:


stevengj:
Can we just bump to 1.0.0 (which probably should have been done long ago)?

ararslan:
I'm also fine with that.

ararslan:
So what should it be then, 1.0.0?

martinholters:
Oops, sorry, just saw this (again) seconds after tagging 0.70.0. But yes, considering that we try quite hard not to do any breaking changes here, jumping to 1.0.0 would be appropriate, I'd say.

martinholters:
I've tentatively bumped the version to 1.0.0 here. Should we go ahead and mere this and tag?

martinholters:
Barring objections, I'll pull the trigger on this one tomorrow.

ararslan:
No objection from me.

PR #589, opened 2018-07-06
Run tests on 0.7-latest

martinholters:
In case you're wondering: No 0.7(-lastest) and latest/nightly are not the same. The former is the latest 0.7 (pre-)release, i.e. 0.7-beta.0 at the moment, while the latter is, well, the latest nightly. (I think this would have exposed e.g. the problem with Statistics in #583, so it's probably worth doing both even though the difference is small.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
     - osx
 julia:
     - 0.6
+    - 0.7
     - nightly
 notifications:
   email: false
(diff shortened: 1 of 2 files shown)


martinholters:
At least locally, if I add an unknown dependency to the Project.toml, tests do fail:
```
(Compat) pkg> test
  Updating registry at `~/.julia/registries/Uncurated`
  Updating git-repo `https://github.com/JuliaRegistries/Uncurated.git`
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package Foo [abcdabcd]:
```
And isn't that the problem with Statistics, that is is unknown in beta.0? And shouldn't that then also cause CI to fail?

fredrikekre:
Ah, you are right, because Compat has a `Project.toml` file. For e.g. `StaticArrays` this was not found this way since you don't add them to the REQUIRE.

PR #590, opened 2018-07-06
(map)reduce with init keyword

martinholters:
Cherry-picked from #583 by @stevengj so we can have it before the Statistics issue is settled. Also re-added the legacy methods taking `v0` as positional argument, but `dims` as keyword argument, to avoid breaking any packages on Julia 0.6 that have already adopted that.


martinholters:
@stevengj are you ok with this?

martinholters:
Barring objections, I'll merge tomorrow.

PR #601, opened 2018-08-04
srand is now Random.seed!

rdeits:
https://github.com/JuliaLang/julia/pull/28295


rdeits:
Ah, this doesn't work because the `Compat.Random` module only ends up containing the *exported* symbols from `Base.Random`. I can either: 

1. Import every symbol from `Base.Random` into `Compat.Random` individually
2. Instead of creating a separate `Compat.Random` module, just do `@eval Random const seed! = srand` 

Any opinions?

stevengj:
Let's just manually import symbols that are documented but not exported.

rdeits:
Ok, this is a bit closer now. The only things I know I'm missing are the new `randperm!` and `randcycle!` functions.

rdeits:
Bump?

AsafManela:
bump

traktofon:
Bump? Or is it considered obsolete by now?

martinholters:
This broke (at least) `StatsBase` because it doens't provide `RangeGenerator` and `GLOBAL_RNG` in `Random`. Fix incoming.

PR #602, opened 2018-08-04
Don't precompile on recent 0.7

rened:


mlubin:
Yes please. Compat appears to be broken on 0.7-rc2 (https://travis-ci.org/JuliaOpt/MathOptInterface.jl/jobs/412077225#L499), which blocks development work to update packages to 0.7.

PR #618, opened 2018-08-16
squeeze is now dropdims

### LINKED ISSUE #617: squeeze is deprecated in favor of dropdims.
I cannot find any mention of `dropdims` in Compat, but according to the [0.7 release notes](https://docs.julialang.org/en/v0.7.0/NEWS/) it has replaced `sqeeze`. I don't know the pull number for this change.

galenlynch:
`squeeze` was renamed `dropdims` in [#28303](https://github.com/JuliaLang/julia/pull/28303)

Fixes #617

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -2011,6 +2011,11 @@ if VERSION < v"0.7.0-beta2.169"
     export floatmin, floatmax
 end
 
+# https://github.com/JuliaLang/julia/pull/28303
+if VERSION < v"0.7.0-beta2.143"
+    dropdims(X; dims = throw(UndefKeywordError("dropdims: keyword argument dims not assigned"))) = squeeze(X, dims)
+end
+
 include("deprecated.jl")
 
 end # module Compat
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -2013,6 +2013,7 @@ end
 
 # https://github.com/JuliaLang/julia/pull/28303
 if VERSION < v"0.7.0-beta2.143"
+    export dropdims
     # https://github.com/JuliaLang/julia/pull/26660
     if VERSION >= v"0.7.0-DEV.4738"
         dropdims(
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -2013,7 +2013,20 @@ end
 
 # https://github.com/JuliaLang/julia/pull/28303
 if VERSION < v"0.7.0-beta2.143"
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -2011,6 +2011,11 @@ if VERSION < v"0.7.0-beta2.169"
     export floatmin, floatmax
 end
 
+# https://github.com/JuliaLang/julia/pull/28303
+if VERSION < v"0.7.0-beta2.143"
+    dropdims(X; dims = throw(UndefKeywordError("dropdims: keyword argument dims not assigned"))) = squeeze(X, dims)
```
martinholters: To be very pedantic, this should be `squeeze(X, dims=dims)` for `VERSION > v"0.7.0-DEV.4738"`.

galenlynch: Added support for that. This is one of my first PRs to compat: is the goal to have compat work properly on all versions of julia, even old development versions?

galenlynch:
Removed `squeeze` from README

src/Compat.jl
```diff
@@ -2011,6 +2011,24 @@ if VERSION < v"0.7.0-beta2.169"
     export floatmin, floatmax
 end
 
+# https://github.com/JuliaLang/julia/pull/28303
+if VERSION < v"0.7.0-beta2.143"
+    # https://github.com/JuliaLang/julia/pull/26660
```
martinholters: I guess we should `export dropdims` here to make it available by default.

test/runtests.jl
```diff
@@ -1935,4 +1935,12 @@ end
 @test floatmax(Float32) == @eval $(Core.Intrinsics.bitcast(Float32, 0x7f7fffff))
 @test floatmin(zero(Float64)) == floatmin(Float64)
 
+# 0.7.0-beta2.143
+if VERSION < v"0.7.0-beta2.143"
+    let a = reshape(Vector(1:4),(2,2,1,1)), b = reshape(Vector(1:4), (2,2,1))
+        @test Compat.dropdims(a; dims=3) == b
+        @test_throws UndefKeywordError Compat.dropdims(a)
```
martinholters: ...and then test without the `Compat.` qualification here.

src/Compat.jl
martinholters: It's nice to have and if it can be done so easily, IMHO it's worth it.

martinholters:
Thanks!

PR #622, opened 2018-08-22
Add Tuple method for argmin/argmax

### LINKED ISSUE #621: argmin/argmax not defined for Tuples and maybe others
Although `argmin`/`argmax` on v0.7+ are defined on anything you can iterate (via `findmin`/`findmax`), that's not the case within Compat at the moment.  Currently, they are defined for:

1. `AbstractArray`
2. `AbstractVector`
3. `Associative`

Therefore, you can't currently write compatible code if you want to use `Tuple`s:

```julia
julia> VERSION
v"0.6.3"

julia> using Compat

julia> argmin((1, 2, 3.0))
ERROR: MethodError: no method matching argmin(::Tuple{Int64,Int64,Int64})
Closest candidates are:
  argmin(::AbstractArray{T,1} where T) at /Users/nowacki/.julia/v0.6/Compat/src/Compat.jl:1401
  argmin(::AbstractArray) at /Users/nowacki/.julia/v0.6/Compat/src/Compat.jl:1400
  argmin(::Associative) at /Users/nowacki/.julia/v0.6/Compat/src/Compat.jl:1402
```

Works fine with `indmin`/`indmax`, of course:

```julia
julia> indmin((1, 2, 3.0))
1
```

anowacki:
Currently, only a limited set of input collections are supported
with argmin/argmax.  Add a fallback to indmin/indmax for remaining
types, which allows them to work on Tuples.

Fixes #621.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1400,9 +1400,11 @@ end
         argmin(x::AbstractArray) = CartesianIndex(ind2sub(x, indmin(x)))
         argmin(x::AbstractVector) = indmin(x)
         argmin(x::Associative) = first(Iterators.drop(keys(x), indmin(values(x))-1))
+        argmin(x::Tuple) = indmin(x)
         argmax(x::AbstractArray) = CartesianIndex(ind2sub(x, indmax(x)))
         argmax(x::AbstractVector) = indmax(x)
         argmax(x::Associative) = first(Iterators.drop(keys(x), indmax(values(x))-1))
+        argmax(x::Tuple) = indmax(x)
     end
     export argmin, argmax
 end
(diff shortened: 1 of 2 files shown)


martinholters:
The problem is that `findmin(::Any)` in 0.6 enumerates the indices starting from 1, while on 0.7+, it uses `pairs` to determine the indices IIUC. Thus I'm not sure these necessarily give the same answer in all cases. The more conservative thing to do would be to restrict this to those cases where it's known to be correct, most notably `::Tuple`.

anowacki:
Fair enough—I've restricted this to `Tuple`s.

PR #624, opened 2018-08-23
Remove Project.toml file 

andreasnoack:
since it not needed yet and causes problems for our reverse dependency testing.

It can be added again easily when it is needed or CIBot has been upgraded.


Project.toml
```diff
@@ -1,27 +0,0 @@
-name = "Compat"
-uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "v1.0.0"
```
martinholters: Someone had tagged a new version without bumping this, which (I think) was the reason I had a problem `pkg> dev`ing this. Might the problems with CIBot be related to that?

martinholters:
CC @KristofferC, @StefanKarpinski

Project.toml
fredrikekre: > Someone had tagged a new version without bumping this, which (I think) was the reason I had a problem pkg> deving this

Likely, although a wrong version there is usually no more than a display problem

> Might the problems with CIBot be related to that?

No, apparently CIBot uses the old package manager and thats why.

martinholters: > there is usually no more than a display problem

With `dev/Compat` checked out to `v1.0.1`:
```julia
julia> VERSION
v"1.1.0-DEV.72"

(pkgtest) pkg> st
    Status `/tmp/pkgtest/Project.toml`

(pkgtest) pkg> dev Compat
  Updating git-repo `https://github.com/JuliaLang/Compat.jl.git`
┌ Warning: project file for Compat at /tmp/tmp0jlHBR is missing a `version` entry
└ @ Pkg.Types stdlib/v1.1/Pkg/src/Types.jl:700
[ Info: Path `/home/hol/.julia/dev/Compat` exists and looks like the correct package, using existing path instead of cloning
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package Compat [34da2185]:
 Compat [34da2185] log:
 ├─possible versions are: 1.0.0 or uninstalled
 ├─Compat [34da2185] is fixed to version 1.0.0
 └─restricted to versions 1.0.1 by an explicit requirement — no versions left

(pkgtest) pkg> st
    Status `/tmp/pkgtest/Project.toml`

julia> using Compat
ERROR: ArgumentError: Package Compat not found in current path:
- Run `import Pkg; Pkg.add("Compat")` to install the Compat package.
```
If I fix the version in `Project.toml`, it works.

fredrikekre: Okay, so maybe some package have a requirement on Compat 1.0.1?

martinholters: Eh, I'm doing this within an empty environment.

martinholters: Should I file an issue at Pkg.jl?

KristofferC: Should try with master of Pkg first which have fixed a number of bugs (https://github.com/JuliaLang/julia/pull/28822)

martinholters: Yep, works with JuliaLang/julia#28822.

PR #625, opened 2018-08-26
`repmat` is now `repeat`

darwindarak:
From JuliaLang/julia#26039


martinholters:
Needs tests and a README entry.

darwindarak:
Is there anything else I should add to this?

martinholters:
Thanks!

PR #626, opened 2018-08-27
Weed out relicts from at-compat

martinholters:
In the spirit of #449, but just a subset to facilitate review (best done commit by commit, I guess). `Compat.collect` ended up here, too, because its README entry was in the syntaxes section.

I've manually verified the deleted tests still pass.


martinholters:
Will merge tomorrow unless anyone objects.

PR #628, opened 2018-08-31
Remove a bunch of definitions only relevant prior to Julia 0.6

martinholters:
I've kept (hopefully) easily reviewable commits. Should be squashed with a commit message listing the individual removals, but omitting the boiler-plate from the individual commit messages.


martinholters:
I'll merge this tomorrow unless anyone objects (or beats me to it).

PR #629, opened 2018-09-10
Remove more pre-0.6-only stuff

martinholters:
This should clean out the last bits from `src/Compat.jl` that have become irrelevant when requiring at least 0.6. (Unless I've missed anything, which is not unlikely). Also removes two (for some time) obsolete README entries. I'm also sure there are tests for stuff that has been removed, but double-testing Base functionality is not too bad, so I haven't gone through `runtests.jl` in detail.

(Should be squashed when merging; the commit messages themselves would be misleading without an appropriate heading.)


martinholters:
I'll merge this later today unless anyone objects (or beats me to it).

Issue #631, opened 2018-09-25
range(start, stop; length, step=1)


Nosferican:
https://github.com/JuliaLang/julia/pull/28708#issuecomment-418433463

Could this be pre-compated?

jlperla:
Spectacular, thanks! Is therein the possibility of a tagged release coming soon with this in it?

martinholters:
https://github.com/JuliaLang/METADATA.jl/pull/18796

PR #633, opened 2018-10-03
range(start, stop; length, step)

Nosferican:
Pre-Compat of #631 per JuliaLang/julia#28708.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,8 +358,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `repmat` is now `repeat` ([#26039])
 
-* `range` supports `range(start, stop; length, step)` (`stop` as positional argument)
-  for Julia 0.7 and 1.0. ([#28708])
+* `range` supports `range(start, stop; length, step)` (`stop` as positional argument) ([#28708])
 
 ## New macros
 
--- a/README.md
+++ b/README.md
@@ -358,6 +358,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `repmat` is now `repeat` ([#26039])
 
+* `range` will support `range(start, stop; length, step)` for Julia 1.0 which
+  will be included in Julia 1.1.
+
 ## New macros
 
 * `@vectorize_1arg` and `@vectorize_2arg` are deprecated on Julia 0.6 in favor
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `Compat.IOBuffer` supporting keyword arguments ([#25873]).
 
-* `Compat.range` supporting keyword arguments ([#25896]).
+* `Compat.range` supporting positional and keyword arguments flavors ([#25896]), ([#28708]).
 
 * `Compat.trunc`, `Compat.floor`, `Compat.ceil`, `Compat.round`, take a keyword argument
   for `base` and `digits`, `Compat.round` also takes `sigdigits` ([#26156], [#26670]).
@@ -357,8 +357,6 @@ Currently, the `@compat` macro supports the following syntaxes:
 * `squeeze` is now `dropdims` ([#28303], [#26660]).
 
 * `repmat` is now `repeat` ([#26039])
-
-* `range` supports `range(start, stop; length, step)` (`stop` as positional argument) ([#28708])
 
 ## New macros
 
--- a/README.md
+++ b/README.md
@@ -358,8 +358,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `repmat` is now `repeat` ([#26039])
 
-* `range` will support `range(start, stop; length, step)` for Julia 1.0 which
-  will be included in Julia 1.1.
+* `range` supports `range(start, stop; length, step)` (`stop` as positional argument)
+  for Julia 0.7 and 1.0. ([#28708])
 
 ## New macros
 
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1839,7 +1839,7 @@ if VERSION < v"0.7.0-beta2.143"
     end
 end
 
-if v"1.0" ≤ VERSION < v"1.1"
+if v"1.0" ≤ VERSION && isempty(methods(range, Tuple{Any,Any}))
     # https://github.com/JuliaLang/julia/pull/28708
     Base.range(start, stop; length::Union{Integer,Nothing}=nothing, step=nothing) =
         Base._range(start, step, stop, length)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1839,7 +1839,7 @@ if VERSION < v"0.7.0-beta2.143"
     end
 end
 
-if v"1.0" ≤ VERSION && isempty(methods(range, Tuple{Any,Any}))
+if v"0.7" ≤ VERSION && isempty(methods(range, Tuple{Any,Any}))
     # https://github.com/JuliaLang/julia/pull/28708
     Base.range(start, stop; length::Union{Integer,Nothing}=nothing, step=nothing) =
         Base._range(start, step, stop, length)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1839,7 +1839,7 @@ if VERSION < v"0.7.0-beta2.143"
     end
 end
 
-if v"0.7" ≤ VERSION && isempty(methods(range, Tuple{Any,Any}))
+if v"0.7" ≤ VERSION
     # https://github.com/JuliaLang/julia/pull/28708
     Base.range(start, stop; length::Union{Integer,Nothing}=nothing, step=nothing) =
         Base._range(start, step, stop, length)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1506,7 +1506,30 @@ end
     const LinRange = Base.LinSpace
     export LinRange
 
-    Compat.range(start; kwargs...) = Base.range(start; kwargs...)
+    function range(start; step=nothing, stop=nothing, length=nothing)
+        have_step = step !== nothing
+        have_stop = stop !== nothing
+        have_length = length !== nothing
+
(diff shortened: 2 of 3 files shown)


src/Compat.jl
```diff
@@ -1839,6 +1839,12 @@ if VERSION < v"0.7.0-beta2.143"
     end
 end
 
+if v"1.0" ≤ VERSION < v"1.1"
```
martinholters: Can't we support this on earlier versions, too? Also, we probably should either wait until https://github.com/JuliaLang/julia/pull/28708 has been merged and we know the exact upper bound, or make this depend on whether the new method is defined. (By looking at `methods(range, Tuple{Any,Any})`, maybe?)

src/Compat.jl
```diff
@@ -1839,6 +1839,12 @@ if VERSION < v"0.7.0-beta2.143"
...
+    # https://github.com/JuliaLang/julia/pull/28708
+    Base.range(start, stop; length::Union{Integer,Nothing}=nothing, step=nothing) =
+        Base._range(start, step, stop, length)
```
martinholters: I don't think we should be calling this internal function. Why not use the old `range` method?

test/runtests.jl
```diff
@@ -1518,4 +1518,9 @@ end
                                    3  4  3  4  3  4]
 @test repeat([1, 2], 1, 2, 3) == [x for x in 1:2, y in 1:2, z in 1:3]
 
+# [1.0, 1.1)
+if v"1.0" ≤ VERSION < v"1.1"
```
martinholters: Even if we only define the method here prior to `v"1.1"`, we should still test it afterwards to make sure it works as promised.

test/runtests.jl
```diff
@@ -1518,4 +1518,9 @@ end
...
+    @test range(0, 5, length = 6) == 0.0:1.0:5.0
+end
```
martinholters: Add a test for `step` given but `length` omitted?

src/Compat.jl
Nosferican: Letting go by https://github.com/JuliaLang/julia/pull/28708#issuecomment-418433463 since it is has already been decided the feature will be added in the next minor release. I can test the `@isdefined` method as well.

src/Compat.jl
Nosferican: I took it from the un merged PR, but good point.

test/runtests.jl
Nosferican: Sure.

test/runtests.jl
Nosferican: Will do.

src/Compat.jl
martinholters: > it is has already been decided the feature will be added in the next minor release

Yes, but we'd need the `XXX` in v"1.1.0-DEV.XXX" for the version bound here.

fredrikekre: I see no reason to merge it now then, why wait?

martinholters: "it" referring to JuliaLang/julia#28708? Agree.

fredrikekre: Right, I rebased it just now.

Nosferican: If we can have it merged now (no more planned patch releases), I can extract the exact version to use here after it has been merged.

src/Compat.jl
Nosferican: Calling the lowering function does make the code simpler than the various cases using the current `range` method.

test/runtests.jl
```diff
@@ -1518,4 +1518,10 @@ end
                                    3  4  3  4  3  4]
 @test repeat([1, 2], 1, 2, 3) == [x for x in 1:2, y in 1:2, z in 1:3]
 
+# [1.0, 1.1)
+if v"1.0" ≤ VERSION && isempty(methods(range, Tuple{Any,Any}))
```
martinholters: The `isempty(methods(...))` test is wrong here. We definitely want the tests to run if the method is defined. Actually, the tests should (implicitly) verify the method is defined.

Nosferican:
```
julia> VERSION
v"1.0.1"

julia> isempty(methods(range, Tuple{Any,Any}))
true

julia> Base.range(start, stop; length::Union{Integer,Nothing}=nothing, step=nothing) =
                  Base._range(start, step, stop, length)

julia> isempty(methods(range, Tuple{Any,Any}))
false
```
I used that to see if it would test for the missing method. Could you point me to in the direction I should go?

README.md
```diff
@@ -358,6 +358,9 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `repmat` is now `repeat` ([#26039])
 
+* `range` will support `range(start, stop; length, step)` for Julia 1.0 which
+  will be included in Julia 1.1.
```
martinholters: Using future tense is not very ... future-proof here. Maybe just "`range` supports `range(start, stop; length, step)` for Julia 1.0" (or "supports `stop` as a positional argument", although that is less clear). Also should link to the Julia PR.

test/runtests.jl
martinholters: Just `v"1.0" ≤ VERSION`? (Although I repeat that I'd prefer this to work from 0.6 on, and I see absolutely no reason why this shouldn't work on 0.7.)

test/runtests.jl
```diff
@@ -1518,4 +1518,10 @@ end
                                    3  4  3  4  3  4]
 @test repeat([1, 2], 1, 2, 3) == [x for x in 1:2, y in 1:2, z in 1:3]
 
+# [1.0, 1.1)
+if v"0.7" ≤ VERSION && isempty(methods(range, Tuple{Any,Any}))
```
martinholters: Still has the bogus `&& isempty(methods(range, Tuple{Any,Any}))`.

Nosferican: What replacement do you recommend given that the PR hasn't been merged?

martinholters: None. We want the tests to be run on all versions (0.7 and up, actually). If you look at the other tests, they don't come with an upper version bound. If Julia is recent enough that Compat doesn't have to do anything, we still test things work as expected/promised.

Nosferican: So that would be
```
if v"0.7" ≤ VERSION
```
then?

martinholters: Yes. (That's what I tried to say in https://github.com/JuliaLang/Compat.jl/pull/633#discussion_r222593330).

src/Compat.jl
martinholters: Doesn't just `Base.range(start, stop; kwargs...) = range(start; stop=stop, kwargs...)` work?

src/Compat.jl
```diff
@@ -1839,6 +1839,11 @@ if VERSION < v"0.7.0-beta2.143"
     end
 end
 
+if v"0.7" ≤ VERSION
```
martinholters: Ah, this should be guarded with `&& isempty(methods(range, Tuple{Any,Any}))`, while the tests should not. However, on 0.7, that does not work, because of the deprecated, old `range(a, len)` methods. That also makes me wonder whether we actually should be adding this method on 0.7. I'd say yes, but might be worth discussing.

src/Compat.jl
```diff
@@ -1839,6 +1839,11 @@ if VERSION < v"0.7.0-beta2.143"
...
+    # https://github.com/JuliaLang/julia/pull/28708
+    Base.range(start, stop; kwargs...) = range(start; stop=stop, kwargs...)
```
martinholters: Heh, thanks to #511, this definition actually works on 0.6, too. But there we have the keyword-argument-less, two-argument methods, too. However, if we omit the `Base` qualification, then this will add a method to the existing `Compat.range` on 0.6 and to `Base.range` on 0.7 and later. Doesn't sound too bad.

martinholters:
So condensing my comments, this definition might be worth a try:
```julia
if VERSION ≤ v"1.0" || isempty(methods(range, Tuple{Any,Any}))
    range(start, stop; kwargs...) = range(start; stop=stop, kwargs...)
end
```
However, that would require the code from #511 to be adjusted to only import `range` after the removal of the deprecated two-arg methods, and define `Compat.range(start; kwargs...) = Base.range(start; kwargs...)` (or similar) otherwise. Then the new method would be accessible as `Compat.range` (and just `range` on 1.0 and above, though I'm not sure we want to advertise that).

src/Compat.jl
```diff
@@ -1527,6 +1527,9 @@ end
             return linspace(start, stop, length)
         end
     end
+elseif VERSION < v"1.0.0-DEV.57"
+    import LinRange
```
martinholters: Sorry, this should be `import Base: LinRange`.

test/runtests.jl
martinholters: The condition should be completely removed now (assuming we succeed in making this work for 0.6).

README.md
```diff
@@ -358,6 +358,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `repmat` is now `repeat` ([#26039])
 
+* `range` supports `range(start, stop; length, step)` (`stop` as positional argument) ([#28708])
```
martinholters: This should be `Compat.range`.

martinholters: Hm, could also be merged with
https://github.com/JuliaLang/Compat.jl/blob/a535c0ae0094241c7714be390969b8c8954ce23a/README.md#L204
but I'm not sure of a good wording.

Nosferican:
I haven't. I would appreciate if you can check against it and let me know (I mostly use the binaries and don't build it from source).

arnavs:
Any progress on this?

martinholters:
Thanks for the bump, I had gotten distracted by other things. So I've successfully checked against JuliaLang/julia#28708 just now, nobody has voiced any objections... Merge!

arnavs:
Thanks so much @martinholters! And appreciate your tagging a new release so we can use the feature.

Issue #640, opened 2019-01-12
`range(start, stop)` without kwargs accidentally allowed


martinholters:
https://github.com/JuliaLang/Compat.jl/blob/ce5d74bbd4b1eced53892a6d6ac6857ce1af53e8/src/Compat.jl#L1853-L1855

So before `1.1.0-DEV.506`, `range(start, stop)` will become `range(start, stop=stop)`, which works, using a default step of 1. On `1.1.0-DEV.506` or later, the Base definition of `range(start, stop)` errors, requiring `step` or `length` to be specified. Unfortunately, ONNX.jl already makes use of the `range(start, stop)` form, making it at preset incompatible with the upcoming julia-1.1.0 (ref. https://github.com/JuliaLang/julia/issues/30374#issuecomment-450448883).

Not sure what to do here. In the long run, we should make `range(start, stop)` error on 1.0, too, but that would be a breaking change. (Altough, we nowhere claim it to work, so one might argue it's really a bugfix). In the short term... Should we override the Base method to allow `range(start, stop)`? That feels so ... wrong.

martinholters:
So the plan is to 
- [ ] override the Base method to allow `range(start, stop)` (#642)
- [ ] tag a release
- [ ] deprecate `range(start, stop)` (and, while at it, remove old deprecations and deprecate any other stuff where Compat has diverged from Base, if any)
- [ ] tag a breaking (i.e. major) release.

PR #642, opened 2019-01-18
Always define `range(start, stop)`

martinholters:
Ref. #640 and https://github.com/JuliaLang/julia/issues/30374#issuecomment-454760918. I aim at implementing option 3 mentioned in the latter, i.e. tag a release after this has been merged, then remove the offending functionality and tag another release.

Downside here is
```
WARNING: Method definition range(Any, Any) in module Base at range.jl:91 overwritten in module Compat at /home/travis/build/JuliaLang/Compat.jl/src/Compat.jl:1857.
WARNING: Method definition #range(Any, typeof(Base.range), Any, Any) in module Base overwritten in module Compat.
```
of course.


martinholters:
Thinking about this again, it would be nice to deprecate `range(start, stop)` in the next major version (instead of making it error directly), but then these warnings would be really unfortunate, so I've updated with the `@eval Base` version.

@ararslan, any comments?

src/Compat.jl
```diff
@@ -1860,8 +1860,18 @@ if VERSION < v"0.7.0-beta2.143"
...
 
+# This definition should be modified to throw an ArgumentError if neither
+# `step` nor `length` are given and be limited to VERSION < v"1.1.0-DEV.506".
+# However, there is a release with this definition, so we need to keep it around
+# to avoid breakage.
 if VERSION < v"1.1.0-DEV.506"
     range(start, stop; kwargs...) = range(start; stop=stop, kwargs...)
+else
+    function __init__()
+        @eval Base begin
```
ararslan: This is really heinous... I think at one point @yuyichao mentioned that it's bad for precompilation as well, though I fully admit to not understanding the details on why.

martinholters: Doing `@eval Base` outside `__init__` runs afoul of precompilation (and gives a warning to that effect).

ararslan: So it's legit to do in `__init__`, just not in the package body?

ararslan: > to the extent it's ever legit to `eval` into Base

...which I thought was never

martinholters: In any case, I haven't found a better way of achieving the (short-term) goal of providing `range(start, stop)` on Julia 1.1.

ararslan: Yeah, I'm hesitant about this approach but I don't have a better suggestion.

martinholters: Great, so we're in agreement here. Who else might have something to contribute here that could bring is to a decision?

martinholters: @StefanKarpinski, https://github.com/JuliaLang/julia/issues/30374#issuecomment-454834238 means you have to chime in here :smile:

martinholters:
Barring objections, I think I'll merge this tomorrow. Even if the implementation has flaws (and I'm not saying it has, just that it might), it's still a step forward IMO.

PR #646, opened 2019-01-25
Remove old deprecations, deprecate obsolete stuff

### LINKED ISSUE #640: `range(start, stop)` without kwargs accidentally allowed

martinholters:
Closes #640, which was the motivation to do this: When doing a breaking change, why not do all pending ones at once (in anticipation of the required major version bump).

Note that the removed tests passed with the deprecations in place (i.e. run against ~666460f~ 
~dc73ae6~ 80afcef).

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,16 +366,6 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New macros
 
-* `@vectorize_1arg` and `@vectorize_2arg` are deprecated on Julia 0.6 in favor
-  of the broadcast syntax ([#17302]). `Compat.@dep_vectorize_1arg` and
-  `Compat.@dep_vectorize_2arg` are provided so that packages can still provide
-  the deprecated definitions without causing a depwarn in the package itself
-  before all the users are upgraded.
-
-  Packages are expected to use this until all users of the deprecated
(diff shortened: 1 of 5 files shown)


martinholters:
I guess we should wait a bit (a week or so?) whether v1.5.1 shows any more problems that need fixing, then merge this, heading towards v2.0. Objections?

PR #647, opened 2019-01-28
Add missing members to `Compat.Random`

martinholters:
These (EDIT: `Random.RangeGenerator` and `Random.GLOBAL_RNG`) are used (at least) by `StatsBase`. A more thorough fix would probably include #440, but I'd like to get back to a usable state quickly.

CC @rdeits, @malmaud

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -646,6 +646,8 @@ else
         if VERSION < v"0.7.0-DEV.3406"
             $((:(using Base.Random: $f) for f in random_fields)...)
             const seed! = Base.Random.srand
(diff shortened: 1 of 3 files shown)


martinholters:
Ah, and `IJulia` uses `uuid4`, so while at it, I've also added `uuid1`, `uuid4`, and `uuid_version`, but only on Julia versions where these are part of `Base.Random` or `Random`, i.e. were available as `Compat.Random` before #601. The plan is to deprecate them to there `UUIDs.*` counterparts.

martinholters:
AV failure is a network issue in one of the jobs.

PR #649, opened 2019-01-30
Define range(::Number, ::Number) instead of overwriting Base

### LINKED ISSUE #648: pre-compilation generating warnings
```
WARNING: eval into closed module Base:
Expr(:block, #= Symbol("/home/savastio/.julia/packages/Compat/hsya0/src/Compat.jl"):1891 =#, Expr(:call, :range, Expr(:parameters, Expr(:..., :kwargs)), :start, :stop) = Expr(:block, #= Symbol("/home/savastio/.julia/packages/Compat/hsya0/src/Compat.jl"):1891 =#, Expr(:call, :range, Expr(:parameters, Expr(:kw, :stop, :stop), Expr(:..., :kwargs)), :start)))
  ** incremental compilation may be broken for this module **

WARNING: Method definition range(Any, Any) in module Base at range.jl:90 overwritten at /home/savastio/.julia/packages/Compat/hsya0/src/Compat.jl:1891.
WARNING: Method definition #range(Any, typeof(Base.range), Any, Any) in module Base overwritten.
```
Is this a Compat issue?

fredrikekre:
Define range(::Number, ::Number) instead of overwriting the (::Any, ::Any) method in base, fixes #648.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1886,11 +1886,9 @@ end
 if VERSION < v"1.1.0-DEV.506"
     range(start, stop; kwargs...) = range(start; stop=stop, kwargs...)
 else
-    function __init__()
-        @eval Base begin
-            range(start, stop; kwargs...) = range(start; stop=stop, kwargs...)
-        end
-    end
+    # This method is restricted to Number, since we don't
+    # want to overwrite the (::Any, ::Any) method in Base.
+    range(start::Number, stop::Number; kwargs...) = range(start; stop=stop, kwargs...)
 end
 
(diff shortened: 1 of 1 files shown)


nalimilan:
This won't work for e.g. Dates, but I guess that's OK. We could also define a function for them.

martinholters:
IIRC only ONNX had been seen in the wild using this. But I don't know whether it uses anything else than numbers.

fredrikekre:
IIUC this method should not be called anyway, ref #640 and perhaps this is good enough then?

PR #650, opened 2019-02-05
Add hasfield and hasproperty

tkf:
This PR backports `hasfield` and `hasproperty` introduced https://github.com/JuliaLang/julia/pull/30496 for fixing https://github.com/JuliaLang/julia/issues/28850.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1896,7 +1896,7 @@ if VERSION < v"1.2.0-DEV.272"
     Base.@pure hasfield(::Type{T}, name::Symbol) where T =
         Base.fieldindex(T, name, false) > 0
     export hasfield
-    if VERSION >= v"0.7-"
+    if VERSION >= v"0.7-DEV.5136"
         hasproperty(x, s::Symbol) = s in propertynames(x)
         export hasproperty
     end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1895,8 +1895,11 @@ end
 if VERSION < v"1.2.0-DEV.272"
     Base.@pure hasfield(::Type{T}, name::Symbol) where T =
         Base.fieldindex(T, name, false) > 0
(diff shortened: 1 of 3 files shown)


martinholters:
Could `hasproperty` just return `false` on Julia 0.6?

tkf:
Isn't it strange to define `hasproperty` when there is no `setproperty!` or `getproperty`?  In 0.6-compatible code, I think it's better to stick with `getfield`/`setfield!`/`hasfield`.

src/Compat.jl
```diff
@@ -1891,6 +1891,17 @@ else
     range(start::Number, stop::Number; kwargs...) = range(start; stop=stop, kwargs...)
 end
 
+# https://github.com/JuliaLang/julia/pull/30496
+if VERSION < v"1.2.0-DEV.272"
+    Base.@pure hasfield(::Type{T}, name::Symbol) where T =
+        Base.fieldindex(T, name, false) > 0
+    export hasfield
+    if VERSION >= v"0.7-"
```
```suggestion
    if VERSION >= v"0.7-DEV.5136"
```
That's the version that exports `propertynames`. But doesn't really matter. Whether Compat defines `hasproperty` in a way that errors or not at all on early `0.7-DEV` is unlikely to ever make a difference.

I do wonder, however, whether we should add `hasproperty(x, s::Symbol) = hasfield(typeof(x), s)` for old versions. (Or maybe `propertynames(x) = fieldnames(typeof(x))`.) Then one can use `hasproperty` on any version to check whether `x.s` is expected to work.

tkf: Now that both 0.7 and 0.6 are unmaintained, is it worth complicating Compat.jl?  Also, isn't it better to wait until someone requests it (or rather implements it and adds tests for it)?

martinholters: True. I think it's still a bit early to drop the 0.6-stuff, but not worth worrying too much when adding new things.

PR #652, opened 2019-02-09
Drop support for Julia < 0.7

### LINKED ISSUE #651: Compat: Forging ahead in the 1.x era
Now that Julia versions 0.6 and 0.7 are [officially unmaintained](https://discourse.julialang.org/t/julia-0-6-and-0-7-are-now-officially-unmaintained/20449), I would propose that we drop support for versions prior to 1.0 and focus on providing features from 1.1 and later to 1.0. That would allow Compat to be simplified tremendously, and may help spur developer interest in contributing, since no one cares (or wants to care) about 0.6 anymore.

ararslan:
The time has come, as 0.6 is unmaintained.

Fixes #651.


martinholters:
On the phone right now, so cannot review properly, but looks like you are already removing the deprecations from #646 here which haven't been active on any release, yet. That would mean we tag 2.0 now and then 3.0 after this one lands (eventually). Is that your plan? Alternatively, we could keep the deprecations and tag 2.0 after meeting this one.

ararslan:
Yeah I figured we'd do another release with deprecations before this lands.

martinholters:
Tag `v2.0.0` now, and leave this open for a while (weeks to months)? Or do a `v3.0.0` with this soon after (similar to julia 0.7/1.0)? I wouldn't like the latter; I don't think anyone would consciously upgrade to 2.0 first, fix deprecations, then update to 3.0.

The former would be ok with me, although I think I'd prefer to to do the dropping of 0.6 now for 2.0. Then packages (or package versions) which don't support julia later than 0.6 don't have to worry about capping their required Compat version.

src/Compat.jl
```diff
@@ -1,1840 +1,55 @@
...
+import Statistics
+import SuiteSparse
+import Test
+import UUIDs
+import Unicode
+
+# Bindings that were available at the top level in Compat
+using Base: notnothing
+using Dates: AbstractDateTime
+using LinearAlgebra: dot, norm, opnorm, qr, rmul!, tr
```
martinholters: Should we keep these? I guess we could safely deprecate them.

martinholters:
I wonder what we should do with all the `import`ed stdlibs. There is not much reason to keep them at this point (i.e. we could deprecated `Compat.XYZ` to `XYZ`). However, if we backport additions to stdlibs, we would need to re-introduce (and extend) them. Should we therefore just keep them?

src/Compat.jl
```diff
@@ -1,1840 +1,55 @@
...
-end
-
-# https://github.com/JuliaLang/julia/pull/22646
-if VERSION < v"0.7.0-DEV.1139"
-    function invokelatest(f, args...; kwargs...)
-        inner() = f(args...; kwargs...)
-        Base.invokelatest(inner)
-    end
-else
-    import Base.invokelatest
```
martinholters: Should add a deprecation for this binding.

src/Compat.jl
```diff
@@ -1,1840 +1,55 @@
...
-    Base.ceil(x::Compat.TimeTypeOrPeriod, ::Type{P}) where P <: Compat.Dates.Period = Base.ceil(x, oneunit(P))
-    function Base.round(x::Compat.TimeTypeOrPeriod, ::Type{P}, r::RoundingMode=RoundNearestTiesUp) where P <: Compat.Dates.Period
-        return Base.round(x, oneunit(P), r)
-    end
-end
-
-if VERSION < v"0.7.0-DEV.3216"
-    const AbstractDateTime = Compat.Dates.TimeType
-else
-    const AbstractDateTime = Compat.Dates.AbstractDateTime
```
martinholters: Deprecation.

src/Compat.jl
```diff
@@ -1,1840 +1,55 @@
...
-    import Base: notnothing
-end
-
-# 0.7.0-DEV.3309
-@static if VERSION < v"0.7.0-DEV.3309"
-    const IteratorSize = Base.iteratorsize
-    const IteratorEltype = Base.iteratoreltype
-else
-    const IteratorSize = Base.IteratorSize
-    const IteratorEltype = Base.IteratorEltype
```
martinholters: Deprecations for these two.

src/Compat.jl
martinholters: On second thought, we can deprecate these in another PR (if we want to), and can keep this PR strictly to dropping pre-0.7 support.

ararslan:
> However, if we backport additions to stdlibs, we would need to re-introduce (and extend) them

Yeah, that is an issue and I'm not sure how best to go about that aside from what we currently do, which is define `Compat.X` for each `X` as an actual submodule of Compat, which kind of sucks...

musm:
👍    Didn't you plan to also drop 0.7 since it is unmaintained?

ararslan:
Oh, good call. I should do that as well, thanks @musm.

nickrobinson251:
Bump? 🙏

ararslan:
I don't know when I'll have time to revisit this, so I encourage anyone who's interested to take the reins and use this branch for reference.

README.md
```diff
@@ -40,357 +40,40 @@ and then as needed add
...
-* `using Compat.Sockets` is provided on versions older than 0.7, where this library is
-  not yet part of the standard library ([#25935])
-
-* `using Compat.Statistics` is provided on versions older than 0.7, where this library is
-  not yet part of the standard library ([#27834]).
+`Compat.X` is available for every stdlib package `X` for backwards compatibility with
+previous verisons of Compat.
+Newer code should use `X` directly.
 
 ## New functions, macros, and methods
```
mcabbott: It might be helpful to state the order here, and organise by what version of Julia supports the feature:
```suggestion
Latest addions are at the top of each list:
### New in Julia 1.4

### New in Julia 1.3

### New in Julia 1.2

### New in Julia 1.1
```

martinholters:
Superseded by #665.

PR #655, opened 2019-04-10
Add Project.toml

martinholters:
This adds a `Project.toml` (again) as it will be needed for registering future versions. While at it, I noticed that we had a `test/REQUIRE` file pulling in `OffsetArrays` which wasn't used anymore, so I've removed that.

I've used the script from https://github.com/JuliaLang/Pkg.jl/pull/1133 but curiously had to fix to dependency UUIDs, cc @StefanKarpinski, see second commit. Note that I've used the script, but the Pkg version from `1.2.0-DEV.629`, not Pkg master, if that's a problem.

I've set the version to `2.1.1-DEV` here, expecting it to be bumped to `2.2.0-DEV` when the first new feature is added and then to `2.2.0` for the next release. (Assuming we're not doing a bugfix release 2.1.1.) The commit that sets version to `2.2.0` should then be registered and the very next commit bump the version to `2.2.1-DEV`. That workflow sounds tedious and error-prone, so suggestions welcome.


fredrikekre:
> I've used the script from JuliaLang/Pkg.jl#1133 but curiously had to fix to dependency UUIDs

Those don't have METADATA compatible UUIDs so should be special cased in the script.

> I've set the version to 2.1.1-DEV here

Just set it to 2.1.0 IMO and then when the next release is about to happen we bump to 2.2.0 or 2.2.1 or w/e is appropriate.

Project.toml
```diff
@@ -0,0 +1,30 @@
...
+SharedArrays = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
+Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
+Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
+
+[compat]
+julia = "≥ 0.6.0"
```
fredrikekre:
```suggestion
julia = "0.6,0.7,1"
```

Project.toml
```diff
@@ -0,0 +1,30 @@
+name = "Compat"
+uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
+version = "2.1.1-DEV"
```
fredrikekre:
```suggestion
version = "2.1.0"
```

martinholters:
> Just set it to 2.1.0 IMO and then when the next release is about to happen we bump to 2.2.0 or 2.2.1 or w/e is appropriate.

Right, that seems much simpler.

Issue #639, opened 2019-01-09
eachrow, eachcol, eachslice support


oxinabox:
For people supporting 1.1 and 1.0,
this is useful.

https://github.com/JuliaLang/julia/issues/29749

jiegillet:
I'd be happy to do a PR for this, as my package already uses `Compat` for `isnothing`, but has an ugly `if VERSION < v"1.1"` thing in it for `eachrow`. 
I think I can mostly figure out how to get started on the PR, but I don't know how to get the `v"1.1.0-DEV.XXX"` precise number from the PR listed above.

oxinabox:
> I don't know how to get the v"1.1.0-DEV.XXX" precise number from the PR listed above.

Don't bother, noone is using `-DEV` versions of `1.1`
having it down to the exact build number is only useful for while that version is at HEAD.

martinholters:
Fixed by #658.

PR #658, opened 2019-07-24
Added eachrow/col/slice for v1.0

jiegillet:
Let's get the ball going for #639.

This won't work at all for 0.6 from what I saw on JuliaBox, too many things missing (`axes`, `OneTo`...).  0.6 support is getting dropped though, right? 

Also, I don't know how to get the exact version where these were defined. It was this PR [JuliaLang/julia#29749](https://github.com/JuliaLang/julia/pull/29749)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 ## New functions, macros, and methods
 
 * `eachcol` to iterate over the second dimension of a matrix and `eachrow` to iterate over the first dimension of a matrix and `eachslice` to iterate over some dimension of a matrix ([#29749]).
+
 * `isnothing` for testing if a variable is equal to `nothing` ([#29674]).
 
 * `Compat.readline` with `keep` keyword argument ([#25646])
--- a/README.md
+++ b/README.md
@@ -114,6 +114,12 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `eachslice` to iterate over some dimension of a matrix ([#29749]).
+
+* `eachcol` to iterate over the second dimension of a matrix ([#29749]).
+
+* `eachrow` to iterate over the first dimension of a matrix ([#29749]).
+
 * `isnothing` for testing if a variable is equal to `nothing` ([#29674]).
 
 * `Compat.readline` with `keep` keyword argument ([#25646])
@@ -570,3 +576,4 @@ includes this fix. Find the minimum version from there.
 [#28850]: https://github.com/JuliaLang/julia/issues/28850
 [#29259]: https://github.com/JuliaLang/julia/issues/29259
 [#29674]: https://github.com/JuliaLang/julia/issues/29674
+[#29749]: https://github.com/JuliaLang/julia/issues/29749
--- a/README.md
+++ b/README.md
@@ -114,12 +114,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
-* `eachslice` to iterate over some dimension of a matrix ([#29749]).
-
-* `eachcol` to iterate over the second dimension of a matrix ([#29749]).
-
-* `eachrow` to iterate over the first dimension of a matrix ([#29749]).
-
+* `eachcol` to iterate over the second dimension of a matrix and `eachrow` to iterate over the first dimension of a matrix and `eachslice` to iterate over some dimension of a matrix ([#29749]).
 * `isnothing` for testing if a variable is equal to `nothing` ([#29674]).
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -114,6 +114,12 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `eachslice` to iterate over some dimension of a matrix ([#29749]).
+
+* `eachcol` to iterate over the second dimension of a matrix ([#29749]).
+
+* `eachrow` to iterate over the first dimension of a matrix ([#29749]).
```
martinholters: Maybe consolidate these into a single item?

src/Compat.jl
```diff
@@ -931,6 +931,21 @@ if VERSION < v"1.1.0-DEV.472"
     isnothing(::Nothing) = true
 end
 
+# https://github.com/JuliaLang/julia/pull/29749
+if VERSION < v"1.1.0"
```
```suggestion
if VERSION < v"1.1.0-DEV.792"
```
(Obtained with `./contrib/commit-name.sh 6b04291` within a jula source tree, where `6b04291` is the respective merge commit.)

test/runtests.jl
```diff
@@ -30,6 +30,21 @@ end
 @test !isnothing(1)
 @test isnothing(nothing)
 
+# https://github.com/JuliaLang/julia/pull/29749
+@testset "row/column/slice iterators" begin
```
martinholters: Wrap the tests in `if VERSION > v"0.7"` to keep CI happy (or work out how to make this work on 0.6).

src/Compat.jl
```diff
@@ -931,6 +931,21 @@ if VERSION < v"1.1.0-DEV.472"
...
+    export eachrow, eachcol, eachslice
+	eachrow(A::AbstractVecOrMat) = (view(A, i, :) for i in axes(A, 1))
```
```suggestion
    eachrow(A::AbstractVecOrMat) = (view(A, i, :) for i in axes(A, 1))
```

src/Compat.jl
jiegillet: OK! Should I also change it to `v"7.0" >= VERSION < v"1.1.0-DEV.792"` since `6.0` won't work with it?

src/Compat.jl
```diff
@@ -936,7 +936,7 @@ if v"0.7" <= VERSION < v"1.1.0-DEV.792"
     export eachrow, eachcol, eachslice
     eachrow(A::AbstractVecOrMat) = (view(A, i, :) for i in axes(A, 1))
     eachcol(A::AbstractVecOrMat) = (view(A, :, i) for i in axes(A, 2))
-    @inline function eachslice(A::AbstractArray; dims)
```
martinholters: Is this change sufficient to make this work on 0.6? Then you could remove the "v0.7" lower bound, right? Otherwise, this change seems to be pointless, or am I missing something?

Also, the `Base` function errors if `dims` is not given, so we should do the same here, e.g. by using `dims=error("keyword argument dims not assigned")`, or better yet, utilizing `@compat` (see #586).

src/Compat.jl
martinholters: Good point, although the keyword argument part is a parser thing, so needs to be `@static if` then, unless I'm mistaken.

src/Compat.jl
jiegillet: No, it still doesn't work on 0.6, it's simply skipped altogether. It seemed complicated to make it work, and I don't think it's worth it. The reason I added `dims = 1` is that the 0.6 parser would not accept the syntax and crash, even though the function would never been called.

`@compat` is much better indeed.

src/Compat.jl
jiegillet: I didn't know about `@static`, thank you. This bypasses the `dims` thing altogether.

README.md
```diff
@@ -114,6 +114,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `eachcol` to iterate over the second dimension of a matrix and `eachrow` to iterate over the first dimension of a matrix and `eachslice` to iterate over some dimension of a matrix ([#29749]).
```
martinholters: Maybe more compactly like so?
```suggestion
* `eachrow`, `eachcol`, and `eachslice` to iterate over first, second, or given dimension
  of an array ([#29749]).
```
No that it matters much...

test/runtests.jl
```diff
@@ -30,6 +30,23 @@ end
 @test !isnothing(1)
 @test isnothing(nothing)
 
+# https://github.com/JuliaLang/julia/pull/29749
+if VERSION > v"0.7"
```
martinholters:
```suggestion
if VERSION >= v"0.7"
```
or the tests won't be executed on 0.7.

PR #660, opened 2019-09-11
Implement `parameter_upper_bound` directly

vtjnash:
Compat should not be relying on accessing internal functionality from Base

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -93,10 +93,21 @@ end
 end
 
 module TypeUtils
-    using Base: parameter_upper_bound, typename
+    using Base: typename
     using Compat: isabstracttype
     const isabstract = isabstracttype
     export isabstract, parameter_upper_bound, typename
+    """
+        parameter_upper_bound(t::UnionAll, idx)
+
+    Determine the upper bound of a type parameter in the underlying datatype.
+    This method should generally not be relied upon:
+    code instead should usually use static parameters in dispatch to extract these values.
+    """
+    function parameter_upper_bound(t::UnionAll, idx)
+        Base.@_pure_meta
+        return Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)
+    end
 end # module TypeUtils
 
 # https://github.com/JuliaLang/julia/pull/25646
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -83,13 +83,29 @@ eval(Expr(
 @test !Compat.TypeUtils.isabstract(ConcreteFoo20006N)
 @test !Compat.TypeUtils.isabstract(ConcreteFoo200061)
 @test !Compat.TypeUtils.isabstract(StridedArray)
-@test Compat.TypeUtils.parameter_upper_bound(ConcreteFoo20006, 1) == Int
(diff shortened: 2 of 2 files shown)


martinholters:
`parameter_upper_bound` is now deprecated, but in favor of `Base.parameter_upper_bound`:
```
WARNING: TypeUtils.parameter_upper_bound is deprecated, use Base.parameter_upper_bound instead.
```
Is it worth changing the deprecation to the definition (`Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)`)? Or can we just close this?

vtjnash:
The function is universally deprecated (as it says in the doc string), but perhaps better to live here?

martinholters:
> perhaps better to live here?

Not in the long term. Generally, `Compat` should make the replacement available in older Julia versions. Here, there is no direct replacement, so we had the `TypeUtils` module to ease transitioning. Now, the point is to get rid of it, but with a suitable deprecation (if at all possible). So the easiest thing seems to be deprecating in favor of the one-line definition. Alternatively, we might consider giving a deprecation that suggests to "use static parameters in dispatch to extract these values" as the Base docstring does.

martinholters:
Um... Is it kosher to use `depwarn` in a `@pure` function? I mean, depwarn is of course not at all pure, but is it non-pure in a harmful way? That is, is the following ok?
```julia
Base.@pure function parameter_upper_bound(t::UnionAll, idx)
    Base.depwarn("`parameter_upper_bound(t::UnionAll, idx)` is deprecated, use `Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)` instead.", :parameter_upper_bound)        
    return Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)
end
```
If not, is it any better to go via `@deprecate_binding`? E.g.
```julia
Base.@pure function _parameter_upper_bound(t::UnionAll, idx)
    return Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)
end
Base.@deprecate_binding parameter_upper_bound _parameter_upper_bound true ", use `Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)` instead."
```

vtjnash:
Yes, I meant only in the short term. The function is exported here and not in Base, so that code is being deleted from Base once this goes away from Compat (https://github.com/JuliaLang/julia/pull/32427). Thus it's actively misleading to suggest folks to access it there, so printing out the information on how to write the code correctly (from the doc string) would be better.

Yeah, doing IO would be a very bad idea. The `@deprecate_binding` approach is OK. Perhaps also OK just to remove the `@pure`, since the user code needs to be fixed anyways when upgrading the Compat dependency.

martinholters:
Ok, PR for the deprecation change in #671.
As the next release will bump the major version, I wonder whether we should do a 2.2.1 from a yet-to-be-created `release-2` branch, where we merge this PR. Otherwise, packages declaring compatibility with `Compat = "2"` might prevent updating `Compat` for too long, causing breakage once JuliaLang/julia#32427 lands; no idea how much of a concern that is in practice.

PR #661, opened 2019-09-26
mod(n, range)

mcabbott:
There are lots of recent small additions to Julia I’d like to be able to use on 1.2. Tell me if I’m doing this right, and then I’ll do a few more?

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1867,7 +1867,7 @@ if v"0.7.0" <= VERSION < v"1.1.0-DEV.594"
 end
 
 # https://github.com/JuliaLang/julia/pull/32628
-if v"0.7.0" <= VERSION < v"1.3.0-DEV.546"
+if v"0.7.0" <= VERSION < v"1.3.0-alpha.8"
     Base.mod(i::Integer, r::Base.OneTo) = mod1(i, last(r))
     Base.mod(i::Integer, r::AbstractUnitRange{<:Integer}) = mod(i-first(r), length(r)) + first(r)
 end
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -1866,6 +1866,12 @@ if v"0.7.0" <= VERSION < v"1.1.0-DEV.594"
     Base.merge(a::NamedTuple) = a
 end
 
+# https://github.com/JuliaLang/julia/pull/32628
+if v"0.7.0" <= VERSION < v"1.3.0-DEV.546"
```
martinholters: I get a different version for the merge commit:
```
$ ./contrib/commit-name.sh 21823899cc1724871a8e679287a94bb4acee72d7
1.3.0-alpha.8
```
Can you double-check?

mcabbott: That’s better, It looks like I picked the last of my commits:
```
$ contrib/commit-name.sh 21823899cc1724871a8e679287a94bb4acee72d7
1.3.0-alpha.8
$ contrib/commit-name.sh f93604aa24ade37fa4ca7a1c0446894dd27543ee
1.3.0-DEV.546
```

martinholters:
Thanks for the contribution!

The list of links at the end of the README needs to be updated, but we can do that in a seperate PR. Just shouldn't forget before tagging.

mcabbott:
Ah I didn’t scroll down that far. What’s the list for? (I also see that I failed to make the README line link to the issue.)

I find the Readme pretty confusing BTW, wasn’t sure whether to add at the top or bottom, apologies if I made it worse. I guess #652 will help a lot.

PR #663, opened 2019-10-08
EachRow/EachCol types (see #32310)

simonbyrne:
Obviously still needs the PR to be merged, but this is just to show we can monkey-patch the aliases to support older versions of Julia.

cc: @nalimilan

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1903,6 +1903,17 @@ if v"0.7.0" <= VERSION < v"1.3.0-alpha.8"
     Base.mod(i::Integer, r::AbstractUnitRange{<:Integer}) = mod(i-first(r), length(r)) + first(r)
 end
 
+# https://github.com/JuliaLang/julia/pull/32310
+if v"0.7.0" <= VERSION
+    const EachRow{A,I} = Union{Base.Generator{I,(typeof(eachrow(ones(Int32,2,2)).f).name.wrapper){A}}, Base.Generator{I,(typeof(eachrow(ones(Int32,2,2)).f).name.wrapper){A}}
+    const EachCol{A,I} = Base.Generator{I,(typeof(eachcol(ones(Int32,2,2)).f).name.wrapper){A}}
+                               
+    Base.parent(x::EachRow) = x.f.A
+    Base.parent(x::EachCol) = x.f.A
+end    
+
+
+
 include("deprecated.jl")
 
 end # module Compat
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1554,4 +1554,17 @@ if VERSION >= v"0.7"
     end
 end
 
+if VERSION >= v"0.7"
+    @testset "EachRow/EachSlice" begin
+        X = rand(3,3)
+        @test eachrow(X) isa EachRow
+        @test !(eachcol(X) isa EachRow)
+        @test eachcol(X) isa EachCol
+        @test !(eachrow(X) isa EachCol)
+
+        @test parent(eachrow(X)) === X
+        @test parent(eachcol(X)) === X
+    end
+end
+
 nothing


src/Compat.jl
```diff
@@ -1903,6 +1903,17 @@ if v"0.7.0" <= VERSION < v"1.3.0-alpha.8"
     Base.mod(i::Integer, r::AbstractUnitRange{<:Integer}) = mod(i-first(r), length(r)) + first(r)
 end
 
+# https://github.com/JuliaLang/julia/pull/32310
+if v"0.7.0" <= VERSION
+    const EachRow{A,I} = Union{Base.Generator{I,(typeof(eachrow(ones(Int32,2,2)).f).name.wrapper){A}}, Base.Generator{I,(typeof(eachrow(ones(Int32,2,2)).f).name.wrapper){A}}
```
```suggestion
    const EachRow{A,I} = Union{Base.Generator{I,(typeof(eachrow(ones(Int32,2,2)).f).name.wrapper){A}}, Base.Generator{I,(typeof(eachrow(ones(Int32,2,2)).f).name.wrapper){A}}}
```

oschulz:
Could we also define `Compat.AbstractSlices` as part of this to "backport" `Base.AbstractSlices` (see https://github.com/JuliaLang/julia/pull/32310#issuecomment-1124631069)?

oschulz:
We could also add

```julia
const EachSlice{A,I} = Base.Generator{I,<:(typeof(eachslice(ones(Int32,2,2), dims = 1).f).name.wrapper){A}}
Base.parent(x::EachSlice) = x.f.A
```

to add support for "legacy" `eachslice`, in addition to "legacy" `eachrow` and `eachcol`. Seems to work fine.

martinholters:
With https://github.com/JuliaLang/julia/pull/32310 finally merged, this could now be picked up again if there is still interest.

oschulz:
I think we should definitely do this. Ideally with `AbstractSlices` and `EachSlice` as well?

nalimilan:
Adding `AbstractSlices` in the spirit of this PR isn't possible AFAICT as on Julia < 1.9 `eachrow` and `eachcol` return `Generator` objects, which cannot be made to inherit from `AbstractSlices`. What can be done is adding unexported `eachrow`, `eachcol` and `eachslice` functions that would be copies of what has been added at https://github.com/JuliaLang/julia/pull/32310, i.e returning `EachSlice <: AbstractSlices` objects.

Probably the best would be to implement both approaches. The former is useful for packages that want to use a special method when users call `f(eachcol(x))` on any Julia version. The latter is useful if they want to dispatch on `AbstractSlices`, and/or call `Compat.eachslice` themselves.

oschulz:
> Adding AbstractSlices in the spirit of this PR isn't possible AFAICT

Ah - I don't want to add it to have eachrow and eachcol as subtypes - that's not possible, I agree. But I think it's important to add it in Compat so that packages like ArraysOfArrays.jl and JuliennedArrays.jl can make their types subtypes of AbstractSlices while still supporting Julia LTS. And so we can do something like

```
@static if isdefined(Base, :Slices)
    const SlicedArrayLike = AbstractSlices
else
    const SlicedArrayLike = Union{EachRow,EachCol,EachSlice,AbstractSlices}
end
```

in ArrayInterfaceCore (prototyping this [here](https://github.com/oschulz/ArrayInterface.jl/blob/sliced-arrays/lib/ArrayInterfaceCore/src/sliced_arrays.jl)). That will then allow code that can take advantage of slices array to dispatch on `SlicedArrayLike`, which will cover the legacy `eachrow` and `eachcol`, as well as the Julia v1.9 `Slices` and subtypes of `AbstractSlices`.

For that to work, we need an `AbstractSlices` that's available from Julia v1.6 on, via Compat.

mcabbott:
1.9alpha is out. What's the plan here?

oschulz:
Bump - is there support for adding this to Compat?

PR #665, opened 2019-10-09
Drop support for Julia versions <1.0

### LINKED ISSUE #651: Compat: Forging ahead in the 1.x era

martinholters:
Closes #651, replaces #652.

This removes anything that was relevant only for Julia older than 1.0. It also deprecates a bunch of bindings that were present in `Compat` also for current Julia (e.g. `Compat.Test`) and removes old deprecations. Thus, anything working with version 2.x without deprecation warnings should also work with this, but possibly with deprecation warnings.

Note: When merging, this should obviously be squashed, but the commit message then should see some love and not just list the individual messages.


martinholters:
The README could use some editing, but I'd like to postpone that to a different PR. This PR just deletes obsolete entries, leaving empty sections and such. Also, the README still mentions `REQUIRE` etc.

martinholters:
Because this will be quite a conflict magnet but also simplify any future work once merged, it don't want to leave it open for too long. So if anyone wants to review (and I'd certainly welcome that) but needs more time, speak up! (Or if anyone has objections...) Otherwise, I'll probably merge in a few days.

PR #666, opened 2019-10-14
Readme subheadings

mcabbott:
It might be helpful to organise things by how new they are, perhaps like this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,21 +47,28 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+#### Julia 1.4
+
 * `only(x)` returns the one-and-only element of a collection `x`. ([#33129])
 
+#### Julia 1.3
+
 * `mod` now accepts a unit range as the second argument ([#32628]).
+
+#### Julia 1.2
+
+* `hasproperty` and `hasfield` ([#28850]).
+
+#### Julia 1.1
+
+* `range` supporting `stop` as positional argument ([#28708]).
+
+* `merge` methods with one and `n` `NamedTuple`s ([#29259]).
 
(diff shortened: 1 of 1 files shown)


martinholters:
I don't know. Not that I'm against this, but what's the benefit of separating by Julia version that introduced the feature?

mcabbott:
I guess the original motivation was not knowing which end of the list to add things to, before the big 1.0 cleanup. Far from essential now, just seemed better than random order. Perhaps a comment in the file to say “add new things at the top of this list” would be plenty.

mcabbott:
In fact #670 looks like a better way to go, order by Compat's version not Julia's.

martinholters:
Note that #670 only adds the required Compat version as additional information, doesn't order (or even group) by it, so in a way, it's orthogonal to what was proposed here.

It would be nice to know how people use the feature-list part of the README (or used it when it still had significant length). I usually came with the question "can I use XXX?" and would answer that by searching the README for XXX, not caring much about ordering or grouping of the entries. Do people do it the other way round, going through the README, saying, "oh cool, Compat supports XXX, I'll use that in my code"? Or some other way?

I guess what I'm after is: Is there an ordering/grouping that is helpful for users of Compat? Or can we just try to maximize maintainability?

mcabbott:
These are good questions, but I don’t have great answers. I know that a few people missed that Compat supported `eachslice` & friends, but don’t know that the readme organisation was to blame. But having some order seems better than none, if only for making it clear where is best to edit in additions.

martinholters:
OTOH, if all additions should happen at the top, say, that would increase the potential for conflicts among concurrent PRs. That's probably the case for the implementation and tests, too, which are usually appended at the bottom, so maybe not much of issue.

PR #671, opened 2019-10-21
Rewrite the deprecation of `parameter_upper_bound`

martinholters:
This replaces the call to the internal `Base.parameter_upper_bound` with its definition. As the function needs to be pure, it has to use `@depreate_binding`, with the actual replacement hidden in an internal function and a customized deprecation message.

Example output:
```julia
julia> Compat.TypeUtils.parameter_upper_bound(Array, 1)
WARNING: Compat.TypeUtils is deprecated, call the respective Base functions directly
 in module Compat
WARNING: TypeUtils.parameter_upper_bound is deprecated, rewrite your code to use static parameters in dispatch or use `Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)`.
 in module TypeUtils
Any
```

The small downside to using `@depreate_binding` is that the first part of the message says just `TypeUtils.parameter_upper_bound`, not `TypeUtils.parameter_upper_bound(t, idx)`, but that doesn't seem too bad.

Replaces #660.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -31,6 +31,10 @@ Base.@deprecate enable_debug(x::Bool) x false
 module TypeUtils
     Base.@deprecate_binding isabstract isabstracttype
     Base.@deprecate_binding typename Base.typename
-    Base.@deprecate_binding parameter_upper_bound Base.parameter_upper_bound
+    Base.@pure function _parameter_upper_bound(t::UnionAll, idx)
+        return Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)
+    end
+    Base.@deprecate_binding(parameter_upper_bound, _parameter_upper_bound, true,
+        ", rewrite your code to use static parameters in dispatch or use `Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)`.")
 end # module TypeUtils
 Base.@deprecate_binding TypeUtils TypeUtils false ", call the respective Base functions directly"


src/deprecated.jl
```diff
@@ -31,6 +31,10 @@ Base.@deprecate enable_debug(x::Bool) x false
 module TypeUtils
     Base.@deprecate_binding isabstract isabstracttype
     Base.@deprecate_binding typename Base.typename
-    Base.@deprecate_binding parameter_upper_bound Base.parameter_upper_bound
+    Base.@pure function _parameter_upper_bound(t::UnionAll, idx)
+        return Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)
+    end
+    Base.@deprecate_binding(parameter_upper_bound, _parameter_upper_bound, true,
+        ", rewrite your code to use static parameters in dispatch or use `Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)`.")
```
vtjnash, line 38: It has been a long time, but the preferred replacement for this is likely actually:
```julia
fieldtype(t, :related_field_name)
```
as demonstrated here: https://github.com/JuliaML/LossFunctions.jl/pull/153/files#diff-3ff7c6506d0aaafc6b13476d1c95b2afeec48f2fa3f2117c0327a3f08db66a70R22-R23

martinholters: With this deprecation in place for 3 and half years (and removed in Compat 4.x), I don't think this warning will be seen particularly often, so updating it seems moot. Or do you think it's worth fixing it? Then a PR against `release-3` would be the way to go.

PR #673, opened 2019-11-05
Update usage section of README for Julia 1.x

### LINKED ISSUE #667: Readme needs an update
https://github.com/JuliaLang/Compat.jl#usage 

Should be updated by removing the now obsolete `Compat.String` and `REQUIRE` files.

martinholters:
Fixes #667.

While at it, fix a broken link in the introduction.


martinholters:
I'm not too proud of my wording here, but it's definitely an improvement compared to the outdated stuff before, and no-one cared enough to suggest improvements, so...

PR #674, opened 2019-11-14
Add `pkgdir`

omus:
Added to Base in: https://github.com/JuliaLang/julia/pull/33128

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `pkgdir(m)` return the root directory of the package that imported module `m` ([#33128]). (since Compat 2.3.0)
+
 * `only(x)` returns the one-and-only element of a collection `x` ([#33129]). (since Compat 2.2.0)
 
 * `mod` now accepts a unit range as the second argument ([#32628]). (since Compat 2.2.0)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -47,6 +47,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `pkgdir(m)` return the root directory of the package that imported module `m` ([#33128]). (since Compat 2.3.0)
```
```suggestion
* `pkgdir(m)` returns the root directory of the package that imported module `m` ([#33128]).
```
(The next Compat version will be 3.0.0. And I propose to add the versions to the README when bumping the version in Project.toml, see https://github.com/JuliaLang/Compat.jl/issues/263#issuecomment-544988202. But if you prefer, you can also use 3.0.0 here now.)

fredrikekre: > The next Compat version will be 3.0.0

What changed?

martinholters: #665 which also removed old derprecations and introduced new ones.

martinholters: Well, by now it needs to be 3.1.0. I'll just commit the suggestion (not  setting the version number now) and we'll fix it when tagging the next version.

martinholters:
Sorry this has fallen through the cracks. Let's move it forward now. And merry Christmas.

PR #675, opened 2019-11-14
Remove at-static

musm:
I don't think `@static` is necessary here since this is defined in the main scope


musm:
I guess I opened this because every other module scope  `if VERSION ....` in the file doesn't prefix it with `@static` so it seemed out of place, so for consistency I opened this PR.

PR #676, opened 2019-11-14
Add Base internal function LIBEXECDIR

musm:
This is being backported to rc5, so I'm not sure if  `VERSION < v"1.4.0-DEV.445"` is appropriate?

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -95,13 +95,11 @@ end
 
 # https://github.com/JuliaLang/julia/pull/33777
 if v"1.4.0-DEV.172" <= VERSION < v"1.4.0-DEV.445"
-    libexecdir = Sys.iswindows() ? "..\\libexec" : "../libexec"
-    @eval(Base, const LIBEXECDIR = $libexecdir)
+    @eval(Base, const LIBEXECDIR = $(joinpath("..", "libexec")))
 elseif v"1.4" <= VERSION < v"1.4.0-DEV.172" 
     @eval(Base, const LIBEXECDIR = "")
 elseif v"1.3.0-rc3" <= VERSION <= v"1.3.0-rc5"     
-    libexecdir = Sys.iswindows() ? "..\\libexec" : "../libexec"
-    @eval(Base, const LIBEXECDIR = $libexecdir)
+    @eval(Base, const LIBEXECDIR = $(joinpath("..", "libexec")))
 elseif VERSION < v"1.3.0-rc3"
     @eval(Base, const LIBEXECDIR = "")
 end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -95,7 +95,7 @@ end
 
 # https://github.com/JuliaLang/julia/pull/33777
 if VERSION < v"1.4.0-DEV.445"
-    Base.LIBEXECDIR = ""
+    @eval Base LIBEXECDIR = ""
 end
 
 include("deprecated.jl")
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -93,6 +93,11 @@ if VERSION < v"1.4.0-DEV.329"
     Base.:∘(f, g, h...) = ∘(f ∘ g, h...)
 end
 
+# https://github.com/JuliaLang/julia/pull/33777
+if VERSION < v"1.4.0-DEV.445"
+    Base.LIBEXECDIR = ""
+end
+
 include("deprecated.jl")
 
 end # module Compat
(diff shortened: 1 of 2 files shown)


martinholters:
> This is being backported to rc5, so I'm not sure if `VERSION < v"1.4.0-DEV.445"` is appropriate?

Yeah, that's tricky. We obviously don't want to mess with `Base.LIBEXECDIR` in 1.3. So maybe `VERSION < v"1.3.0-rc5 || v"1.4" <=  VERSION < v"1.4.0-DEV.445`? Also, I guess we should wait with merging this until rc5 is out to be sure it really works there.

src/Compat.jl
```diff
@@ -93,6 +93,11 @@ if VERSION < v"1.4.0-DEV.329"
     Base.:∘(f, g, h...) = ∘(f ∘ g, h...)
 end
 
+# https://github.com/JuliaLang/julia/pull/33777
+if VERSION < v"1.4.0-DEV.445"
+    @eval Base LIBEXECDIR = ""
```
martinholters: Should this have a `const`? (And does the empty string make sense? I haven't followed the discussion...)

musm: Yeah it should have `const`. And yes the empty string makes sense because this is relative to `Sys.BINDIR`.    `LIBEXECDIR` is equal to  `"..\\libexecdir"`  on newer release but should be empty on releases where this constant is not defined, so that `joinpath(Sys.BINDIR, Base.LIBEXECDIR, "7z.exe")` works universally.

martinholters:
Is there a good way to test this? Make sure `7z.exe` can be found on windows?

musm:
Unfortunately it's really hard to test this since it's an optional build dependency whether 7z.exe exists in that folder,  and it also  actually not tested even on julia due to these difficulties.

musm:
Handling the rc situation is a bit tricky since there was a bug in earlier RC candidates where the 7z dll was not correctly copied over so I'm not sure if we need more advanced Versioning to handle all possible cases (this seems really overkill)

musm:
I think I fixed up all the version logic here which is a bit complex due to all the backports

https://github.com/JuliaLang/julia/pull/33125 
moved things into a new libexec folder  (backported to rc-3 https://github.com/JuliaLang/julia/pull/33221)

https://github.com/JuliaLang/julia/pull/33777 (backported to rc-5 https://github.com/JuliaLang/julia/pull/33630)


cc @staticfloat

musm:
The assumption is that all commits before JuliaLang/julia#33125  don't have a libexec folder so we emulate it's behavior where the relevant external exe's are located simply in the root julia `bin` folder.

musm:
Not sure why this is failing on 1.3, since it works locally

src/Compat.jl
```diff
@@ -93,6 +93,19 @@ if VERSION < v"1.4.0-DEV.329"
     Base.:∘(f, g, h...) = ∘(f ∘ g, h...)
 end
 
+# https://github.com/JuliaLang/julia/pull/33777
+if v"1.4.0-DEV.172" <= VERSION < v"1.4.0-DEV.445"
+    libexecdir = Sys.iswindows() ? "..\\libexec" : "../libexec"
```
```suggestion
    libexecdir = joinpath("..", "libexec)
```
(same below)

martinholters:
The test failure is due to resolving the `Sys` binding which makes this deprecation fail:

https://github.com/JuliaLang/Compat.jl/blob/0e8446145bfdb84c655537f3a5643f160adf0a20/src/deprecated.jl#L5

Probably explicitly using `Base.Sys` would work, or circumvent accessing `Sys` as I've suggested.

However, there is also this:
```
WARNING: eval into closed module Base:
Expr(:const, :LIBEXECDIR = "../libexec")
  ** incremental compilation may be fatally broken for this module **
```
You probably have to move the code into `__init()__`.

musm:
Great thanks, the problem was with the `Sys`  usage

musm:
I'm having second thoughts on whether we should include this. It's not apart of the stable API. Including it makes updating some packages easier, but I wonder if it's better to let them just handle it instead of including it as part of Compat.

src/Compat.jl
```diff
@@ -98,6 +98,17 @@ if VERSION < v"1.4.0-DEV.493"
     Base.Order.ReverseOrdering() = Base.Order.ReverseOrdering(Base.Order.Forward)
 end
 
+# https://github.com/JuliaLang/julia/pull/33777
+if v"1.4.0-DEV.172" <= VERSION < v"1.4.0-DEV.445"
+    @eval(Base, const LIBEXECDIR = $(joinpath("..", "libexec")))
+elseif v"1.4" <= VERSION < v"1.4.0-DEV.172"
```
musm: need also find the commit where this was moved, so this is not 100% correct in all cases.

PR #678, opened 2019-11-16
Reorder (and restructure) the sections in the README

martinholters:
Previously, the sections were
* Supported syntax
* Module Aliases
* New functions, macros, and methods
* Renaming
* New macros
* Other changes
* New types
* Developer tips

This changes the order to
* Supported syntax
* New functions and methods
* New macros
* New types
* Renaming
* Other changes
* Developer tips

getting rid of the ambiguity where to put new macros and removing the 
"Module Aliases" section (if needed, these could probably go into 
"Renaming").

Also, the entry from #668 is moved from "Other changes" to "New 
functions and methods" as it is a new method.

Finally, presently empty sections get a "None." placeholder entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,7 @@ the syntax on Julia `master`. However, in a few cases where this is not possible
 a slightly different syntax might be used.
 Please check the list below for the specific syntax you need.
 
-## Supported syntax
-
-Currently, the `@compat` macro supports the following syntaxes:
-
-## Module Aliases
-
-## New functions, macros, and methods
+## Supported features
 
 * `only(x)` returns the one-and-only element of a collection `x` ([#33129]). (since Compat 2.2.0)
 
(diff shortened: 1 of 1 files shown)


martinholters:
As you also have some ideas for the README (#680), mind taking a look here and commenting, @oxinabox?

oxinabox:
I suggest `Supported syntax`  go at the bottom, as the `@compat` macro is rarely used, 
and having it at the top makes it look like all the other things require it

martinholters:
Yes, let's trim down the number of sections. I wasn't sure about keeping the empty ones anyway. If things start to become a mess, we can still re-introduce new sections as needed then.

Looking back at when we still had `@compat` entries in the README, they were always explicitly mentioning the macro, e.g. "`@compat foo(; x, y)` makes `x` and `y` required keyword arguments". So I'm not even convinced they need their own section. (Anyway, until development of Julia 2 starts, it seems unlikely that we will need to use `@compat` much.)

So maybe we can just collect all supported features in one section, tentatively title "Supported features"? That would leave
* Usage
* Supported features
* Devloper tips

Nice and clean :smile:
What do people think?

oxinabox:
Sounds good to me

martinholters:
Incorporated the sorting from #680 (newest on top) here and will merge soonish unless anyone objects.

PR #679, opened 2019-11-17
add zero arg ReverseOrdering constructor

oxinabox:
From https://github.com/JuliaLang/julia/pull/33736/

Needed for https://github.com/JuliaCollections/DataStructures.jl/pull/547

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "2.2.0"
+version = "2.3.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "2.3.0"
+version = "3.0.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `merge` methods with one and `n` `NamedTuple`s ([#29259]). (since Compat 2.0.0)
 
-* `Base.Order.ReverseOrdering` has a zero arg constructor [#33736]. (since Compat 2.3.0)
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "2.2.0"
+version = "2.3.0"
```
martinholters: Has to be 3.0.0

martinholters:
On second thought, also needs a README entry.

Project.toml
oxinabox: Why does this have to be 3.0.0?
This isn't a breaking change ?

martinholters: No, this one isn't, but #665 was, and we haven't tried since then.

README.md
```diff
@@ -65,6 +65,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 * `merge` methods with one and `n` `NamedTuple`s ([#29259]). (since Compat 2.0.0)
 
+* `Base.Order.ReverseOrdering` has a zero arg constructor [#33736]. (since Compat 2.3.0)
```
```suggestion
* `Base.Order.ReverseOrdering` has a zero arg constructor [#33736]. (since Compat 3.0.0)
```

Project.toml
martinholters: Also, if this bumps the package version, it should add `(since Compat 3.0.0)` to all new entries in the README, i.e. the one multi-arg function composition.

oxinabox: I'm not sure which are which.
Is it just anything that doesn't list a version?

martinholters: Yes.
(Might be easier to not do the version bump and instead create a PR for it on its own, but I leave that you. In any case, I'm all in favor of releasing 3.0.0 soon.)

oxinabox: I am a big fan of tagging a release after every non-breaking PR.
Also of making sure the version number in the Project.toml is correct after every PR,
and set to `vX.Y.Z-DEV` if not releasing.

See my blog post
https://white.ucc.asn.au/2019/09/28/Continuous-Delivery-For-Julia-Packages.html#what-if-i-dont-want-to-release-right-now--dev-versions

oxinabox:
can we tag a release please

martinholters:
Long done: https://github.com/JuliaRegistries/General/pull/5534 and https://github.com/JuliaLang/Compat.jl/releases/tag/v3.0.0

PR #680, opened 2019-11-17
sort by version in readme

oxinabox:
AFAICT entries readme was ordered randomly?

Feel free to close if there is some pattern that I am missing

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,19 +50,19 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
-* `range` supporting `stop` as positional argument ([#28708]). (since Compat 1.3.0)
-
-* `hasproperty` and `hasfield` ([#28850]).  (since Compat 2.0.0)
-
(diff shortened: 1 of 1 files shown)


martinholters:
Xref discussion in #666. You're right that the entries have no particular order. And not that I'm against sorting them, but is there any tangible benefit? Or is it just aesthetic?

oxinabox:
It is a lot easier to navigate:
Find the last thing from the list of 5 things I need, scrolling up from the bottom.
I know that since it is sorted that means the other 4 must all also be covered by setting to this level

martinholters:
Yes, that makes perfect sense! Although maybe reverse the order with newest on top? Scrolling down seems more intuitive (to me, at least).

fredrikekre:
> Although maybe reverse the order with newest on top?

Was about to suggest this.

martinholters:
Can you rebase to include the `Base.Order.ReverseOrdering` entry in the sorting? (Or we merge #678 first?)

oxinabox:
I will rebase

martinholters:
Superseded by #678.

PR #682, opened 2019-12-03
Filter(f, tuple)

mcabbott:
This adds https://github.com/JuliaLang/julia/pull/32968 .

I bumped the version and put that number in the readme, but can remove if that should be done separately.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.0.1"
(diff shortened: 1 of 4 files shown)


martinholters:
> I bumped the version and put that number in the readme, but can remove if that should be done separately.

That's ok, but should be 3.1.0. (And of course, runs risk of needing to be updated if something else get merged and released first.) Otherwise LGTM.

martinholters:
I've manually merged master with the updated README from #678 to put the new entry on top. Should be good to go now.

PR #683, opened 2019-12-04
Add dot(x,A,y) 

mcabbott:
This adds most of https://github.com/JuliaLang/julia/pull/32739. 

I got tired of copying and skipped some of the more exotic matrix types. They should all hit the generic fallback though, so should still work. 

I commented out bits of `src/deprecated.jl` to make tests pass. I’m not too sure what those are meant to do, and whether they are still needed post-1.0?

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
-* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.1.0)
+* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.2.0)
 
 * `only(x)` returns the one-and-only element of a collection `x` ([#33129]). (since Compat 2.2.0)
 
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.0.1)
+
 * `only(x)` returns the one-and-only element of a collection `x` ([#33129]). (since Compat 2.2.0)
 
 * `mod` now accepts a unit range as the second argument ([#32628]). (since Compat 2.2.0)
@@ -114,6 +116,7 @@ Note that you should specify the correct minimum version for `Compat` in the
 [#29674]: https://github.com/JuliaLang/julia/issues/29674
 [#29749]: https://github.com/JuliaLang/julia/issues/29749
 [#32628]: https://github.com/JuliaLang/julia/issues/32628
+[#32739]: https://github.com/JuliaLang/julia/pull/32739
 [#33129]: https://github.com/JuliaLang/julia/issues/33129
 [#33568]: https://github.com/JuliaLang/julia/pull/33568
 [#33736]: http://github.com/JuliaLang/julia/pull/33736
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
-* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.0.1)
+* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.1.0)
 
 * `only(x)` returns the one-and-only element of a collection `x` ([#33129]). (since Compat 2.2.0)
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -50,6 +50,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.0.1)
```
fredrikekre:
```suggestion
* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.1.0)
```

martinholters:
Removing the deprecations makes this a breaking change, which I'd like to avoid. If you use `dot` qualified everywhere (i.e. `LinearAlgebra.dot`), that should go a long way. What other problems appear with the deprecations in place?

martinholters:
Ok, I've re-enabled the deprecations after some tweaking.

martinholters:
Can we get some nicer code formatting here? E.g. no lines longer than 92 chars and `import` and `using` statements collected at the top. (Yes, I missed the opportunity to do so in the commit I've pushed.)

README.md
```diff
@@ -50,6 +50,8 @@ Currently, the `@compat` macro supports the following syntaxes:
 
 ## New functions, macros, and methods
 
+* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.1.0)
```
```suggestion
* `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.2.0)
```

mcabbott:
See if you like the current version, not strictly 92 but narrower than earlier things. And thanks for fixing the deprecations earlier!

src/Compat.jl
```diff
@@ -88,6 +88,89 @@ if VERSION < v"1.3.0-alpha.8"
...
+    LinearAlgebra.dot(x::AbstractVector, a::Union{Real,Complex}, y::AbstractVector) =
+        a*LinearAlgebra.dot(x, y)
+end
+
+# https://github.com/JuliaLang/julia/pull/30630
+if VERSION < v"1.2.0-DEV.125" # 1da48c2e4028c1514ed45688be727efbef1db884
+    require_one_based_indexing(A...) = !Base.has_offset_axes(A...) || throw(ArgumentError(
+        "offset arrays are not supported but got an array with index other than 1"))
+else
+    using Base: require_one_based_indexing
```
martinholters: I'm not sure we should be using this. It's an internal function that might just go away eventually. Unless I missed something, it's used only once above, so we could just unconditionally replace it with its definition there. Or at least, we should move this block into the other conditional, so that `Base.require_one_based_indexing` is only resolved for Julia versions where we know it exists.

src/Compat.jl
```diff
@@ -88,6 +88,89 @@ if VERSION < v"1.3.0-alpha.8"
     Base.mod(i::Integer, r::AbstractUnitRange{<:Integer}) = mod(i-first(r), length(r)) + first(r)
 end
 
+# https://github.com/JuliaLang/julia/pull/32739
+# This omits special methods for more exotic matrix types, Triangular and worse.
+if VERSION < v"1.4.0-DEV.92" # 2425ae760fb5151c5c7dd0554e87c5fc9e24de73
+
+    import LinearAlgebra
+    using LinearAlgebra: Adjoint, Diagonal, Transpose, UniformScaling, RealHermSymComplexHerm
```
martinholters: I think I'd prefer these imports to happen unconditionally and be placed at the top of the file. Or is there any good reason to have them here instead?

test/runtests.jl
```diff
@@ -90,6 +90,87 @@ end
...
+                    @test dot(x, Symmetric(aherm, uplo), y) ≈ dot(x, Symmetric(aherm, uplo)*y) ≈ dot(x, Matrix(Symmetric(aherm, uplo)), y)
+                    @test dot(x, Symmetric(aherm, uplo), x) ≈ dot(x, Symmetric(aherm, uplo)*x) ≈ dot(x, Matrix(Symmetric(aherm, uplo)), x)
+                end
+            end
+        end
+    end
+
+    # stdlib/LinearAlgebra/test/uniformscaling.jl
+
+    # Quaternion tests don't work on Julia 1.1
```
martinholters: I think we should either just copy `testhelpers/Quaternions.jl` over from Julia and enable these tests using it or drop those comments.

mcabbott:
The generic case is tested, in that commenting it out leads to `MethodError: no method matching dot(::Int64, ::Int64, ::Int64)`. I can't see a test for `Diagonal` in https://github.com/JuliaLang/julia/pull/32739/files , so I've invented one.

test/runtests.jl
mcabbott: OK I've deleted these lines.

src/Compat.jl
mcabbott: Just trying to keep things from one PR in one place I guess. But can move them upwards.

src/Compat.jl
mcabbott: Indeed it's only used once now, although I think I have run into it elsewhere. The reason to add it was just to keep the `dot` methods as close to identically copied as possible, rather than inlining the definition.

martinholters:
I've added a merge of master and moved the README entry to the top of the list, and bumped the version to 3.2.0. Once CI is green, I'll merge and tag a release.

PR #685, opened 2020-02-01
Add uuid5

tkf:
This PR adds `uuid5` from https://github.com/JuliaLang/julia/pull/28761

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Please check the list below for the specific syntax you need.
 
 ## Supported features
 
+* `uuid5` is added ([#28761]). (since Compat 3.3.0)
+
 * `dot` now has a 3-argument method `dot(x, A, y)` without storing the intermediate result `A*y` ([#32739]). (since Compat 3.2.0)
 
 * `pkgdir(m)` returns the root directory of the package that imported module `m` ([#33128]). (since Compat 3.2.0)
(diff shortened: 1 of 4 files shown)


martinholters:
This makes `Compat,namespace_dns`, `Compat,namespace_url`, `Compat,namespace_oid`, and `Compat,namespace_x500` available but doesn't mention them in the README. Providing functionality but not telling anyone seems inconsistent to me. What's the reason for this?

tkf:
Thanks for catching it. It was an overlook.

martinholters:
Hmm,  https://github.com/JuliaLang/julia/pull/28761 neither exports nor documents those constants. Are they actually part of the API and guaranteed to stay for 1.x? Otherwise I'd prefer not to have a Compat for them. How much are these needed anyway?

tkf:
I removed `namespace_(dns|oid|url|x500)` from `src/Compat.jl`. I suppose whoever wants them has to first document them in the UUIDs stdlib as stable API.

martinholters:
Thanks. Better to be conservative here than to provide functionality that could happen to be removed from Julia...

README.md
```diff
@@ -44,6 +44,8 @@ Please check the list below for the specific syntax you need.
 
 ## Supported features
 
+* `uuid5` is added ([#28761]). (since Compat 3.3.0)
```
martinholters: Usually these give a brief summary of what the function/method/... does.
```suggestion
* `uuid5` generates a version 5 universally unique identifier (UUID), as specified by RFC 4122 ([#28761]). (since Compat 3.3.0)
```

martinholters:
Modulo a slightly more descriptive entry in the README this looks good to me. If want, you can include a version bump in `Project.toml` and we make a release immediately after merging.

tkf:
Thanks for the review. I updated the version number.

martinholters:
Thanks!

PR #689, opened 2020-02-09
Support fancy CartesianIndex operations on Julia 1.0

timholy:
This supports the code in https://julialang.org/blog/2016/02/iteration/

This has started becoming an issue, e.g., https://github.com/JuliaImages/ImageMorphology.jl/pull/24

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,11 @@ Please check the list below for the specific syntax you need.
 
 ## Supported features
 
+* `I1:I2`, when `I1` and `I2` are CartesianIndex values, constructs a CartesianIndices
+  iterator ([#29440]). (Since Compat 3.5.0)
+
+* `oneunit(::CartesianIndex)` replaces `one(::CartesianIndex)` ([#29442]). (Since Compat 3.5.0)
+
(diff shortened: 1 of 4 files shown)


martinholters:
Can you add an entry to the README?

timholy:
Done. I was guessing this would be released as part of 3.5.0, but happy to change it if that's not correct.

README.md
```diff
@@ -44,6 +44,11 @@ Please check the list below for the specific syntax you need.
 
 ## Supported features
 
+* `I1:I2`, when `I1` and `I2` are CartesianIndex values, constructs a CartesianIndices
+  iterator ([#29440]). (Since Compat 3.5.0)
+
+* `oneunit(::CartesianIndex)` is supported ([#29442]). (Since Compat 3.5.0)
```
```suggestion
* `oneunit(::CartesianIndex)` replaces `one(::CartesianIndex)` ([#29442]). (Since Compat 3.5.0)
```
or something to that effect?

martinholters:
> I was guessing this would be released as part of 3.5.0, but happy to change it if that's not correct.

Sounds plausible. Feel free to also bump the version in `Project.toml` and we'll release immediately when merged.

PR #693, opened 2020-04-06
similar(::PermutedDimsArray)

mcabbott:
Adds https://github.com/JuliaLang/julia/pull/35304, originally for  https://github.com/JuliaGPU/CuArrays.jl/issues/658

PR #694, opened 2020-04-22
Add mergewith[!](combine, dicts...)

tkf:
Quoting README:

> * `mergewith(combine, dicts...)` and `mergewith!(combine, dicts...)` are like `merge(combine, dicts...)` and `merge!(combine, dicts...)` but without the restriction that the argument `combine` must be a `Function` ([#34296]). (since Compat 3.9.0).

Ref: https://github.com/JuliaLang/julia/pull/34296

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.8.0"
+version = "3.9.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.8.0"
+version = "3.9.0"
```
martinholters, line 3:
```suggestion
version = "3.9.0-DEV"
```
I'll try to get #693, #695 and this one in soonish, and then do a new version, instead of bumping the version three times in a row, especially as that would mean having to update the not-yet-merged PRs after every version bump.

tkf:
@martinholters Thanks for the review! I updated the patch.

PR #697, opened 2020-05-02
Support hadamard and tensor

timholy:


martinholters:
Should this be closed given https://github.com/JuliaLang/julia/pull/35744? (I didn't follow all that discussion along...)

timholy:
Yes, thanks for noticing!

PR #704, opened 2020-06-01
Add `union(::OneTo, ::OneTo)`

goretkin:
I didn't bump the version number because I see there's another recent merge request.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -396,6 +396,11 @@ if VERSION < v"1.5.0-DEV.124"
     export isdisjoint
 end
 
+# https://github.com/JuliaLang/julia/pull/35577
+if VERSION < v"1.5.0-DEV.675"
+    Base.union(r::Base.OneTo, s::Base.OneTo) = Base.OneTo(max(r.stop,s.stop))
+end
+
 include("deprecated.jl")
 
 end # module Compat
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -397,7 +397,7 @@ if VERSION < v"1.5.0-DEV.124"
 end
 
 # https://github.com/JuliaLang/julia/pull/35577
-if VERSION < v"1.5.0-DEV.675"
+if VERSION < v"1.5.0-DEV.681"
     Base.union(r::Base.OneTo, s::Base.OneTo) = Base.OneTo(max(r.stop,s.stop))
 end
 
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -396,6 +396,11 @@ if VERSION < v"1.5.0-DEV.124"
     export isdisjoint
 end
 
+# https://github.com/JuliaLang/julia/pull/35577
+if VERSION < v"1.5.0-DEV.675"
```
martinholters, line 400:
```suggestion
if VERSION < v"1.5.0-DEV.681"
```
(At lest that's what I get from `commit-name.sh 641f444`. How did you determine that version bound?)

test/runtests.jl
```diff
@@ -400,4 +400,9 @@ end
     end
 end
 
+# https://github.com/JuliaLang/julia/pull/35577
+@testset "union on OneTo" begin
+    @test union(Base.OneTo(3), Base.OneTo(4)) == Base.OneTo(4)
```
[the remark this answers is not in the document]
goretkin: Great point! That means the test for Base is also broken in this way :-X

src/Compat.jl
goretkin: oh I see,

641f444, which is the commit where the test went in
75c68d, (which I used) is the commit where the change went in

martinholters: But https://github.com/JuliaLang/julia/commit/641f444a0beec17ae93ca4af5a2992d1c31cef43 adds definition and test, no? And I can't find a 75c68d. Otherwise I'd agree with you.

goretkin: Not sure where I got confused (maybe I squashed, or force-pushed), but you're absolutely right! Sorry about that.

goretkin:
I think this is good to go now! Thanks for the careful review, @martinholters

martinholters:
Hm, I'm having second thoughts as this changes the return type of `union(::Base.OneTo, ::Base.OneTo)` when `Compat` is loaded (on older Julia). It might be safer to instead define `Compat.union` to follow what Julia does in newer versions and leave `Base.union` alone.

In this case, I guess it's unlikely to cause any trouble (and https://github.com/JuliaLang/julia/pull/35577 doesn't seem to have) so I'm in favor of merging as is, just wanted to point out this potential for problems and give others a bit more time to chime in.

martinholters:
Well, no objections voiced, so here we go...

PR #705, opened 2020-06-09
Searching a character in a string

emmt:
Extend search methods for looking for a character in a string.  These appear in Julia 1.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Please check the list below for the specific syntax you need.
 
 ## Supported features
 
+* Search a character in a string with `findfirst`, `findnext`, `findlast`, and `findprev`.
+
 * `∘(f) = f` is defined ([#34251]). (since Compat 3.11.0)
 
 * `union` supports `Base.OneTo` ([#35577]). (since Compat 3.11.0)
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Please check the list below for the specific syntax you need.
 
 ## Supported features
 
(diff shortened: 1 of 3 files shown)


src/Compat.jl
```diff
@@ -65,6 +65,15 @@ if VERSION < v"1.2.0-DEV.272"
     export hasproperty
 end
 
+if VERSION < v"1.3.0-alpha"
```
martinholters, line 68:
```suggestion
if VERSION < v"1.3.0-DEV.349"
```
Not that it matters much, but we generally try to use the exact version.

emmt: Sure. Thanks for the suggestion !
BTW how do you figure out this version number.

emmt: I have just added tests as you suggested.

README.md
```diff
@@ -44,6 +44,8 @@ Please check the list below for the specific syntax you need.
 
 ## Supported features
 
+* Search a character in a string with `findfirst`, `findnext`, `findlast`, and `findprev`.
```
martinholters, line 47:
```suggestion
* Search a character in a string with `findfirst`, `findnext`, `findlast`, and `findprev` ([#31664]). (since Compat 3.12.0)
```
Also needs the link to be added at the bottom. (And assumes this will make it into the next release, which I assume it will.)

emmt: Ok thanks!

martinholters: The corresponding link still needs to be added in the list at the bottom of the file.

martinholters:
Thanks!

PR #709, opened 2020-07-01
Add BLAS.get_num_threads

mcabbott:
Ref https://github.com/JuliaLang/julia/pull/36360

Note that I didn't export this, leaving it as `Compat.get_num_threads()`, but did import the version from LinearAlgebra when defined. 

Also, I copied the new version of `set_num_threads()` here, rather than calling the existing one.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -557,8 +557,7 @@ if VERSION < v"1.6.0-DEV.322" # b8110f8d1ec6349bee77efb5022621fdf50bd4a5
 
 else
     # Ensure that these can still be accessed as Compat.get_num_threads() etc:
-    using LinearAlgebra
-    import BLAS: set_num_threads, get_num_threads
+    import LinearAlgebra.BLAS: set_num_threads, get_num_threads
 end
 
 include("deprecated.jl")
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1,7 +1,7 @@
 module Compat
 
 import LinearAlgebra
-using LinearAlgebra: Adjoint, Diagonal, Transpose, UniformScaling, RealHermSymComplexHerm
+using LinearAlgebra: Adjoint, Diagonal, Transpose, UniformScaling, RealHermSymComplexHerm, BLAS
 
 include("compatmacro.jl")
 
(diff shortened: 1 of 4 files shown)


martinholters:
Should `set_num_threads()` also be mentioned in the README and `set_num_threads(nothing)` be tested (that it doesn't error, at least)? Or why is it included here?

mcabbott:
I tried to minimally adapt the PR's tests, but I agree that it's odd that `set_num_threads(::Nothing)` isn't tested, I can add something. The difference from the existing function is that it accepts `nothing`, and that it prints more warnings. Seemed simpler to copy the whole new implementation than to split it.

test/runtests.jl
```diff
@@ -464,4 +464,39 @@ end
...
+    @test Compat.get_num_threads() === 1
+    Compat.set_num_threads(default)
+    @test Compat.get_num_threads() === default
+
+    # Run the ::Nothing method, to check no error:
+    Compat.set_num_threads(nothing)
+    Compat.set_num_threads(default)
+
+    if VERSION < v"1.6.0-DEV.322"
+        # These tests from PR rely on internal functions which would be BLAS. not Compat.
```
martinholters, line 482: This is a bit ugly but I can't think of a better way either.

src/Compat.jl
```diff
@@ -464,6 +464,102 @@ if VERSION < v"1.2.0-DEV.77"
...
 end
 
+# https://github.com/JuliaLang/julia/pull/36360
+if VERSION < v"1.6.0-DEV.322" # b8110f8d1ec6349bee77efb5022621fdf50bd4a5
+
+    function guess_vendor()
+        # like determine_vendor, but guesses blas in some cases
+        # where determine_vendor returns :unknown
+        ret = BLAS.vendor()
+        if Sys.isapple() && (ret == :unknown)
```
martinholters, line 474:
```suggestion
        if Base.Sys.isapple() && (ret == :unknown)
```

src/Compat.jl
```diff
@@ -464,6 +464,102 @@ if VERSION < v"1.2.0-DEV.77"
...
+        end
+        return nothing
+    end
+    _tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
+
+    function _set_num_threads(::Nothing; _blas = guess_vendor())
+        n = something(
+            _tryparse_env_int("OPENBLAS_NUM_THREADS"),
+            _tryparse_env_int("OMP_NUM_THREADS"),
+            max(1, Sys.CPU_THREADS ÷ 2),
```
martinholters, line 521:
```suggestion
            max(1, Base.Sys.CPU_THREADS ÷ 2),
```

test/runtests.jl
mcabbott: Could also dump these & call testing the overall behaviour enough, here.

martinholters: But IIUC, some of the code below explicitly invokes code paths not hit by usual CI, so might be worth it. Not sure.

PR #710, opened 2020-07-01
Strides for Adjoint & Transpose

mcabbott:
This adds https://github.com/JuliaLang/julia/pull/35929 which updated https://github.com/JuliaLang/julia/pull/29135 .

Edit: the more recently merged PR https://github.com/JuliaLang/julia/pull/35929 is marked "backport 1.5". I presume that this means the VERSION cutoff used here may need to change.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -414,6 +414,30 @@ if VERSION < v"1.5.0-DEV.681"
     Base.union(r::Base.OneTo, s::Base.OneTo) = Base.OneTo(max(r.stop,s.stop))
 end
 
+# https://github.com/JuliaLang/julia/pull/35929
+# and also https://github.com/JuliaLang/julia/pull/29135 -> Julia 1.5
+if VERSION < v"1.6.0-DEV.323"
+
+    # This is Int not Integer, to be more strict than an existing method:
+    function stride(A::AbstractArray, k::Int)
+        st = strides(A)
+        k ≤ ndims(A) && return st[k]
+        return sum(st .* size(A))
+    end
+
+    # These were first defined for Adjoint{...,StridedVector} etc in #29135
+    Base.strides(A::Adjoint{<:Real, <:AbstractVector}) = (stride(A.parent, 2), stride(A.parent, 1))
+    Base.strides(A::Transpose{<:Any, <:AbstractVector}) = (stride(A.parent, 2), stride(A.parent, 1))
+    Base.strides(A::Adjoint{<:Real, <:AbstractMatrix}) = reverse(strides(A.parent))
+    Base.strides(A::Transpose{<:Any, <:AbstractMatrix}) = reverse(strides(A.parent))
+    Base.unsafe_convert(::Type{Ptr{T}}, A::Adjoint{<:Real, <:AbstractVecOrMat}) where {T} = Base.unsafe_convert(Ptr{T}, A.parent)
+    Base.unsafe_convert(::Type{Ptr{T}}, A::Transpose{<:Any, <:AbstractVecOrMat}) where {T} = Base.unsafe_convert(Ptr{T}, A.parent)
+
+    Base.elsize(::Type{<:Adjoint{<:Real, P}}) where {P<:AbstractVecOrMat} = Base.elsize(P)
+    Base.elsize(::Type{<:Transpose{<:Any, P}}) where {P<:AbstractVecOrMat} = Base.elsize(P)
+
+end
+
 # https://github.com/JuliaLang/julia/pull/27516
 if VERSION < v"1.2.0-DEV.77"
     import Test: @inferred
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -416,7 +416,7 @@ end
 
 # https://github.com/JuliaLang/julia/pull/35929
 # and also https://github.com/JuliaLang/julia/pull/29135 -> Julia 1.5
-if VERSION < v"1.6.0-DEV.323"
+if VERSION < v"1.5.0-rc1.13" || v"1.6.0-" < VERSION < v"1.6.0-DEV.323"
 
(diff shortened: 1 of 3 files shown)


martinholters:
I'll review later, but on first glance: Needs a README entry.

mcabbott:
Sure. If I do it after #709 then I won't have to fix clashes. 

This one may need to wait for the backport anyway.

src/Compat.jl
```diff
@@ -414,6 +414,30 @@ if VERSION < v"1.5.0-DEV.681"
     Base.union(r::Base.OneTo, s::Base.OneTo) = Base.OneTo(max(r.stop,s.stop))
 end
 
+# https://github.com/JuliaLang/julia/pull/35929
+# and also https://github.com/JuliaLang/julia/pull/29135 -> Julia 1.5
+if VERSION < v"1.6.0-DEV.323"
+
+    # This is Int not Integer, to be more strict than an existing method:
+    function stride(A::AbstractArray, k::Int)
```
martinholters, line 422: There is no other method for `Compat.stride`. Should this method be added to `Base.stride`? And this a genuine question, not necessarily a suggestion. Maybe just fixing the comment is enough if the extension of `strides` is the only relevant aspect here.

mcabbott: Good catch, I meant to add a method to `Base.stride`, but make it more specific than the existing one (which was replaced in the PR). On Julia 1.4:
```
julia> methods(stride)
# 3 methods for generic function "stride":
[1] stride(V::SubArray, d::Integer) in Base at subarray.jl:318
[2] stride(a::Union{Base.ReinterpretArray{T,N,S,A}  ... this is StridedArray ...  DenseArray}, i::Int64) in Base at reinterpretarray.jl:64
[3] stride(A::AbstractArray, k::Integer) in Base at abstractarray.jl:395
```
But in fact this will introduce ambiguities:
```
julia> stride(view(rand(2,2),:,1:2), 1)
ERROR: MethodError: stride(::SubArray{Float64,2,Array{Float64,2},Tuple{Base.Slice{Base.OneTo{Int64}},UnitRange{Int64}},true}, ::Int64) is ambiguous. Candidates:
  stride(A::AbstractArray, k::Int64) in Main at REPL[3]:1
  stride(V::SubArray, d::Integer) in Base at subarray.jl:311
Possible fix, define
  stride(::SubArray, ::Int64)
```
But if I overwrite the base method with one of the same signature, I get
```
WARNING: Method definition stride(AbstractArray{T, N} where N where T, Integer) in module Base at abstractarray.jl:395 overwritten in module Compat at /Users/me/.julia/dev/Compat/src/Compat.jl:423.
```
So perhaps it should in fact be `Compat.stride`, for use here, with a fallback to `Base.stride`.

martinholters:
LGTM. Once the backport is done and the version conditional updated accordingly, this should be good to go.

src/Compat.jl
```diff
@@ -414,6 +414,31 @@ if VERSION < v"1.5.0-DEV.681"
     Base.union(r::Base.OneTo, s::Base.OneTo) = Base.OneTo(max(r.stop,s.stop))
 end
 
+# https://github.com/JuliaLang/julia/pull/35929
+# and also https://github.com/JuliaLang/julia/pull/29135 -> Julia 1.5
+if VERSION < v"1.6.0-DEV.323"
```
martinholters, line 419:
```suggestion
if VERSION < v"1.5.0-rc1.13" || v"1.6.0-" < VERSION < v"1.6.0-DEV.323"
```

martinholters:
Travis Linux Julia 1.0:
```
generalized dot #32739: Test Failed at /home/travis/build/JuliaLang/Compat.jl/test/runtests.jl:173

  Expression: dot(x, transpose(A), y) ≈ dot(x, transpose(A) * y) ≈ x' * transpose(A) * y ≈ (x' * transpose(A)) * y

   Evaluated: 0.0016273856f0 ≈ 0.0016267318f0 ≈ 0.0016271947f0 ≈ 0.0016271947f0
```
AV Julia latest (1.6.0-DEV.452) 32 bit:
```
generalized dot #32739: Test Failed at C:\projects\compat-jl\test\runtests.jl:171
  Expression: dot(x, A, y) ≈ dot(A' * x, y) ≈ x' * A * y ≈ (x' * A) * y
   Evaluated: 0.0003067404f0 ≈ 0.00030654308f0 ≈ 0.00030654308f0 ≈ 0.00030654308f0
```
All others pass. Does this influence the dispatch of `dot` in some way here? E.g., going through BLAS more often than before or something?

The inputs here are randomly generated, that might explain why it only fails sometimes, but I haven't seen this before, so most likely related to this PR.

martinholters:
On second thought, this PR shouldn't change anything on Julia 1.6.0-DEV.452 (except for adding tests, which are not the ones failing here), so hard to blame it. But the same change in Base could lead to same problem in the here, of course.

mcabbott:
I don't understand this at all. It's both 1.0 on Travis and master on Appveyor, on the same test. Both seem like they ought to have been identical on earlier runs of this PR. 

And I can't see how `strides` would change `dot`, `rand(2,2)' isa StridedArray` remains false so it shouldn't dispatch differently.

martinholters:
Running
```julia
@testset begin
    n = 10
    for iter in 1:1000000
        A = randn(Float32, n, n)
        x = rand(Float32, n)
        y = rand(Float32, n)
        @test dot(x, transpose(A), y) ≈ dot(x, transpose(A)*y) ≈ *(x', transpose(A), y) ≈ (x'*transpose(A))*y
    end
end
```
gives around 200 fails across different Julia versions. Seems we were just very unlucky with the latest CI run. I'll restart and if it looks better this time, this should be ready to merge. The flaky test is an orthogonal issue.

PR #711, opened 2020-07-03
Instruct using package manager

goretkin:
Perhaps there's a good reason to instruct manually editing the `[deps]` section of `Project.toml`, but I couldn't think of one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ it is important to maintain cross-version compatibility.
 
 ## Usage
 
-To use Compat in your Julia package, add the package
+To use Compat in your Julia package, add it as a dependency of your package using the package manager
 
 ```julia
 pkg> add Compat
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ To use Compat in your Julia package, add it as a dependency of your package usin
 ```julia
 pkg> add Compat
 ```
-and add a line `Compat = "..."` in the `[compat]`section of the `Project.toml` file
+and add a line such as `Compat = "2.2, 3"` in the `[compat]`section of the `Project.toml` file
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,9 +18,12 @@ it is important to maintain cross-version compatibility.
 
 ## Usage
 
-To use Compat in your Julia package, add a line
-`Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"` in the `[deps]` section
-and a line `Compat = "..."` in the `[compat]`section to the `Project.toml` file
+To use Compat in your Julia package, add the package
```
fredrikekre, line 21:
```suggestion
To use Compat in your Julia package, add it as a dependency of your package using the package manager
```

README.md
```diff
@@ -18,9 +18,12 @@ it is important to maintain cross-version compatibility.
...
 
-To use Compat in your Julia package, add a line
-`Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"` in the `[deps]` section
-and a line `Compat = "..."` in the `[compat]`section to the `Project.toml` file
+To use Compat in your Julia package, add it as a dependency of your package using the package manager
+
+```julia
+pkg> add Compat
+```
+and add a line `Compat = "..."` in the `[compat]`section of the `Project.toml` file
```
mcabbott, line 26: Perhaps it's clearest to give an example, to make notation obvious:
```suggestion
and add a line such as `Compat = "2.2, 3"` in the `[compat]`section of the `Project.toml` file
```

goretkin: I think that's a great suggestion. I had to look up the notation on another package to be sure I was doing it right.

PR #713, opened 2020-07-16
Add random seed before dot(x,A,y) tests

### LINKED ISSUE #712: Generalized dot test sensitve to numerical precision
The tests at
https://github.com/JuliaLang/Compat.jl/blob/e667af7e6e31c090776249f9a1b5df2684a571d2/test/runtests.jl#L171-L173
sometimes fail for `Float32` eltype with an error that can probably be attributed to numerical precision issues, e.g.
```
  Expression: dot(x, transpose(A), y) ≈ dot(x, transpose(A) * y) ≈ x' * transpose(A) * y ≈ (x' * transpose(A)) * y
   Evaluated: -0.0029978678f0 ≈ -0.0029971108f0 ≈ -0.0029982424f0 ≈ -0.0029982424f0
```
The same tests appear in the LinearAlgebra tests:
https://github.com/JuliaLang/julia/blob/29826c2c0893488dce0d4ee92b457f7f1a586dfb/stdlib/LinearAlgebra/test/generic.jl#L472-L474
So Julia is probably effected as well.

The probability of failing seems to be quite low (around 0.02% in a brief experiment), but it would still be nice to sort this out. CC @dkarrasch who added these tests in julia and @mcabbott who brought them here.

mcabbott:
Closes #712, I hope.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -149,9 +149,11 @@ end
     end
 end
 
-using LinearAlgebra
+using LinearAlgebra, Random
 
 @testset "generalized dot #32739" begin
+    Random.seed!(42) # https://github.com/JuliaLang/Compat.jl/issues/712
+
     # stdlib/LinearAlgebra/test/generic.jl
     for elty in (Int, Float32, Float64, BigFloat, Complex{Float32}, Complex{Float64}, Complex{BigFloat})
         n = 10


test/runtests.jl
```diff
@@ -149,9 +149,11 @@ end
     end
 end
 
-using LinearAlgebra
+using LinearAlgebra, Random
 
 @testset "generalized dot #32739" begin
+    Random.seed!(42) # https://github.com/JuliaLang/Compat.jl/issues/712
```
stevengj, line 155: This should be done at the top of the runtests.jl file so that it applies to all tests that use random numbers.

test/runtests.jl
```diff
@@ -149,9 +149,11 @@ end
     end
 end
 
-using LinearAlgebra
+using LinearAlgebra, Random
```
stevengj, line 152: You need to add Random as a test dependency to Project.toml, I think.

mcabbott: It's a full-time dependency already:

https://github.com/JuliaLang/Compat.jl/blob/master/Project.toml#L19

martinholters:
I'm a bit torn here. This is simple enough a fix to just apply it and be happy to have the issue fixed. OTOH, I prefer to only have random tests in cases where some property should really hold for any input data, which is not the case here. Thus, I think I'd prefer to just use fixed, explicitly given values for `A`, `x`, and `y`, I guess. Or alternative, restrict the randomly generated data to numerically benign cases, but I don't know how to achieve that.

mcabbott:
This is why I didn't want to change change random behaviour for all the tests, but thought changing it for just these would be the quick fix.

Perhaps ideally you'd reject & replace random vectors too close to zero dot, or something? But that seems like too much work.

martinholters:
> This is why I didn't want to change change random behaviour for all the tests, but thought changing it for just these would be the quick fix.

Yeah, that makes sense.

test/runtests.jl
martinholters: See https://github.com/JuliaLang/Compat.jl/pull/713#issuecomment-670015545 below: For tests which are actually supposed to run on random data, it may make sense not to provide a specific seed, so this is minimally invasive in that sense.

PR #715, opened 2020-07-20
Create Compat.Iterators; add takewhile, dropwhile and map

tkf:
This PR:

* creates a submodule `Compat.Iterators`
* adds `takewhile` and `dropwhile` from https://github.com/JuliaLang/julia/pull/33437
* adds `Iterators.map` from https://github.com/JuliaLang/julia/pull/34352

I added a submodule `Compat.Iterators` as otherwise I don't think it's possible to provide `Iterators.map`.

This PR includes the patch #714 since I need this for the test.

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -1,4 +1,16 @@
 module CompatIterators
+# Defining this module with the name `CompatIterators` and define
+# `const Iterators = CompatIterators` at the end of this file so that
+#
+#     julia> using Compat.Iterators
+#
+#     julia> Iterators
+#     Base.Iterators
+#
+# works without conflict in identifiers.  Users still need to
+# explicitly import `Compat.Iterators` via `using Compat: Iterators`
+# to use, e.g., `Iterators.map` before Julia 1.6.  This is the case
+# with and without the `CompatIterators` hack.
 
 using Base:
     @inline, Pair, AbstractDict, IndexLinear, IndexCartesian, IndexStyle, AbstractVector, Vector,
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -7,10 +7,13 @@ module CompatIterators
 #     julia> Iterators
 #     Base.Iterators
 #
-# works without conflict in identifiers.  Users still need to
-# explicitly import `Compat.Iterators` via `using Compat: Iterators`
-# to use, e.g., `Iterators.map` before Julia 1.6.  This is the case
-# with and without the `CompatIterators` hack.
+#     julia> takewhile
+#     takewhile (generic function with 1 method)
+#
+# works without conflict in identifiers in all Julia versions.  Users
+# still need to explicitly import `Compat.Iterators` via `using
+# Compat: Iterators` to use, e.g., `Iterators.map` before Julia 1.6.
+# This is the case with and without the `CompatIterators` hack.
 
 using Base: SizeUnknown
 
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -12,19 +12,12 @@ module CompatIterators
 # to use, e.g., `Iterators.map` before Julia 1.6.  This is the case
 # with and without the `CompatIterators` hack.
 
-using Base:
-    @inline, Pair, AbstractDict, IndexLinear, IndexCartesian, IndexStyle, AbstractVector, Vector,
-    tail, tuple_type_head, tuple_type_tail, tuple_type_cons, SizeUnknown, HasLength, HasShape,
-    IsInfinite, EltypeUnknown, HasEltype, OneTo, @propagate_inbounds, Generator, AbstractRange,
-    LinearIndices, (:), |, +, -, !==, !, <=, <, missing, any, @boundscheck, @inbounds
+using Base: SizeUnknown
 
-import .Base:
-    first, last,
-    isempty, length, size, axes, ndims,
-    eltype, IteratorSize, IteratorEltype,
-    haskey, keys, values, pairs,
-    getindex, setindex!, get, iterate,
-    popfirst!, isdone, peek
+# We normally use `Base.f(...) = ...` style for overloading.  However,
+# we use `import Base: f` style here to make synchronizing the code
+# with `Base` easier.
+import Base: IteratorEltype, IteratorSize, eltype, iterate
 
 # Import exported APIs
 for n in names(Base.Iterators)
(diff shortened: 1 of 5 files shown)


src/iterators.jl
```diff
@@ -0,0 +1,87 @@
+module CompatIterators
+
+using Base:
+    @inline, Pair, AbstractDict, IndexLinear, IndexCartesian, IndexStyle, AbstractVector, Vector,
+    tail, tuple_type_head, tuple_type_tail, tuple_type_cons, SizeUnknown, HasLength, HasShape,
+    IsInfinite, EltypeUnknown, HasEltype, OneTo, @propagate_inbounds, Generator, AbstractRange,
+    LinearIndices, (:), |, +, -, !==, !, <=, <, missing, any, @boundscheck, @inbounds
```
martinholters, line 7: Why all of these? Some should be available without explicit `using`, others seem not to be used at all?

src/iterators.jl
```diff
@@ -0,0 +1,87 @@
...
+
+import .Base:
+    first, last,
+    isempty, length, size, axes, ndims,
+    eltype, IteratorSize, IteratorEltype,
+    haskey, keys, values, pairs,
+    getindex, setindex!, get, iterate,
+    popfirst!, isdone, peek
```
martinholters, line 15: Personally, I prefer explicit qualification when adding methods over `import`ing them. That said, this list looks much longer than necessary anyway.

src/iterators.jl
```diff
@@ -0,0 +1,87 @@
+module CompatIterators
```
martinholters, line 1: Why can't this just be `Iterators`?

src/iterators.jl
```diff
@@ -0,0 +1,87 @@
...
+for n in names(Base.Iterators)
+    n === :Iterators && continue
+    @eval begin
+        using Base.Iterators: $n
+        export $n
+    end
+end
+
+# Import unexported public APIs
+using Base.Iterators: filter
```
martinholters, line 30: Could consider `using` everything from `names(Base.Iterators; all = true)` to be future-proof.

src/iterators.jl
tkf: I noted the rationale in

https://github.com/JuliaLang/Compat.jl/blob/38b0a37e21703105f68a205b3ed1181084f94788/src/iterators.jl#L1-L17

but I'm also OK with removing this hack.

src/iterators.jl
tkf: I've tried it but it'll try to import `eval` etc.

src/iterators.jl
tkf: I thought it was better for synchronizing with `julia`.  But I just realize that it doesn't work because a change like https://github.com/JuliaLang/julia/commit/331efe3ab8f8e8bb5fb42ecd5f92acb33df4de0c can introduce imports of some internal function from `Base`.

src/iterators.jl
tkf: I totally agree and I don't use this style usually.  I noted the motivation in

https://github.com/JuliaLang/Compat.jl/blob/38b0a37e21703105f68a205b3ed1181084f94788/src/iterators.jl#L20-L23

But I'm also OK with switching to the non-`import` style.

martinholters: As this is encapsulated in its own module, I'm OK with the `import` here, too.

src/iterators.jl
martinholters: That would only be `eval` and `include` I think, but all the names starting with `#` should maybe also be filtered out.

What about going the other way round: Not making `Compat.Iterators` a copy of `Base.Iterators` with extensions but just the extensions? I.e. `using Compat.Iterators` would only give you `takewhile` and `dropwhile` and you could also access `Compat.Iterators.map`, but that's it.

tkf: Hmm... For example, `using Compat: eachrow` works even for Julia 1.2 where `eachrow` is already in `Base`. It's nice that version-dependent branches are all in `Compat` and not in user packages. Wouldn't it make sense to keep this property for `Compat.Iterators` as well?  This way, you can just `using Compat: Iterators` and then `Iterators` is a good approximation of `Base.Iterators` of the latest Julia release.

martinholters: A misunderstanding, maybe: Of course, I'd want `using Compat.Iterators: takewhile` work across all Julia versions. It's just a question of whether we want, say, `using Compat.Iterators: flatten` to work at all. And I'm not saying we shouldn't, just that we don't have to and should think about it.

But what gave a pause here is this binding of `filter`. If `Base.Iterators` gets a new, unexported (but meant for public access) function, does that mean we _have_ to come back here and add it, too? If yes, that seems like an unfortunate maintenance burden. If no, then why do we bind `filter` here?

But I get the point about `using Compat: Iterators` bringing a good approximation of most recent `Iterators` into scope. That looks convincingly useful indeed.

We could also just leave `Compat.Iterators === Base.Iterators` on sufficiently recent Julia, but that has a gotcha: Say Julia 1.8 introduces a new, unexported function `foo` to `Base.Iterators`, then `Compat.Iterators` would automatically provide that (by being identical to `Base.Iterators`) on Julia 1.8. Sounds good; a stray forgotten `using Compat: Iterators` does no harm. But then Julia 1.9 adds some other function to `Iterators` and someone provides a `Compat.Iterators` implementation of it. Great! Except that after `pkg> up` (updating `Compat`), `Iterators.foo` starts erroring, although `Compat` did not increase it's major version. Not so great.

So I think the current approach is best indeed: If someone wants to use a feature not yet available in `Compat.Iterators`, then he should just stop `using Compat: Iterators` and bump the required Julia version. (Or better: Add it to `Compat.Iterators`.) But let me ponder this for a moment longer.

bramtayl: I'm not sure if it's really necessary for Compat.Iterators to export the entire contents of Base.Iterators, just like it's not necessary for Compat to export the entire contents of Base.

martinholters: The difference being that `using Base` is implicit everywhere (except for `baremodule`), while `using .Iterators` is not.

Still, I completely agree that it is not "really necessary for Compat.Iterators to export the entire contents of Base.Iterators". I do think that it would be useful though. The question is: Is there any downside to it? I'm afraid of this backfiring on us in some way, but I cannot make up a case where it actually would. I'm tempted to merge as is, but if anyone sees a problem with this, please do speak up.

bramtayl:
Yay! Tag?

tkf:
@martinholters Sorry that I wasn't replying promptly but I appreciate your detailed analysis! It overlaps with my thought process. Thanks for taking care of this PR.

PR #717, opened 2020-08-16
Fix Compat 2.X on Julia 1.6

fredrikekre:


fredrikekre:
Merging since tests pass on Julia 0.6 - 1.6.

PR #718, opened 2020-08-29
Add `contains`

nickrobinson251:
- add `contains(haystack, needle)` and curried-verion `contains(needle)` from https://github.com/JuliaLang/julia/pull/35132

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -594,6 +594,12 @@ if VERSION < v"1.2.0-DEV.257" # e7e726b3df1991e1306ef0c566d363c0a83b2dea
     Base.:(<)(x) = Base.Fix2(<, x)
 end
 
+# https://github.com/JuliaLang/julia/pull/35132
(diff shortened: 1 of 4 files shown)


fredrikekre (commented):
LGTM but can you add tests and an entry in the README?

src/Compat.jl
```diff
@@ -594,6 +594,12 @@ if VERSION < v"1.2.0-DEV.257" # e7e726b3df1991e1306ef0c566d363c0a83b2dea
     Base.:(<)(x) = Base.Fix2(<, x)
 end
 
+# https://github.com/JuliaLang/julia/pull/35132
+if VERSION < v"1.5.0-DEV.639" # cc6e121386758dff6ba7911770e48dfd59520199
+    contains(haystack::AbstractString, needle) = occursin(needle, haystack)
```
fredrikekre, line 599: Needs to be exported from Compat too (see test failures.)

nickrobinson251: ah, ofc - thanks!

PR #720, opened 2020-09-10
Backport ComposedFunction

tkf:
I suggest adding `Compat.ComposedFunction` when/if https://github.com/JuliaLang/julia/pull/37517 is merged. This way, individual packages like DataFrames.jl don't have to implement the same hack.

* [x] Wait for https://github.com/JuliaLang/julia/pull/37517
* [x] Add changelog in README
* [x] Bump version

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -614,6 +614,9 @@ if VERSION < v"1.6.0-" # TODO: specify the version when JuliaLang/julia#37517 is
         const ComposedFunction = let h = identity ∘ convert
             Base.typename(typeof(h)).wrapper
         end
+        @eval ComposedFunction{F,G}(f, g) where {F,G} =
+            $(Expr(:new, :(ComposedFunction{F,G}), :f, :g))
+        ComposedFunction(f::F, g::G) where {F,G} = ComposedFunction{F,G}(f, g)
     else
         using Base: ComposedFunction
     end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -608,7 +608,7 @@ if VERSION < v"1.5.0-DEV.438" # 0a43c0f1d21ce9c647c49111d93927369cd20f85
 end
 
 # https://github.com/JuliaLang/julia/pull/37517
-if VERSION < v"1.6.0" # TODO: specify the version when JuliaLang/julia#37517 is merged
+if VERSION < v"1.6.0-DEV.1037"
     export ComposedFunction
     # https://github.com/JuliaLang/julia/pull/35980
     if VERSION < v"1.6.0-DEV.85"
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -608,7 +608,7 @@ if VERSION < v"1.5.0-DEV.438" # 0a43c0f1d21ce9c647c49111d93927369cd20f85
 end
 
 # https://github.com/JuliaLang/julia/pull/37517
-if VERSION < v"1.6.0-" # TODO: specify the version when JuliaLang/julia#37517 is merged
+if VERSION < v"1.6.0" # TODO: specify the version when JuliaLang/julia#37517 is merged
     # https://github.com/JuliaLang/julia/pull/35980
     if VERSION < v"1.6.0-DEV.85"
         const ComposedFunction = let h = identity ∘ convert
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -609,6 +609,7 @@ end
 
 # https://github.com/JuliaLang/julia/pull/37517
 if VERSION < v"1.6.0" # TODO: specify the version when JuliaLang/julia#37517 is merged
+    export ComposedFunction
     # https://github.com/JuliaLang/julia/pull/35980
     if VERSION < v"1.6.0-DEV.85"
         const ComposedFunction = let h = identity ∘ convert
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -612,7 +612,7 @@ if VERSION < v"1.6.0-" # TODO: specify the version when JuliaLang/julia#37517 is
     # https://github.com/JuliaLang/julia/pull/35980
     if VERSION < v"1.6.0-DEV.85"
         const ComposedFunction = let h = identity ∘ convert
-            getfield(parentmodule(typeof(h)), nameof(typeof(h)))
+            Base.typename(typeof(h)).wrapper
         end
     else
         using Base: ComposedFunction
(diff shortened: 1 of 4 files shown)


src/Compat.jl
```diff
@@ -607,6 +607,33 @@ if VERSION < v"1.5.0-DEV.438" # 0a43c0f1d21ce9c647c49111d93927369cd20f85
     Base.startswith(s) = Base.Fix2(startswith, s)
 end
 
+# https://github.com/JuliaLang/julia/pull/37517
+if VERSION < v"1.6.0-" # TODO: specify the version when JuliaLang/julia#37517 is merged
+    # https://github.com/JuliaLang/julia/pull/35980
+    if VERSION < v"1.6.0-DEV.85"
+        const ComposedFunction = let h = identity ∘ convert
+            getfield(parentmodule(typeof(h)), nameof(typeof(h)))
```
martinholters, line 615: Just for my understanding, not that I propose changing this, but it is equivalent to `Base.typename(typeof(h)).wrapper`, right?

martinholters:
Would we want to add the constructors, too? It's a bit tricky, but something along these lines would work:
```julia
@eval ComposedFunction{F,G}(f,g) where {F,G} = $(Expr(:new, :(ComposedFunction{F,G}), :f, :g))
parentmodule(typeof(h)).eval(:($(nameof(typeof(h)))(f::F, g::G) where {F,G} = $(nameof(typeof(h))){F,G}(f,g)))
```

src/Compat.jl
tkf: I usually use this hack to rely only on the public API. But you and the other compiler devs watch out for Compat.jl so maybe it doesn't matter?

martinholters: This code is only run for past versions, so it can't be broken by future changes. (Unless we do something really devious that gets broken in a point release of 1.5).

tkf: Ah, yes, that's a good point. I'll change it.

src/Compat.jl
```diff
@@ -607,6 +607,36 @@ if VERSION < v"1.5.0-DEV.438" # 0a43c0f1d21ce9c647c49111d93927369cd20f85
...
+# https://github.com/JuliaLang/julia/pull/37517
+if VERSION < v"1.6.0-" # TODO: specify the version when JuliaLang/julia#37517 is merged
+    # https://github.com/JuliaLang/julia/pull/35980
+    if VERSION < v"1.6.0-DEV.85"
+        const ComposedFunction = let h = identity ∘ convert
+            Base.typename(typeof(h)).wrapper
+        end
+        @eval ComposedFunction{F,G}(f, g) where {F,G} =
+            $(Expr(:new, :(ComposedFunction{F,G}), :f, :g))
+        ComposedFunction(f::F, g::G) where {F,G} = ComposedFunction{F,G}(f, g)
```
tkf, line 619: Re https://github.com/JuliaLang/Compat.jl/pull/720#issuecomment-690917404, I think it makes sense so I added the constructor support. This way, people can look at the complexity of real working code when assessing the maintenance cost.

I've tried `parentmodule(...).eval(...)` approach but it threw

> `ERROR: LoadError: Evaluation into the closed module `Base` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `Base` with `eval` during precompilation - don't do this.`

in 1.5 (and I didn't understand why we can't just do `@eval` like the first line).  Fortunately, it looks like we can just add a method like normal type. Is this fine?

martinholters: That's much better. I have no idea what I was thinking when coming up with that complicated `eval` construct.

tkf: Ah, OK. Thanks.

src/Compat.jl
```diff
@@ -607,6 +607,36 @@ if VERSION < v"1.5.0-DEV.438" # 0a43c0f1d21ce9c647c49111d93927369cd20f85
...
+# https://github.com/JuliaLang/julia/pull/37517
+if VERSION < v"1.6.0" # TODO: specify the version when JuliaLang/julia#37517 is merged
+    # https://github.com/JuliaLang/julia/pull/35980
+    if VERSION < v"1.6.0-DEV.85"
+        const ComposedFunction = let h = identity ∘ convert
+            Base.typename(typeof(h)).wrapper
+        end
+        @eval ComposedFunction{F,G}(f, g) where {F,G} =
+            $(Expr(:new, :(ComposedFunction{F,G}), :f, :g))
+        ComposedFunction(f::F, g::G) where {F,G} = ComposedFunction{F,G}(f, g)
```
martinholters, line 619: Looking at the code in Base:
```suggestion
        ComposedFunction(f, g) = ComposedFunction{Core.Typeof(f),Core.Typeof(g)}(f, g)
```
I think this is better if `f` or `g` are types. For e.g. `f==Int`, `F` would be `DataType` but `Core.Typeof(f)` gives `Type{Int}`.

src/Compat.jl
```diff
@@ -607,6 +607,36 @@ if VERSION < v"1.5.0-DEV.438" # 0a43c0f1d21ce9c647c49111d93927369cd20f85
     Base.startswith(s) = Base.Fix2(startswith, s)
 end
 
+# https://github.com/JuliaLang/julia/pull/37517
+if VERSION < v"1.6.0" # TODO: specify the version when JuliaLang/julia#37517 is merged
```
jw3126:
```suggestion
if VERSION < v"1.6.0-DEV.1037"
```

src/Compat.jl
```diff
@@ -607,6 +607,37 @@ if VERSION < v"1.5.0-DEV.438" # 0a43c0f1d21ce9c647c49111d93927369cd20f85
     Base.startswith(s) = Base.Fix2(startswith, s)
 end
 
+# https://github.com/JuliaLang/julia/pull/37517
+if VERSION < v"1.6.0-DEV.1037"
+    export ComposedFunction
```
tkf, line 612: I added export since `Base` now exports `ComposedFunction`: https://github.com/JuliaLang/julia/pull/37517#issuecomment-693606076

test/runtests.jl
```diff
@@ -556,6 +556,22 @@ end
...
 end
 
+# https://github.com/JuliaLang/julia/pull/37517
+@testset "ComposedFunction" begin
+    @test sin ∘ cos isa Compat.ComposedFunction
+    @test sin ∘ cos === Compat.ComposedFunction(sin, cos)
+    c = sin ∘ cos
+    @test c.outer === sin
+    @test c.inner === cos
+    if VERSION < v"1.6.0" # TODO: specify the version when JuliaLang/julia#37517 is merged
```
jw3126:
```suggestion
    if VERSION < v"1.6.0-DEV.1037"
```

src/Compat.jl
tkf: Looks like we had a race condition :)

jw3126 (approved):
LGTM

PR #723, opened 2020-10-07
Poorman's 5-arg mul!

### LINKED ISSUE #681: 5-arg mul!s?
Should the 5-arg `mul!`s be included in Compat.jl, i.e. `mul!(w, A, b, α=1, β=0)`? I would like to start writing code like this, but retain compatibility with Julia 1.x, x < 3.

oxinabox:
This would close #681 
It's very much a simplified stand-in.
Doesn't get the same performance, but shouldn't be terrible (not that I have run benchmarks)

If people think that that is a bad idea then i can just drop this.
Otherwise if people think its good, I can add tests and stuff to finish it off.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -745,7 +745,7 @@ if VERSION < v"1.3.0-alpha.115"
 
     function _mul!(C, A, B, alpha, beta)
         Y = similar(C)
-        mul!(Y, A, B)
+        LinearAlgebra.mul!(Y, A, B)
         C .= Y .* alpha .+ C .* beta
         return C
     end
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -739,7 +739,8 @@ if VERSION < v"1.6.0-DEV.1083"
     end
 end
 
-if VERSION < v"1.3"
+if VERSION < v"1.3.0-alpha.115"
+    # https://github.com/JuliaLang/julia/pull/29634
     # Note this is much less performant than real 5-arg mul!, but is provided so old versions of julia don't error at least
 
     function _mul!(C, A, B, alpha, beta)
@@ -748,6 +749,7 @@ if VERSION < v"1.3"
         C .= Y .* alpha .+ C .* beta
         return C
     end
+
     # all combination of Number and AbstractArray for A and B except both being Number
     function LinearAlgebra.mul!(C::AbstractArray, A::Number, B::AbstractArray, alpha::Number, beta::Number)
         return _mul!(C, A, B, alpha, beta)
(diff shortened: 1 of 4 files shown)


martinholters:
Are there any realistic situations imaginable where people think "ah, this piece of code could well be replaced with a 5-arg-`mul!`" but the version in this PR is slower than whatever they had before? If no, I'd definitely be fine with this.

src/Compat.jl
```diff
@@ -739,6 +739,27 @@ if VERSION < v"1.6.0-DEV.1083"
...
 end
 
+if VERSION < v"1.3"
+    # Note this is much less performant than real 5-arg mul!, but is provided so old versions of julia don't error at least
+
+    function _mul!(C, A, B, alpha, beta)
+        Y = similar(C)
+        mul!(Y, A, B)
+        Y .= alpha.*Y .+ beta.*C
+        return Y
```
dkarrasch, line 749:
```suggestion
        C .= Y .* alpha .+ C .* beta
        return C
```
(1) It should write into `C`, and (2) the scalars are formally multiplied from the right (doesn't matter for `RealOrComplex`, but still).

dkarrasch (commented):
I think this is good, since it benefits from specialized 3-arg `mul!`s. Do you want to restrict `A, B, C` to `AbstractVecOrMat` or something like that, or split if there are ambiguities? Well, or let `mul!` catch it, I guess that's fine.

oxinabox:
> Are there any realistic situations imaginable where people think "ah, this piece of code could well be replaced with a 5-arg-mul!" but the version in this PR is slower than whatever they had before? If no, I'd definitely be fine with this.

If the alternative is them calling hand written GEMM calls then yes.
e.g. I am wanting to replace this complex (and currently broken) code in Nabla https://github.com/invenia/Nabla.jl/blob/4cadc87677fb1187354999dcf93bd528f45f85d0/src/sensitivities/linalg/strided.jl#L25-L26
with 5-arg mul! calls, to simplify it.
But I don't want to fully drop 1.0 support.
(I am ok with it performing suboptimially)

oxinabox:
> Do you want to restrict A, B, C to AbstractVecOrMat or something like that, or split if there are ambiguities?

There won't be ambiguities since if not using 1.3+ there will be no other `mul!`

I duplicated the most generic signature that I found for 5 arg `mul!` in julia 1.5

src/Compat.jl
```diff
@@ -739,6 +739,29 @@ if VERSION < v"1.6.0-DEV.1083"
     end
 end
 
+if VERSION < v"1.3.0-alpha.115"
+    # https://github.com/JuliaLang/julia/pull/29634
+    # Note this is much less performant than real 5-arg mul!, but is provided so old versions of julia don't error at least
+
+    function _mul!(C, A, B, alpha, beta)
+        Y = similar(C)
+        mul!(Y, A, B)
```
oxinabox, line 748:
```suggestion
        LinearAlgebra.mul!(Y, A, B)
```

PR #724, opened 2020-10-07
add parseatom and parseall

simeonschaub:
Especially `parseatom` is quite useful for implementing custom string interpolation, so it would definitely be nice to be able to use it on previous versions of Julia.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,13 @@ changes in `julia`.
 
 ## Supported features
 
+* `Compat.parseatom(text::AbstractString, pos::Integer; filename="none")` parses a single
+  atom from `text` starting at index `pos`. Returns a `Tuple` consisting of the
+  parsed expression and the index to resume parsing from. ([#35243]) (since Compat 3.20)
+
+* `Compat.parseall(text::AbstractString; filename="none")` parses the whole string `text`
+  and returns the parsed expression. ([#35243]) (since Compat 3.20)
+
 * `mul!(C, A, B, alpha, beta)` now performs in-place multiply add ([#29634]). (since Compat 3.19).
 
 * `reinterpret(reshape, T, a::AbstractArray{S})` reinterprets `a` to have eltype `T` while
@@ -211,3 +218,4 @@ Note that you should specify the correct minimum version for `Compat` in the
 [#37517]: https://github.com/JuliaLang/julia/pull/37517
 [#37559]: https://github.com/JuliaLang/julia/pull/37559
 [#29634]: https://github.com/JuliaLang/julia/pull/29634
+[#35243]: https://github.com/JuliaLang/julia/pull/35243
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -764,21 +764,43 @@ end
 
 # https://github.com/JuliaLang/julia/pull/35243
 if VERSION < v"1.6.0-DEV.15"
-    _replace_filename(@nospecialize(x), filename) = x
-    _replace_filename(x::LineNumberNode, filename) = LineNumberNode(x.line, filename)
-    function _replace_filename(ex::Expr, filename)
-        return Expr(ex.head, Any[_replace_filename(i, filename) for i in ex.args]...)
+    _replace_filename(@nospecialize(x), filename, line_offset=0) = x
+    function _replace_filename(x::LineNumberNode, filename, line_offset=0)
+        return LineNumberNode(x.line + line_offset, filename)
+    end
+    function _replace_filename(ex::Expr, filename, line_offset=0)
+        return Expr(
+            ex.head,
+            Any[_replace_filename(i, filename, line_offset) for i in ex.args]...,
+        )
     end
 
     function parseatom(text::AbstractString, pos::Integer; filename="none")
         ex, i = Meta.parse(text, pos, greedy=false)
         return _replace_filename(ex, Symbol(filename)), i
     end
+
+    function _skip_newlines(text, line, i)
+        while i <= lastindex(text) && isspace(text[i])
+            line += text[i] == '\n'
+            i = nextind(text, i)
+        end
+        return line, i
+    end
     
     function parseall(text::AbstractString; filename="none")
-        _filename = Symbol(filename)
-        ex = _replace_filename(Meta.parse(text), _filename)
-        return Expr(:toplevel, LineNumberNode(1, _filename), ex)
+        filename = Symbol(filename)
+        ex = Expr(:toplevel)
+        line, prev_i = _skip_newlines(text, 1, firstindex(text))
+        ex_n, i = Meta.parse(text, prev_i)
+        while ex_n !== nothing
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -55,6 +55,13 @@ changes in `julia`.
 
 ## Supported features
 
+* `parseatom(text::AbstractString, pos::Integer; filename="none")` parses a single
```
martinholters, line 58:
```suggestion
* `Compat.parseatom(text::AbstractString, pos::Integer; filename="none")` parses a single
```

README.md
```diff
@@ -55,6 +55,13 @@ changes in `julia`.
...
+  atom from `text` starting at index `pos`. Returns a `Tuple` consisting of the
+  parsed expression and the index to resume parsing from. ([#35243]) (since Compat 3.??)
+
+* `parseall(text::AbstractString; filename="none")` parses the whole string `text`
```
martinholters, line 62:
```suggestion
* `Compat.parseall(text::AbstractString; filename="none")` parses the whole string `text`
```

martinholters (approved):
LGTM modulo the minor tweaks in the README, but I don't think I've ever used these functions and as this is not just copy/paste job, a second review by someone who knows more about these would be welcome.

README.md
```diff
@@ -55,6 +55,13 @@ changes in `julia`.
 
 ## Supported features
 
+* `Compat.parseatom(text::AbstractString, pos::Integer; filename="none")` parses a single
+  atom from `text` starting at index `pos`. Returns a `Tuple` consisting of the
+  parsed expression and the index to resume parsing from. ([#35243]) (since Compat 3.20)
```
martinholters, line 60: I think the link in the list at the end of the file is missing.

simeonschaub: Ah, right! I was wondering how those links were supposed to work and missed that.

src/Compat.jl
```diff
@@ -762,6 +762,27 @@ if VERSION < v"1.3.0-alpha.115"
...
+        return Expr(ex.head, Any[_replace_filename(i, filename) for i in ex.args]...)
+    end
+
+    function parseatom(text::AbstractString, pos::Integer; filename="none")
+        ex, i = Meta.parse(text, pos, greedy=false)
+        return _replace_filename(ex, Symbol(filename)), i
+    end
+    
+    function parseall(text::AbstractString; filename="none")
+        ex = Meta.parse(text)
```
fredrikekre, line 779: What is the difference between `Meta.parse` and `Meta.parseall`?

fredrikekre (approved):
LGTM. Squash and rerun CI?

src/Compat.jl
simeonschaub: It respects the `filename` for macrocalls and the likes and also wraps everything in an `Expr(:toplevel, ...)`. Since the `toplevel` expression head was only added in 1.5, I thought it would be ok to skip the latter, but we could perhaps support that for Julia 1.5.

fredrikekre: `:toplevel` is not new for 1.5 though?

simeonschaub: Oh, you are right! I confused it with `softscope`. I think it makes sense to add that then,

simeonschaub: Done.

src/Compat.jl
```diff
@@ -762,6 +762,28 @@ if VERSION < v"1.3.0-alpha.115"
...
+    end
+
+    function parseatom(text::AbstractString, pos::Integer; filename="none")
+        ex, i = Meta.parse(text, pos, greedy=false)
+        return _replace_filename(ex, Symbol(filename)), i
+    end
+    
+    function parseall(text::AbstractString; filename="none")
+        _filename = Symbol(filename)
+        ex = _replace_filename(Meta.parse(text), _filename)
```
fredrikekre, line 780: I think you need the other version of `Meta.parse` here and keep parsing from the returned index since `Meta.parse(::String)` errors if there are more than one expression:
```julia
julia> str = """
       a = 1
       b = 2
       """
"a = 1\nb = 2\n"

julia> Meta.parse(str)
ERROR: Base.Meta.ParseError("extra token after end of expression")
Stacktrace:
 [1] parse(str::String; raise::Bool, depwarn::Bool)
   @ Base.Meta ./meta.jl:222
 [2] parse(str::String)
   @ Base.Meta ./meta.jl:217
 [3] top-level scope
   @ REPL[6]:1

julia> Meta.parseall(str)
:($(Expr(:toplevel, :(#= none:1 =#), :(a = 1), :(#= none:2 =#), :(b = 2))))
```

src/Compat.jl
```diff
@@ -762,6 +762,50 @@ if VERSION < v"1.3.0-alpha.115"
...
+
+    function parseatom(text::AbstractString, pos::Integer; filename="none")
+        ex, i = Meta.parse(text, pos, greedy=false)
+        return _replace_filename(ex, Symbol(filename)), i
+    end
+
+    function _skip_newlines(text, line, i)
+        while i <= lastindex(text) && isspace(text[i])
+            line += text[i] == '\n'
+            i += 1
```
simeonschaub, line 786:
```suggestion
            i = nextind(text, i)
```

src/Compat.jl
```diff
@@ -762,6 +762,50 @@ if VERSION < v"1.3.0-alpha.115"
...
+    
+    function parseall(text::AbstractString; filename="none")
+        filename = Symbol(filename)
+        ex = Expr(:toplevel)
+        line, prev_i = _skip_newlines(text, 1, 1)
+        ex_n, i = Meta.parse(text, prev_i)
+        while ex_n !== nothing
+            push!(ex.args, LineNumberNode(line, filename))
+            push!(ex.args, _replace_filename(ex_n, filename, line-1))
+            line += count(==('\n'), SubString(text, prev_i:i-1))
```
simeonschaub, line 799:
```suggestion
            line += count(==('\n'), SubString(text, prev_i:prevind(text, i)))
```

src/Compat.jl
simeonschaub: Should be fixed now.

src/Compat.jl
```diff
@@ -762,6 +762,50 @@ if VERSION < v"1.3.0-alpha.115"
...
+    
+    function parseall(text::AbstractString; filename="none")
+        filename = Symbol(filename)
+        ex = Expr(:toplevel)
+        line, prev_i = _skip_newlines(text, 1, firstindex(text))
+        ex_n, i = Meta.parse(text, prev_i)
+        while ex_n !== nothing
+            push!(ex.args, LineNumberNode(line, filename))
+            push!(ex.args, _replace_filename(ex_n, filename, line-1))
+            line += count(==('\n'), SubString(text, prev_i:prevind(text, i)))
```
fredrikekre, line 799: This doesn't count `\r\n` line endings. Perhaps just add
```
text = replace(text, "\r\n" => "\n")
```
on the first line of this function is the simplest solution.

PR #726, opened 2020-10-15
Add `Dates.canonicalize(::Period)`

nickrobinson251:
- Added to Dates stdlib in  https://github.com/JuliaLang/julia/pull/37391

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.21.0"
+version = "3.22.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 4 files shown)


martinholters:
https://github.com/JuliaLang/julia/pull/37391 brings some more tests. Would it make sense to also include them here?

nickrobinson251:
> [JuliaLang/julia#37391](https://github.com/JuliaLang/julia/pull/37391) brings some more tests. Would it make sense to also include them here?

🤷‍♂️ I can bring them across if you like? I didn't since one test is enough to test the dispatch this PR adds (and those other tests are testing that the `canonicalize` logic in Dates produces the expected results). Would you prefer i bring them across to match the linked PR?

martinholters:
Yes, I think I'd prefer to bring them along. We can easily afford the extra run-time that costs and it would help catch any nasty surprises.

nickrobinson251:
Cool :) i've brought over all the tests from the original PR and rebased to resolve conflicts

PR #730, opened 2020-11-02
RFC: Switch to Actions for CI

martinholters:
I'm not aware of the pros and cons here so I thought I put this up for discussion. Note that this is somewhat more extensive than the old configuration which did not test on 32bit linux.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,22 @@
...
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: ['1.0', '1.1', '1.2', '1.3', '1.4', '1', 'nightly']
+        julia-arch: [x64, x86]
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+        exclude:
+          - os: macOS-latest
+            julia-arch: x86
```
simeonschaub, line 15:
```suggestion
            julia-arch: x86
     fail-fast: false
```
Otherwise a single failing job will cancel all other jobs.

timholy:
Shoot, I didn't see this. Superseded by #739. FYI you can easily update many projects almost automatically with https://github.com/julia-actions/MassInstallAction.jl/tree/master/demos/MaintenanceDec2020, although when I'm doing a single repo I often don't submit them as multiple PRs and instead manually combine all maintenance action into a single PR.

PR #731, opened 2020-11-24
tryparse VersionNumber

mattBrzezinski:
Looking to add in the [tryparse](https://github.com/JuliaLang/julia/blob/master/base/version.jl#L108) functionality from `master` here.

I'm not sure the proper procedure for requesting changes here but I've tested this functionality locally with Julia `1.0.5, 1.3.0, 1.5.0` and it checks out. ~~I'm also a bit unsure if I used the proper version in `if VERSION < v"X.Y.Z"`, as I was unable to find any of the `DEV` releases.~~

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -820,6 +820,27 @@ if VERSION < v"1.6.0-DEV.292" # 6cd329c371c1db3d9876bc337e82e274e50420e8
     sincospi(x) = (sinpi(x), cospi(x))
 end
 
+# https://github.com/JuliaLang/julia/pull/37803
+if VERSION < v"1.5.3"
+    function Base.tryparse(::Type{VersionNumber}, v::AbstractString)
+        VerTuple = Tuple{Vararg{Union{UInt64,String}}}
+
+        v == "∞" && return typemax(VersionNumber)
+        m = match(Base.VERSION_REGEX, v)
+        m === nothing && return nothing
+        major, minor, patch, minus, prerl, plus, build = m.captures
+        major = parse(Base.VInt, major)
+        minor = minor !== nothing ? parse(Base.VInt, minor) : Base.VInt(0)
+        patch = patch !== nothing ? parse(Base.VInt, patch) : Base.VInt(0)
+        if prerl !== nothing && !isempty(prerl) && prerl[1] == '-'
+            prerl = prerl[2:end] # strip leading '-'
+        end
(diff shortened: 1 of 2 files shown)


iamed2:
The version bound is `1.6.0-DEV.1090`, found using `contrib/commit-name.sh` in the julia repo.

src/Compat.jl
```diff
@@ -820,6 +820,27 @@ if VERSION < v"1.6.0-DEV.292" # 6cd329c371c1db3d9876bc337e82e274e50420e8
...
 end
 
+# https://github.com/JuliaLang/julia/pull/37803
+if VERSION < v"1.6.0-DEV.1090"
+    function Base.tryparse(::Type{VersionNumber}, v::AbstractString)
+        VerTuple = Tuple{Vararg{Union{UInt64,String}}}
+
+        v == "∞" && return typemax(VersionNumber)
+        m = match(Base.VERSION_REGEX, v)
+        m === nothing && return nothing
```
stevengj, line 830: Couldn't you just do `return m === nothing ? nothing : parse(VesionNumber, v)` here?

(Slightly less efficient, since it will match the regex twice, but code size seems more of a priority for Compat.jl.)

mattBrzezinski: I took this directly from [base/versions.jl](https://github.com/JuliaLang/julia/blob/master/base/version.jl#L108-L122), slightly modified to include the `VerTuple` type, and a few `Base.` prefaces.

stevengj: I understand, I'm just suggesting that the Compat version can be simpler; it doesn't need to be as optimized as the Base version.

mattBrzezinski: That makes sense, my only concern with modfiying functions that are ported into this package is having adequate testing to ensure they do exactly what they're doing in Base.

With that said, I'm not quite following along with your suggestion for change. Are you suggesting bringing in [parse(VersionNumber, String)](https://github.com/JuliaLang/julia/blob/master/base/version.jl#L124) from Base in, It looks like we'd overflow at that point?

stevengj: Why would you overflow?

stevengj: Sorry, I mean calling
```jl
return m === nothing ? nothing : VersionNumber(v)
```

The point is, all of the functionality that you copy-and-pasted here after the regex is the same as `VersionNumber(v::String)` in earlier Julia versions.

mattBrzezinski:
Closing this as I don't actually need these changes.

PR #733, opened 2020-12-10
`startswith` and `endswith` that work with `Regex`

fchorney:
Implements `startswith(s,r)` and `endswith(s,r)` where `r` is a `Regex`

https://github.com/JuliaLang/julia/pull/29790

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ without incrementing the major version of Compat.jl if necessary to match
 changes in `julia`.
 
 ## Supported features
-* `startswith(s, r)`, and `endswith(s, r)` where `r` is a `regex` ([#29790]) (since Compat 3.24)
+* `startswith(s, r::Regex)` and `endswith(s, r::Regex)` ([#29790]) (since Compat 3.24)
 
 * `sincospi(x)` for calculating the tuple `(sinpi(x), cospi(x))` ([#35816]) (since Compat 3.23)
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -54,6 +54,7 @@ without incrementing the major version of Compat.jl if necessary to match
 changes in `julia`.
 
 ## Supported features
+* `startswith(s, r)`, and `endswith(s, r)` where `r` is a `regex` ([#29790]) (since Compat 3.24)
```
omus, line 57:
```suggestion
* `startswith(s, r::Regex)` and `endswith(s, r::Regex)` ([#29790]) (since Compat 3.24)
```

omus (approved):

test/runtests.jl
```diff
@@ -802,3 +802,18 @@ import LinearAlgebra
     @test ImportRename.hm === LinearAlgebra.BLAS.hemm
     @test !isdefined(ImportRename, :hemm)
 end
+
+# https://github.com/JuliaLang/julia/pull/29790
+@testset "regex startswith and endswith" begin
```
mjram0s, line 807: Not familiar with the `PCRE` stuff but what happens if the regex is longer than 1 character? Should you add tests for that?

fchorney: I just pulled the same tests out of the Julia commit, but I could add a few more just in case

mjram0s (approved):

PR #738, opened 2021-03-29
Support 2-arg argmin/argmax/findmin/findmax

timholy:
xref https://github.com/JuliaLang/julia/pull/35316

Thanks for adding this to Julia, @cmcaine! I like it so much I want to use it for packages that support current/older Julia versions.


ericphanson:
Since Compat had a release with this, updating to https://github.com/JuliaLang/julia/pull/41076 will mean Compat v4, right? (I’m actually using the Compat v3 version in some code so I am concerned about breakage).

PR #742, opened 2021-05-14
Add `@something` and `@coalesce`

omus:
Added in https://github.com/JuliaLang/julia/pull/40729


omus:
Just to be safe I'm going to rebase this change once https://github.com/JuliaLang/Compat.jl/pull/743 is merged.

PR #739, opened 2021-03-31
Switch to GitHub Actions for CI

timholy:
Also updates:
- TagBot workflow for triggers
- adds version-check on a test that no longer errors
  on Julia 1.7 (#39228)

closes  #736

PR #743, opened 2021-05-14
Run CI against Julia 1.0

omus:
Using `1.0` in YAML gets automatically converted to `1` which results in the latest version of Julia 1 being run and not the latest version of 1.0. I quoted the other Julia version numbers just to be consistent. 

Introduced in #739


omus:
🎉 Luckily none of the Compat additions since we switched to GHA broke Julia 1.0 support

simeonschaub (approved):
LGTM

Issue #745, opened 2021-06-23
`findmax` does different things on Julia 1.6 and 1.7 with the same version of Compat


ericphanson:
```bash
❯ julia-1.7 -q
julia> using Compat # v3.31.0

julia> findmax(identity, 5:9)
(9, 5)

julia> VERSION
v"1.7.0-beta2"

❯ julia-1.6 -q
julia> using Compat # v3.31.0

julia> findmax(identity, 5:9)
(9, 9)

julia> VERSION
v"1.6.1"
```

This is, of course, since the choice of what `findmax` does changed between when #738 was made and the 1.7-beta2 release, and in Compat, the definitions are gated around versions, so on 1.7 the real definition is let through, while on 1.6, the Compat definition is used.

But how do we fix this?

fredrikekre:
Just update to match 1.7-beta?

ericphanson:
Yeah, so Compat v4 can do that, and I guess v3 is just going to be inconsistent?

fredrikekre:
Why not in v3? This is a bug in Compat.

ericphanson:
Is it though? Changing the definition basically means any code using Compat for the findmax methods added in #738 will break, and it was very explicit about what those methods did.

Compat tagged a release with those methods, Julia did not, so Julia can change them, but I don't think Compat can...

fredrikekre:
I guess it was premature to tag the Compat release, but the job of Compat is to behave like Julia, and there have not been a Julia release with this behavior yet, so I would 100% call this a bug in Compat.

PR #746, opened 2021-06-26
Port Base.current_exceptions() to older Julia versions

c42f:
See https://github.com/JuliaLang/julia/pull/29901

This is limited to julia-1.1 and above because earlier versions don't
have the necessary runtime library support (`Base.catch_stack()` etc).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.33.0"
+version = "3.34.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ changes in `julia`.
 
 ## Supported features
 
+* The function `current_exceptions()` has been added to get the current exception stack. ([#29901]) (since Compat 3.34)
+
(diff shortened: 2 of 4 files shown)


c42f:
Failing tests on nightly are #745

martinholters:
With a README entry and after a rebase, this looks good to me.

c42f:
I added the readme entry and rebased.

Also I've added julia-1.0 compatibility (in that case, the best we can do is just return the current exception as a single-element stack)

martinholters (approved):
LGTM superficially, but you probably know better than me whether this does the right thing.

But maybe we should mention the limited functionality on Julia 1.0 in the README?

c42f:
> whether this does the right thing.

It should do the right thing to the extent that the right thing can be done at all in julia 1.0 (there were some nasty bugs with exceptions and backtraces going missing completely in julia-1.0 due to task switching. But Compat can't work around that.)

I figured it was better to have the API with only approximate runtime compatibility in 1.0, than to not have the API at all — it's a lot easier for users if they can at least rely on the API being present on all versions supported by Compat.

martinholters:
I don't disagree. Only question: Merge as is or mention the limitation re: Julia 1.0 in the README?

c42f:
Yeah, it's worth mentioning the Julia-1.0 limitations in the readme. I did that.

martinholters:
Alright, there was plenty of time for others to chime in on this, bu no one did, so here we go...

Issue #699, opened 2020-05-07
Implicit named tuple and keyword argument names


garrison:
Not sure if anybody is interested in making `@compat` for [implicit named tuple and keyword argument names](https://github.com/JuliaLang/julia/pull/34331) (introduced in julia 1.5), but it would be welcome.  There is [an existing package for this](https://github.com/jw3126/EponymKeywordSyntax.jl) which works on older versions of julia.  It could be the start of such functionality here.  More discussion in  jw3126/EponymKeywordSyntax.jl#3.

PR #747, opened 2021-07-28
Port implicit named tuple and keyword argument names

omus:
Fixes https://github.com/JuliaLang/Compat.jl/issues/699. I implemented this without noticing that issue so this implementation was created independently from the mentioned package

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ changes in `julia`.
 
 ## Supported features
 
-* Implicit keywords arguments or named tuples is supported using the `@compat` macro, e.g. `@compat f(; x, f.p)` and `@compat (; x, f.p)` is equivalent to `f(; x=x, p=f.p)` and `(; x=x, f.p)` respectively. ([#34331]) (since Compat 3.32.0)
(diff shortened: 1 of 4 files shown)


omus:
Rebased to use the fix for #745

README.md
```diff
@@ -54,6 +54,8 @@ changes in `julia`.
 
 ## Supported features
 
+* Implicit keywords arguments or named tuples is supported using the `@compat` macro, e.g. `@compat f(; x, f.p)` and `@compat (; x, f.p)` is equivalent to `f(; x=x, p=f.p)` and `(; x=x, f.p)` respectively. ([#34331]) (since Compat 3.32.0)
```
omus, line 57:
```suggestion
* Implicit keywords arguments or named tuples is supported using the `@compat` macro, e.g. `@compat f(; x, f.p)` and `@compat (; x, f.p)` is equivalent to `f(; x=x, p=f.p)` and `(; x=x, p=f.p)` respectively. ([#34331]) (since Compat 3.32.0)
```

PR #748, opened 2021-07-28
Fix 2-arg argmin/argmax/findmin/findmax

omus:
Updates #738 to use the changes made to https://github.com/JuliaLang/julia/pull/41076. Fixes https://github.com/JuliaLang/Compat.jl/issues/745.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.31.0"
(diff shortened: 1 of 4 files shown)


omus:
As discussed in https://github.com/JuliaLang/Compat.jl/issues/745#issuecomment-866862784 although the behaviour of Compat.jl is changing this counts as non-breaking as the original behaviour was never officially released.

omus:
As this is a bugfix I've marked it as such. If someone cared to do so this could be backported as 3.26.1

PR #751, opened 2021-09-07
Add `Returns`

mcabbott:
From https://github.com/JuliaLang/julia/pull/39794


martinholters (approved):
LGTM, but might want to add a test for the explicitly typed constructor, although it appears that is also not tested in base...

PR #755, opened 2021-09-21
Test against Julia 1.7 on CI

omus:
Now that "nightly" runs Julia 1.8 we should test against Julia 1.7 which is currently RC1

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -20,6 +20,7 @@ jobs:
           - "1.4"
           - "1.5"
           - "1.6"
+          - "1.7"
           - nightly
         os:
           - ubuntu-latest
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -20,7 +20,7 @@ jobs:
           - "1.4"
           - "1.5"
           - "1.6"
-          - "1.7"
+          - "~1.7.0-0"
           - nightly
         os:
           - ubuntu-latest


.github/workflows/CI.yml
```diff
@@ -20,6 +20,7 @@ jobs:
           - "1.4"
           - "1.5"
           - "1.6"
+          - "1.7"
```
simeonschaub, line 23: Since 1.7 is not released yet, I think it needs to be:

```suggestion
          - "1.7-nightly"
```

omus: Interesting. When did that change? Does that mean that "1.7-nightly" will break when 1.7 is released?

simeonschaub: I don't think it will break, but I believe it will always follow the `release-1.7` branch, so it may still be ahead of the latest official 1.7 release.

DilumAluthge: I think what you want here is either `'~1.7.0-0'` or `'^1.7.0-0'`.

https://github.com/julia-actions/setup-julia/blob/master/README.md#julia-versions

simeonschaub: Yeah, that should be fine as well. I don't think the difference really matters.

omus: > `1.7-nightly` ... This version will only be available during certain phases of the Julia release cycle.

I'll go with `'~1.7.0-0'` to be on the safe side

omus:
```suggestion
          - "~1.7.0-0"
```

PR #756, opened 2021-09-22
Add support for Base.display_error over ExceptionStack.

Sacha0:
https://github.com/JuliaLang/Compat.jl/pull/746 added support for `current_exceptions`/`ExceptionStack` (née `catch_stack`), and augmented a few functions with methods that commonly handle `ExceptionStack`s. This pull request extends that to support to `Base.display_error`, which also commonly handles `ExceptionStack`s. Best! :)

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1156,6 +1156,17 @@ if VERSION < v"1.7.0-DEV.1106"
         end
     end
 
+    if VERSION >= v"1.2"
+        # Base.scrub_repl_backtrace only exists as of Julia 1.2,
+        # hence the immediately preceding version guard.
+        function Base.display_error(io::IO, stack::ExceptionStack)
+            printstyled(io, "ERROR: "; bold=true, color=Base.error_color())
+            bt = Any[ (x[1], Base.scrub_repl_backtrace(x[2])) for x in stack ]
+            show_exception_stack(IOContext(io, :limit => true), bt)
+            println(io)
+        end
+    end
+
     function Base.show(io::IO, ::MIME"text/plain", stack::ExceptionStack)
         nexc = length(stack)
         printstyled(io, nexc, "-element ExceptionStack", nexc == 0 ? "" : ":\n")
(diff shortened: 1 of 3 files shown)


Sacha0:
It looks like Julia 1.1 lacks `Base.scrub_repl_backtrace`; I will guard the new definition and associated test to limit those to >= Julia 1.2.

Sacha0:
CI appears happy now; review much appreciated :).

Sacha0:
Rebased over https://github.com/JuliaLang/Compat.jl/pull/758. What say you, CI? :)

Sacha0:
@c42f, might I trouble you to have a look at this related patch as well? Thanks again! :)

src/Compat.jl
```diff
@@ -1156,6 +1156,17 @@ if VERSION < v"1.7.0-DEV.1106"
         end
     end
 
+    if VERSION >= v"1.2"
+        # Base.scrub_repl_backtrace only exists as of Julia 1.2,
+        # hence the immediately preceding version guard.
+        function Base.display_error(io::IO, stack::ExceptionStack)
+            printstyled(io, "ERROR: "; bold=true, color=Base.error_color())
+            bt = Any[ (x[1], Base.scrub_repl_backtrace(x[2])) for x in stack ]
```
c42f, line 1164: Maybe better to just supply this API for consistency, but use a small hack to make it work on 1.0

```suggestion
            bt = Any[ (x[1], (VERSION >= v"1.2" ? Base.scrub_repl_backtrace : identity)(x[2])) for x in stack ]
```

c42f (approved):
Looks good to me, I think this can be merged as is, or for a slight improvement also tweaked to work on 1.0 and 1.1

src/Compat.jl
Sacha0: Great idea! I will give that a whirl. Thanks for the suggestion and for reviewing Chris! Much appreciated :).

Sacha0: Implemented the suggestion and expanded the tests. Thanks again Chris!

NHDaly (commented):
LGTM. This is cool - i didn't realize Compat.jl monkey-patched (type pirated?) methods like this to backport them outside julia itself. Neat! :)

PR #758, opened 2021-09-23
Fix current_exceptions when called with backtrace=false (#757).

### LINKED ISSUE #757: current_exceptions implementation fails when backtrace=false
#746 introduced support for `current_exceptions` (née `catch_stack`), which bears a `backtrace` keyword argument. When `backtrace=false`, the returned `ExceptionStack`'s `stack` field should be a `Vector` of `NamedTuple`s where the second (backtrace) element of each tuple is `nothing`. Instead the Compat implementation throws, e.g. on Julia 1.6.2:
```
julia> using Compat: current_exceptions

julia> try
           throw(UndefVarError(:cat))
       catch
           current_exceptions(;backtrace=false)
       end
ERROR: MethodError: no method matching getindex(::UndefVarError, ::Int64)
Stacktrace:
 [1] current_exceptions(task::Task; backtrace::Bool)
   @ Compat ~/pkg/Compat.jl/src/Compat.jl:1105
 [2] top-level scope
   @ REPL[7]:4

caused by: UndefVarError: cat not defined
Stacktrace:
 [1] top-level scope
   @ REPL[7]:2
```
The trick is that the implementation
```
function current_exceptions(task=current_task(); backtrace=true)
    stack = Base.catch_stack(task, include_bt=backtrace)
    ExceptionStack(Any[(exception=x[1],backtrace=x[2]) for x in stack])
end
```
passes the `backtrace` kwarg through to `catch_stack` which, when `include_bt=false`, emits a `Vector` of exceptions, rather than a `Vector` of two-tuples (exception-backtrace pairs). The array comprehension on the next line expects a `Vector` of two-tuples, though, and consequently ends up attempting to index into exceptions.

Patch inbound :).

Sacha0:
Fixes #757, which contains a description of the issue, and strengthens associated tests. Best! :)

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1101,17 +1101,37 @@ if VERSION < v"1.7.0-DEV.1106"
 
     if VERSION >= v"1.1"
         function current_exceptions(task=current_task(); backtrace=true)
-            stack = Base.catch_stack(task, include_bt=backtrace)
-            ExceptionStack(Any[(exception=x[1],backtrace=x[2]) for x in stack])
+            old_stack = Base.catch_stack(task, include_bt=backtrace)
+            # If include_bt=true, Base.catch_stack yields a Vector of two-tuples,
+            # where the first element of each tuple is an exception and the second
+            # element is the corresponding backtrace. If instead include_bt=false,
+            # Base.catch_stack yields a Vector of exceptions.
+            #
+            # Independent of its backtrace keyword argument, Base.current_exceptions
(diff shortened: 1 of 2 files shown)


Sacha0:
Rebased over https://github.com/JuliaLang/Compat.jl/pull/758. What say you, CI? :)

c42f (approved):
This looks great :+1:

Sacha0:
Thanks for reviewing Chris! I hope you're doing well! :)

PR #763, opened 2022-01-18
`display` newline

t-bltg:
Per https://github.com/JuliaLang/julia/pull/43787 and https://github.com/JuliaLang/julia/pull/43787#issuecomment-1013898482.

@mkitti, since I'm unfamiliar with `Compat.jl`, I hope this is correct ?

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1314,6 +1314,16 @@ if VERSION < v"1.8.0-DEV.487"
     eachsplit(str::AbstractString; limit::Integer=0, keepempty=false) =
         eachsplit(str, isspace; limit=limit, keepempty=keepempty)
 end
+
+# https://github.com/JuliaLang/julia/pull/43787
+if VERSION < v"1.8.0-DEV.1310"
+    Base.display(d::TextDisplay, M::MIME"text/plain", @nospecialize x) = (show(d.io, M, x); println(d.io))
+    function Base.display(d::TextDisplay, M::MIME, @nospecialize x)
+        displayable(d, M) || throw(MethodError(display, (d, M, x)))
+        show(d.io, M, x); println(d.io)
+    end
+end
+
 include("iterators.jl")
 include("deprecated.jl")
 
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1317,8 +1317,9 @@ end
 
 # https://github.com/JuliaLang/julia/pull/43787
 if VERSION < v"1.8.0-DEV.1310"
-    Base.display(d::TextDisplay, M::MIME"text/plain", @nospecialize x) = (show(d.io, M, x); println(d.io))
-    function Base.display(d::TextDisplay, M::MIME, @nospecialize x)
+    display(d::TextDisplay, M::MIME"text/plain", @nospecialize x) = (show(d.io, M, x); println(d.io))
+    display(d::TextDisplay, @nospecialize x) = display(d, MIME"text/plain"(), x)
+    function display(d::TextDisplay, M::MIME, @nospecialize x)
         displayable(d, M) || throw(MethodError(display, (d, M, x)))
         show(d.io, M, x); println(d.io)
     end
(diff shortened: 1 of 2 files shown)


fredrikekre:
It would have to be a separate function (`Compat.display`) I believe.

mkitti:
Perhaps alternate `TextDisplay` might make sense rather than creating an alternate `display`?

t-bltg:
My implementation doesn't seem to work.
```
ERROR: MethodError: no method matching display(::TextDisplay, ::Int64)
You may have intended to import Base.display
```

mkitti:
I would implement a `CompatTextDisplay <: AbstractDisplay`. You could then `push!` that into `Base.Multimedia.displays` to get compatible behavior.

t-bltg:
> I would implement a `CompatTextDisplay <: AbstractDisplay`. You could then `push!` that into `Base.Multimedia.displays` to get compatible behavior.

@fredrikekre suggested that we shouldn't override the default behavior and use `Compat.display`. Pushing to `Base.Multimedia.displays` will affect the default display, no ?

mkitti:
The approaches are not mutually exclusive although the trick is in creating separate namespaces. You might need to create submodules to contain one or both.

The advantage of making a separate `CompatTextDisplay` is that you will be able to influence how `Base.display` works in code that you do not have access to. Perhaps you are using another module that uses `display`, and you do not want to have to vendor or modify that package directly.

Just to clear, `Compat.jl` should not necessarily `push!` `CompatTextDisplay` automatically on `__init__`. Rather this is something the end application developer might do.

src/Compat.jl
```diff
@@ -1314,6 +1314,17 @@ if VERSION < v"1.8.0-DEV.487"
     eachsplit(str::AbstractString; limit::Integer=0, keepempty=false) =
         eachsplit(str, isspace; limit=limit, keepempty=keepempty)
 end
+
+# https://github.com/JuliaLang/julia/pull/43787
+if VERSION < v"1.8.0-DEV.1310"
+    display(d::TextDisplay, M::MIME"text/plain", @nospecialize x) = (show(d.io, M, x); println(d.io))
```
mkitti, line 1320:
```suggestion
    display(d::TextDisplay, M::MIME"text/plain", @nospecialize x) = (show(d.io, M, x); println(d.io))
```

```suggestion
    import Base.Multimedia.display
    display(d::TextDisplay, M::MIME"text/plain", @nospecialize x) = (show(d.io, M, x); println(d.io))
```

mkitti:
I just read this more carefully, and I see that now that you are mainly trying to make the new behavior available in older Julia as opposed to making the old behavior available in new Julia.

src/Compat.jl
mkitti: That would be the change needed with the current approach. Without, this the current behavior of `Base.display` is not affected.

However, for a separate `Compat.display`, you would NOT want to do this.

t-bltg:
> I just read this more carefully, and I see that now that you are mainly trying to make the new behavior available in older Julia as opposed to making the old behavior available in new Julia.

Exactly, if I'm not misunderstanding the word `latest`, quoting from `README.md`:
> the Compat package provides a macro that lets you use the latest syntax in a backwards-compatible way.

src/Compat.jl
t-bltg: Why is `import Base.Multimedia.display` necessary here ?

t-bltg:
> The advantage of making a separate `CompatTextDisplay` is that you will be able to influence how `Base.display` works in code that you do not have access to. Perhaps you are using another module that uses `display`, and you do not want to have to vendor or modify that package directly.

That might be handy.

> Just to clear, `Compat.jl` should not necessarily `push!` `CompatTextDisplay` automatically on `__init__`. Rather this is something the end application developer might do.

I'd be happy to review a PR, my implementation is not necessarily the one we need, and I don't have a clear vision of what you propose as an alternative.

LilithHafner:
I don't think we need this anymore. IIUC breakage was all transient in scripts and the window for adding tooling with a significant impact on that has closed.

PR #765, opened 2022-02-16
Remove deprecations, weed out dependencies

### LINKED ISSUE #764: too many dependencies
As we all know, *many* packages depend on Compat. Unfortunately, Compat depends on every standard library package, which is a problem for static compilation via e.g. `filter_stdlibs` in PackageCompiler --- you can't use a simple compatibility shim function without bringing in the REPL, Pkg, etc. Is there any possible path to removing these dependencies? I see we have binding deprecations for them here, but I don't remember what they do exactly.

martinholters:
Closes  #764, which is the sole motivation for this PR at this point. Although I wouldn't expect anyone to be relying on those deprecations, it's technically breaking, so bumps the major version to 4.

Going forward, we might pick up new dependencies if we add compat code for other stdlibs. If we want to avoid that, we could think about introducing individual compat packages like `CompatDates`, `CompatLinearAlgebra` etc., But let's cross that bridge when we get there---if ever; the relatively small number of dependencies we have now should hopefully be ok.


Project.toml
```diff
@@ -1,31 +1,19 @@
...
-Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
-REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
-Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
-Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
-SharedArrays = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
-Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
-SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
JeffBezanson, line 9: Can this move to extras?

martinholters: Unfortunately no, due to
https://github.com/JuliaLang/Compat.jl/blob/a2de10749ab9d8155c3ae88a408e68f3497a2c33/src/Compat.jl#L489-L496

martinholters:
Before going ahead here, I'd like to get some feedback; bumping the major version shouldn't be done too easily.

KristofferC:
It seems like an accident that these are still here since the binding was removed in 1.0 and AFAIU this package dropped everything before that.

Also, since 1.6 is the LTS and we are doing a breaking version, I would suggest removing everything that has to do with the pre 1.6 era. This would for example allow us to drop the Test dependency as well.

martinholters:
>  I would suggest removing everything that has to do with the pre 1.6 era. This would for example allow us to drop the Test dependency as well.

I like that. Packages still supporting pre-1.6 julia could nevertheless use almost everything Compat supports now and at the same time declare compatibility with the new, trimmed down Compat version. Only things that were not exported by Compat and hence accessed with explicit qualification would be problematic. Still, we probably shouldn't rush this. But this PR has been sitting idle for more than a month, so nobody is rushing anything here :smile: 

I'll add a commit dropping anything pre-1.6 so we can see what it feels like.

martinholters:
Superseded by #767.

PR #766, opened 2022-02-27
Add `allequal`

mcabbott:
This adds https://github.com/JuliaLang/julia/pull/43354, which is in Julia 1.8

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.41.0"
+version = "3.42.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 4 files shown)


martinholters (approved):

mcabbott:
Bump, can this be merged?

mcabbott:
BTW, using this seems to cause https://github.com/JuliaDiff/ChainRules.jl/issues/598 . I've no idea why, though.

martinholters:
I have no idea what's going on, but https://github.com/JuliaDiff/ChainRules.jl/issues/598#issuecomment-1063935985 says:

> This wasn't an issue with Compat either, but with an older version of JuliaFormatter in my startup.jl!

And I failed to reproduce the problem. If someone can come up with a reproducer, I'm willing to look at it, but I doubt it's Compat's fault.

mcabbott:
Thanks for looking. I can't reproduce it either, but just wondered if you (or anyone) had ever seen something like this. I see nothing wrong with what either package is doing.

PR #768, opened 2022-04-11
Add support for property destructuring in `@compat` macro

kleinschmidt:
This currently supports only the basic `(; a, b) = something` syntax, and does not handle the parallel function argument syntax.

~~I didn't bump the package version because I'm not sure what the release policy is but I'm happy to do that too!~~  I bumped the minor version since this is a new feature and the last merged commit was a release so it feels pretty safe.

### DIFF
--- a/src/compatmacro.jl
+++ b/src/compatmacro.jl
@@ -16,7 +16,8 @@ function _compat(ex::Expr)
         end
     end
 
-    @static if VERSION < v"1.7.0"
+    # https://github.com/JuliaLang/julia/pull/39285
+    @static if VERSION < v"1.7.0-DEV.364"
         ex = _destructure_named_tuple(ex)
     end
     
--- a/src/compatmacro.jl
+++ b/src/compatmacro.jl
@@ -16,6 +16,10 @@ function _compat(ex::Expr)
         end
     end
 
+    @static if VERSION < v"1.7.0"
+        ex = _destructure_named_tuple(ex)
+    end
+    
     return Expr(ex.head, map(_compat, ex.args)...)
 end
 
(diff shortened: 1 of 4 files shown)


src/compatmacro.jl
```diff
@@ -16,6 +16,10 @@ function _compat(ex::Expr)
         end
     end
 
+    @static if VERSION < v"1.7.0"
```
ararslan, line 19:
```suggestion
    # https://github.com/JuliaLang/julia/pull/39285
    @static if VERSION < v"1.7.0-DEV.364"
```

kleinschmidt: just out of curiosity, how do you get the DEV version from that?

src/compatmacro.jl
```diff
@@ -94,3 +98,18 @@ function _assign_implicit_keywords(ex::Expr)
 
     return ex
 end
+
+function _destructure_named_tuple(ex::Expr)
+    if ex.head === :(=) && ex.args[1] isa Expr && ex.args[1].head === :tuple &&
+        ex.args[1].args[1] isa Expr && ex.args[1].args[1].head === :parameters
```
ararslan, line 104: Doesn't really matter but FWIW I would put the checking on whether this function is applicable as a guard at the call site like the other function called inside of `_compat`.

Also, you can simplify the `ex.stuff isa Expr && ex.head === :thing` to `Meta.isexpr(ex.stuff, :thing)`.

src/compatmacro.jl
ararslan: Scroll to where the PR is merged and click the link to the commit:
image
That will show you the full commit hash, in this case `d9e26327949d0ccc938b0e88c5c8a6adb36ddec5`. Then in the Julia repo, use
```bash
./contrib/commit-name.sh d9e26327949d0ccc938b0e88c5c8a6adb36ddec5
```

ararslan (approved):

PR #772, opened 2022-05-18
Expand usage section in README to mention backports

martinholters:
Ref https://github.com/JuliaLang/Compat.jl/pull/767#issuecomment-1129645422, cc @bkamins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,8 @@ such as `Compat = "3.22, 4"` in the `[compat]`section of the `Project.toml` file
 in your package directory. The version in the latter should be the minimum
 version that supports all needed features (see list below). Note that Compat v4
 requires Julia v1.6, but some features may have been backported to Compat v3
-(see the [feature list of the release-3 branch]
-(https://github.com/JuliaLang/Compat.jl/tree/release-3#supported-features)).
+(see the
+[feature list of the release-3 branch](https://github.com/JuliaLang/Compat.jl/tree/release-3#supported-features)).
 If you require any of those backported features, be sure to specify the correct
 compatibility in your `Project.toml`. E.g. if the feature from Compat v4.x has
 been backported to v3.y, use `Compat = 3.y, 4.x`. If you use a feature that had
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ compatibility with v4 with `Compat = 3.x, 4` (unless you use one the very few
 things that got removed between Compat v3 and v4, which you most probably
 don't).
 
+To minimize dependency conflicts between packages it is recommended that packages allow for both appropriate v4 and v3 versions of Compat.jl in their Project.toml (except for rare cases of packages that support only v4 or v3 version of Compat.jl).
+
 Then, in your package, shortly after the `module` statement include a line like
 this:
 
--- a/README.md
+++ b/README.md
@@ -22,11 +22,23 @@ To use Compat in your Julia package, add it as a dependency of your package usin
 ```julia
 pkg> add Compat
 ```
-and add a [version specifier line](https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1) such as `Compat = "2.2, 3"` in the `[compat]`section of the `Project.toml` file
+and add a [version specifier line](https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1)
+such as `Compat = "3.22, 4"` in the `[compat]`section of the `Project.toml` file
 in your package directory. The version in the latter should be the minimum
-version that supports all needed fatures (see list below), and (if applicable)
-any newer major versions verified to be compatible. Then, in your package,
-shortly after the `module` statement a line like this:
+version that supports all needed features (see list below). Note that Compat v4
+requires Julia v1.6, but some features may have been backported to Compat v3.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,11 +22,23 @@ To use Compat in your Julia package, add it as a dependency of your package usin
...
+So if compatibility with older Julia is important to you, also check the
+[feature list of the release-3 branch](https://github.com/JuliaLang/Compat.jl/tree/release-3#supported-features).
+If you require any of those backported features, be sure to specify the correct
+compatibility in your `Project.toml`. E.g. if the feature from Compat v4.x has
+been backported to v3.y, use `Compat = 3.y, 4.x`. If you use a feature that had
+originally been added in Compat v3 (e.g. in 3.x), don't forget to also declare
+compatibility with v4 with `Compat = 3.x, 4` (unless you use one the very few
+things that got removed between Compat v3 and v4, which you most probably
+don't).
+
```
bkamins, line 39: Maybe we can add something like:

```suggestion

To minimize dependency conflicts between packages it is recommended that packages allow for both appropriate v4 and v3 versions of Compat.jl in their Project.toml (except for rare cases of packages that support only v4 or v3 version of Compat.jl).

```

martinholters: I tried to imply that in my last two or so sentences. Stating it explicitly might indeed be better. But do you feel that would add unnecessary redundancy? I.e. should we just add this, or instead rephrase the second half of my original text entirely?

bkamins: I know you did this :). I just thought that adding a shorter explicit summary with a reason for the recommendation would be useful.

I do not have a problem with duplication. While I agree that for experienced package maintainers this information is not needed, I think, that in general in Julia package ecosystem we should start adding more "new user friendly" explanations along with precise technical explanations. I hope this is one of the means that, if we adopted it, would lead to resolution of issues like https://discourse.julialang.org/t/discussion-on-why-i-no-longer-recommend-julia-by-yuri-vishnevsky/81151 in long term.

@nalimilan - what do you think?

nalimilan: Yeah better be explicit about this. But maybe this information could indeed be added before saying "So if compatibility with older Julia is important".

martinholters: I'll go with the suggestion for now. We can always improve wording or paragraph organization in another iteration.

README.md
```diff
@@ -22,11 +22,25 @@ To use Compat in your Julia package, add it as a dependency of your package usin
...
+To minimize dependency conflicts between packages it is recommended that packages allow for both appropriate v4 and v3 versions of Compat.jl in their Project.toml (except for rare cases of packages that support only v4 or v3 version of Compat.jl).
```
nalimilan, line 40:
```suggestion
To minimize dependency conflicts between packages it is recommended that packages
allow for both appropriate v4 and v3 versions of Compat.jl in their Project.toml
(except for rare cases of packages that support only v4 or v3 version of Compat.jl).
```

README.md
```diff
@@ -22,11 +22,25 @@ To use Compat in your Julia package, add it as a dependency of your package usin
...
+and add a [version specifier line](https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1)
+such as `Compat = "3.22, 4"` in the `[compat]`section of the `Project.toml` file
 in your package directory. The version in the latter should be the minimum
-version that supports all needed fatures (see list below), and (if applicable)
-any newer major versions verified to be compatible. Then, in your package,
-shortly after the `module` statement a line like this:
+version that supports all needed features (see list below). Note that Compat v4
+requires Julia v1.6, but some features may have been backported to Compat v3.
+So if compatibility with older Julia is important to you, also check the
+[feature list of the release-3 branch](https://github.com/JuliaLang/Compat.jl/tree/release-3#supported-features).
```
nalimilan, line 31: This sentence is kind of contradictory with the next paragraph:
```suggestion
requires Julia v1.6, but some features may have been backported to Compat v3
(see the [feature list of the release-3 branch]
(https://github.com/JuliaLang/Compat.jl/tree/release-3#supported-features)).
```

martinholters: Yes, true.

README.md
```diff
@@ -22,11 +22,27 @@ To use Compat in your Julia package, add it as a dependency of your package usin
...
+and add a [version specifier line](https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1)
+such as `Compat = "3.22, 4"` in the `[compat]`section of the `Project.toml` file
 in your package directory. The version in the latter should be the minimum
-version that supports all needed fatures (see list below), and (if applicable)
-any newer major versions verified to be compatible. Then, in your package,
-shortly after the `module` statement a line like this:
+version that supports all needed features (see list below). Note that Compat v4
+requires Julia v1.6, but some features may have been backported to Compat v3
+(see the [feature list of the release-3 branch]
+(https://github.com/JuliaLang/Compat.jl/tree/release-3#supported-features)).
```
martinholters, line 31:
```suggestion
(see the
[feature list of the release-3 branch](https://github.com/JuliaLang/Compat.jl/tree/release-3#supported-features)).
```
Github markdown doesn't like the line break here, i.e. it doesn't render the link properly.

nalimilan (approved):

PR #777, opened 2022-08-21
`stack`

mcabbott:
From https://github.com/JuliaLang/julia/pull/43334


src/Compat.jl
```diff
@@ -362,6 +362,241 @@ end
...
+    With keyword `dims::Integer`, instead the `i`th element of `iter` becomes the slice
+    [`selectdim`](@ref)`(result, dims, i)`, so that `size(result, dims) == length(iter)`.
+    In this case `stack` reverses the action of [`eachslice`](@ref) with the same `dims`.
+
+    The various [`cat`](@ref) functions also combine arrays. However, these all
+    extend the arrays' existing (possibly trivial) dimensions, rather than placing
+    the arrays along new dimensions.
+    They also accept arrays as separate arguments, rather than a single collection.
+
+    !!! compat "Julia 1.9"
```
bkamins, line 389: is this comment needed in Compat.jl?

mcabbott: I've removed them, since they are pretty long.

ericphanson: I think Bogumił was referring to the compat annotation (which doesn't make as much sense bc you can use the function pre-1.9 if you are using Compat), not the whole docstring. IMO the docstring is still useful from Compat.

ararslan (approved):

src/Compat.jl
ararslan: I don't disagree that it's useful but I think most Compat definitions don't retain the docstring for some reason.

mcabbott: Ah ok. I can restore them, and edit the compat note to say when this is in Base julia? Perhaps that should be the policy.

ararslan (approved):

PR #782, opened 2022-09-29
add promoting ops for Dates.Period

### LINKED ISSUE #781: add `div(::Period, ::Period)` method that's present in 1.8
1.8.1 has a method for `div` on time periods which uses `promote`: 

https://github.com/JuliaLang/julia/blob/b73bf8618579d5ae0327917af590d823a100ec82/stdlib/Dates/src/periods.jl#L101

1.7.3 doesn't have this though (looks like it was added [this commit](https://github.com/JuliaLang/julia/commit/bdf9ead91e5a8dfd91643a17c1626032faada329) and included in 1.8.0).  We could add these methods in compat behind a version guard to add support for this.

kleinschmidt:
Fixes #781 

I'm not sure how to figure out exactly which 1.8-DEV.XXX is the right version guard for this; the relevant changes were introduced in this commit: https://github.com/JuliaLang/julia/commit/bdf9ead91e5a8dfd91643a17c1626032faada329

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "4.2.0"
+version = "4.3.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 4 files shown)


martinholters (commented):
LGTM, but needs an entry in the README.

Also, feel free to bump the version in Project.toml and we can tag a release immediately after merging.

PR #783, opened 2022-11-01
Assume effects

### LINKED ISSUE #769: compat for @assume_effects
introduced in https://github.com/JuliaLang/julia/pull/43852

Tokazama:
Fixes #769 

This allows packages with a 1.6 compat to replace `Base.@pure` with `Base.@assume_effects :total` similar to what's been done in base. All other effects are ignored

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -51,6 +51,21 @@ end
     end
 end
 
+# https://github.com/JuliaLang/julia/pull/43852
+if isdefined(Base, Symbol("@assume_effects"))
+    using Base: @assume_effects
+else
+    macro assume_effects(args...)
+        n = nfields(args)
+        call = getfield(args, n)
+        if n === 2 && getfield(args, 1) === QuoteNode(:total)
+            return esc(:(Base.@pure $(call)))
+        else
+            return esc(call)
+        end
+    end
+end
+
 if VERSION < v"1.7.0-DEV.119"
     # Part of:
     # https://github.com/JuliaLang/julia/pull/35316
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -52,7 +52,7 @@ end
 end
 
 # https://github.com/JuliaLang/julia/pull/43852
-if VERSION < v"1.8.0-DEV.1484"
+@static if VERSION < v"1.8.0-DEV.1484"
     using Base: @assume_effects
 else
     macro assume_effects(args...)
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -52,7 +52,7 @@ end
 end
 
 # https://github.com/JuliaLang/julia/pull/43852
-if isdefined(Base, Symbol("@assume_effects"))
+if VERSION < v"1.8.0-DEV.1484"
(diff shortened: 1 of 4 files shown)


Keno:
`@pure` is stronger than `@assume_effects`. I don't think this is a good idea.

Tokazama:
> `@pure` is stronger than `@assume_effects`. I don't think this is a good idea.

Sorry, I thought this was how all uses of `Base.@pure` were managed in Base. What's the appropriate way to transition code that uses `Base.@pure`?

Keno:
It should be generally safe to replace calls to `@pure`, by `@assume_effects :total`, but not the other way around. The issue is that with this, you'll write `@assume_effects :total`, but get `@pure` semantics on older versions of julia, which is not the same.

Keno:
More generally, though, packages should revisit every use of `@pure` and select the minimum set of effect assumptions that they need (preferably none).

Tokazama:
I remove the mapping to `@pure` so there aren't any added assumptions on 1.6.
I guess there could be an extended help section for `Base.@pure` detailing how to make the transition but a quick search on [JuliaHub](https://juliahub.com/ui/Search?q=%40pure&type=code) reveals that most packages that still use it are maintained by one particular individual. If I recall correctly, previous attempts to dissuade their use of `@pure` were met with quite a bit of resistance.

src/Compat.jl
```diff
@@ -51,6 +51,15 @@ end
     end
 end
 
+# https://github.com/JuliaLang/julia/pull/43852
+if isdefined(Base, Symbol("@assume_effects"))
```
martinholters, line 55:
```suggestion
if VERSION < v"1.8.0-DEV.1484"
```
Makes it easier in the future to figure out whether we can remove this when we bump the minimum Julia version.

src/Compat.jl
```diff
@@ -51,6 +51,15 @@ end
     end
 end
 
+# https://github.com/JuliaLang/julia/pull/43852
+if VERSION < v"1.8.0-DEV.1484"
```
martinholters, line 55:
```suggestion
@static if VERSION < v"1.8.0-DEV.1484"
```
should make this work even if `@assume_effects` is undefined.

martinholters: No, scratch that, it's just that the two branches of the `if` need to be swapped, right? Before `v"1.8.0-DEV.1484"`, we want to define our own macro, after we use the one from `Base`. Not the other way round.

src/Compat.jl
martinholters: Whoops, that reverses the logic though, sorry about that.

martinholters (approved):
LGTM, but we should get #784 in first and then test on top of that to verify it's working as intended on latest Julia.

That said, I wonder how helpful this is given it cannot be used as a replacement for `@pure` without regressions on older Julia. Still better than nothing, though.

@Keno any more thoughts on this?

martinholters:
Can you rebase on (or merge) current master to (hopefully) fix the test failures on nightly?

PR #785, opened 2022-11-17
add `Splat`

MasonProtter:
Ref: https://github.com/JuliaLang/julia/pull/42717

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -617,7 +617,7 @@ if VERSION < v"1.9.0-DEV.1163"
     _empty_stack(_...) = throw(ArgumentError("`stack` on an empty collection is not allowed"))
 end
                                 
-@static if !isdefined(Base, :Splat)
+@static if VERSION < v"1.9.0-DEV.513"
     # https://github.com/JuliaLang/julia/pull/42717
     export Splat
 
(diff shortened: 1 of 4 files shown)


src/Compat.jl
```diff
@@ -616,6 +616,44 @@ if VERSION < v"1.9.0-DEV.1163"
 
     _empty_stack(_...) = throw(ArgumentError("`stack` on an empty collection is not allowed"))
 end
+                                
+@static if !isdefined(Base, :Splat)
```
martinholters, line 620:
```suggestion
if VERSION < v"1.9.0-DEV.513"
```
To make it easier in the future to figure out at what point this can be removed.

MasonProtter: Okay, but I think it's better to do it with a `@static if`.

martinholters: And why is that?

martinholters (approved):

mcabbott:
After https://github.com/JuliaLang/julia/pull/48038 this should probably export some lower-case `splat` function. Probably its own `Compat.splat`, not the existing `Base.splat`.

martinholters:
Ah, this is quite unfortunate. Un-exporting `Splat` would technically be breaking. So I guess we should just keep `Splat` and and add `splat` and change the documentation to only mention `splat`?

> Probably its own `Compat.splat`, not the existing `Base.splat`.

And why is that?

PR #786, opened 2023-01-18
Add `splat`

martinholters:
Ref. #785 and  JuliaLang/julia#48038.

cc @MasonProtter and @mcabbott

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -578,8 +578,16 @@ end
     end
 end
 
-@testset "Splat" begin
-    @test Splat(+)((1,2,3)) == 6
-    @test repr(Splat(+)) == "Splat(+)"
-    @test repr(MIME"text/plain"(), Splat(+)) == "Splat(+)"
+@testset "splat" begin
+    @test splat(+)((1,2,3)) == 6
+
+    if v"1.10.0-" <= VERSION < v"1.10.0-DEV.360" ||
+        v"1.9.0-DEV.513" <= VERSION < v"1.9.0-beta3"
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -578,8 +578,16 @@ end
...
-    @test repr(Splat(+)) == "Splat(+)"
-    @test repr(MIME"text/plain"(), Splat(+)) == "Splat(+)"
+@testset "splat" begin
+    @test splat(+)((1,2,3)) == 6
+
+    if v"1.10.0-" <= VERSION < v"1.10.0-DEV.360" ||
+        v"1.9.0-DEV.513" <= VERSION < v"1.9.0-beta3"
+        # these versions of Base export Splat (which we use) but pretty-print with capital `S`
+        @test repr(splat(+)) == "Splat(+)"
+        @test repr(MIME"text/plain"(), splat(+)) == "Splat(+)"
```
martinholters, line 588: This is from between JuliaLang/julia#42717 and JuliaLang/julia#48038. Note that it only affects pre-release versions, so I decided not to worry about it too much and try to print lower-case everywhere, but instead just make the tests pass.

martinholters:
Should probably get #787 in first, to have a CI run with 1.9 on this one.

martinholters:
Ok, CI looks good (including on 1.9), so good to go from my side.

martinholters:
Our tests currently fail on Julia 1.9 and master because they test for `Splat`. With the development here being relatively slow, fixing it is not too urgent, but I'd nevertheless to do so soonish by merging this PR, unless anyone objects.

PR #787, opened 2023-01-18
Also run CI on Julia 1.9

martinholters:


martinholters:
Failure on nightly will be fixed by #786 (and is obviously not caused by this PR).

PR #791, opened 2023-02-16
Make `Dates` a weakdep

martinholters:
This is the little brother of #790.

It exploits that `Compat` only really needs `Dates` (to extend it) on Julia versions without weakdep support (i.e. before 1.9). So on newer Julia versions, no (unnecessary) dependence on `Dates` is introduced without loosing anything on older Julia versions.

Should a post-1.9 feature of `Dates` (or any other stdlib, actually) be added to `Compat` in the future, the way to do so is probably via a package extension (as proposed in #790), cc @oscardssmith.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,11 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 julia = "1.6"
 
 [extras]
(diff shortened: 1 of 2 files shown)


oscardssmith (approved):

oscardssmith:
sounds good.

PR #792, opened 2023-02-18
[release-3] Put uses of `@pure` and `@_pure_meta` behind `@static if`

martinholters:
Ref. https://github.com/JuliaLang/julia/pull/48682#issuecomment-1435515671. This makes Compat at least load again on Julia master.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.46.0"
+version = "3.46.1"
 
 [deps]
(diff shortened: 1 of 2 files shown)


martinholters:
Tests still fail on Julia nightly, but at least, loading works, so this should fix the immediate problems in the ecosystem.

Fixing the tests might still be worthwhile but is less urgent and therefore should happen separately.

maleadt (approved):
LGTM, confirmed that this works.

martinholters:
Test fix is at #794 (I hope). If that works, I'll merge that one first, then verify tests pass here as well, merge this and tag.

If any problems in #794 show up, I'll merge this as is and tag.

PR #793, opened 2023-02-18
[release-3] Also run CI on Julia 1.9 (backport of #787)

martinholters:
While the 3.x release still sees some development or at least maintenance, probably good to do thorough testing there as well.


martinholters:
Test failures on 1.9 should be fixed by #794, load errors on nightly by #792.

PR #796, opened 2023-03-21
Fix 2-arg `@inferred`

### LINKED ISSUE #795: 2-arg @inferred problem
I'm trying to use 2-arg `@inferred` macro from Compat 3.46.1, but I'm getting
```
ERROR: LoadError: LoadError: LoadError: UndefVarError: _args_and_call not defined
```
See [JuliaStats/Clustering.jl#248](https://github.com/JuliaStats/Clustering.jl/actions/runs/4475285895/jobs/7864558642?pr=248).

I guess that's because `_args_and_call()` is actually defined in *Test*, not *Base*.

alyst:
Fixes problems with 2-args `@inferred` when the inferred call has kwargs (#795).
The problem was that the kwarg path uses non-imported calls from *Test* and *InteractiveUtils*.
Now this is fixed, and the test is added.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "3.46.1"
+version = "3.46.2"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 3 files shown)


martinholters (approved):
LGTM, thanks. You could also bump the patch version in Project.toml so that we can tag a new release immediately after merging.

PR #799, opened 2023-06-21
Support `round`/`trunc`/`ceil`/`floor` for `Bool`

### LINKED ISSUE #798: Support `round`/`trunc`/`ceil`/`floor` for `Bool`
`round`/`trunc`/`ceil`/`floor` for `Bool` was introduced in Julia 1.8 (https://github.com/JuliaLang/julia/pull/25085), it would be nice to add support in `Compat` for that.

Related to https://github.com/oxinabox/LayeredLayouts.jl/pull/26.

mtfishman:
Closes #798.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -643,11 +643,11 @@ end
 
 # https://github.com/JuliaLang/julia/pull/25085
 if VERSION < v"1.8.0-beta2.17"
-  import Base: trunc, floor, ceil, round
-  trunc(::Type{Bool}, x::AbstractFloat) = (-1 < x < 2) ? 1 <= x : throw(InexactError(:trunc, Bool, x))
-  floor(::Type{Bool}, x::AbstractFloat) = (0 <= x < 2) ? 1 <= x : throw(InexactError(:floor, Bool, x))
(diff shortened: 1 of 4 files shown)


src/Compat.jl
```diff
@@ -641,6 +641,15 @@ if v"1.10.0-" <= VERSION < v"1.10.0-DEV.360" || VERSION < v"1.9.0-beta3"
     splat(f) = Splat(f)
 end
 
+# https://github.com/JuliaLang/julia/pull/25085
+if VERSION < v"1.8.0-beta2.17"
```
martinholters, line 645: Strictly speaking, this should probably be:
```suggestion
if VERSION < v"1.8.0-beta2.17" || v"1.9.0-" <= VERSION < v"1.9.0-DEV.94
```
I doubt that'll make any difference in practice, though.

martinholters (changes requested):
Basically lgtm, but needs an entry in the README. Also feel free to bump the version in Project.toml for an immediate release.

mtfishman:
@martinholters I added an entry to the README and bumped the version, good from my end.

martinholters (approved):

PR #804, opened 2023-07-31
`sort` for NTuple and other iterables

### LINKED ISSUE #803: `sort` for `NTuple` and other iterables
`sort` for `NTuple` and other iterables was added in https://github.com/JuliaLang/julia/pull/46104, it would be nice to add that to `Compat`.

mtfishman:
Closes #803.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -643,11 +643,10 @@ end
 
 # https://github.com/JuliaLang/julia/pull/25085
 if VERSION < v"1.8.0-beta2.17" || v"1.9.0-" <= VERSION < v"1.9.0-DEV.94"
-    import Base: trunc, floor, ceil, round
-    trunc(::Type{Bool}, x::AbstractFloat) = (-1 < x < 2) ? 1 <= x : throw(InexactError(:trunc, Bool, x))
-    floor(::Type{Bool}, x::AbstractFloat) = (0 <= x < 2) ? 1 <= x : throw(InexactError(:floor, Bool, x))
-    ceil(::Type{Bool}, x::AbstractFloat)  = (-1 < x <= 1) ? 0 < x : throw(InexactError(:ceil, Bool, x))
-    round(::Type{Bool}, x::AbstractFloat) = (-0.5 <= x < 1.5) ? 0.5 < x : throw(InexactError(:round, Bool, x))
+    Base.trunc(::Type{Bool}, x::AbstractFloat) = (-1 < x < 2) ? 1 <= x : throw(InexactError(:trunc, Bool, x))
+    Base.floor(::Type{Bool}, x::AbstractFloat) = (0 <= x < 2) ? 1 <= x : throw(InexactError(:floor, Bool, x))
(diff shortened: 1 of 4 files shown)


src/Compat.jl
```diff
@@ -706,6 +706,42 @@ if VERSION < v"1.7.0-DEV.1187"
     export redirect_stdio
 end
 
+# https://github.com/JuliaLang/julia/pull/46104
+if VERSION < v"1.10.0-DEV.1404"
+    import Base: sort, Ordering, Forward, ord, lt, tail, copymutable, DEFAULT_STABLE, IteratorSize, HasShape, IsInfinite
```
martinholters, line 711: Shouldn't import things not meant to be extended:
```suggestion
    import Base: sort
    using Base: Ordering, Forward, ord, lt, tail, copymutable, DEFAULT_STABLE, IteratorSize, HasShape, IsInfinite
```
or (and that would be my personal preference)
```suggestion
    using Base: Ordering, Forward, ord, lt, tail, copymutable, DEFAULT_STABLE, IteratorSize, HasShape, IsInfinite
```
and explicitly qualify `sort` below.

martinholters (commented):
LGTM except for some minor nit regarding `import` usage.

src/Compat.jl
```diff
@@ -706,6 +705,43 @@ if VERSION < v"1.7.0-DEV.1187"
...
+# https://github.com/JuliaLang/julia/pull/46104
+if VERSION < v"1.10.0-DEV.1404"
+    using Base: Ordering, Forward, ord, lt, tail, copymutable, DEFAULT_STABLE, IteratorSize, HasShape, IsInfinite
+    function Base.sort(v; kws...)
+        size = IteratorSize(v)
+        size == HasShape{0}() && throw(ArgumentError("$v cannot be sorted"))
+        size == IsInfinite() && throw(ArgumentError("infinite iterator $v cannot be sorted"))
+        sort!(copymutable(v); kws...)
+    end
+    sort(v::AbstractVector; kws...) = sort!(copymutable(v); kws...) # for method disambiguation
```
martinholters, line 717: I think this line just moved in https://github.com/JuliaLang/julia/issues/46104 and so there is no need to reproduce it here?
```suggestion
```
AFAICT, it is present in all relevant past versions:
https://github.com/JuliaLang/julia/blob/v1.6.0/base/sort.jl#L794
https://github.com/JuliaLang/julia/blob/v1.7.0/base/sort.jl#L770
https://github.com/JuliaLang/julia/blob/v1.8.0/base/sort.jl#L768
https://github.com/JuliaLang/julia/blob/v1.9.0/base/sort.jl#L1400

Is there a specific reason you added?

mtfishman: Ah right. I knew there was a reason I left it out to begin with.

martinholters: Doing the first review round, I had wondered why this line was missing until I found it had just moved...

ericphanson:
This should probably be reverted as it was reverted upstream: https://github.com/JuliaLang/julia/pull/52010

martinholters:
#811

PR #805, opened 2023-09-15
Add `public` to the `@compat` macro

LilithHafner:
The documentation included in this PR should make the PR itself self-explanatory. If it's not self-explanatory, then it's not acceptably documented.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -741,31 +741,6 @@ if VERSION < v"1.10.0-DEV.1404"
         (lt(o, y[1], x[1]) ? (y[1], merge(x, tail(y), o)...) : (x[1], merge(tail(x), y, o)...))
 end
 
-# https://github.com/JuliaLang/julia/pull/50105
-export @public
-"""
-    @public foo, bar, baz
-
-Equivalent to `public foo, bar, baz` on versions of Julia that support the `public` keyword;
-a no-op on versions of Julia that do not support the `public` keyword.
-"""
-macro public(arg)
-    symbols = _get_symbols(arg)
-    if VERSION >= v"1.11.0-DEV.469"
-        esc(Expr(:public, symbols...))
-    end
-end
-
-_get_symbols(symbol::Symbol) = (symbol,)
-function _get_symbols(symbols)
-    if symbols isa Expr && symbols.head == :tuple && all(x -> x isa Symbol, symbols.args)
-        symbols.args
-    else
-        throw(ArgumentError("cannot mark `$symbols` as public. Try `@public foo, bar, baz`."))
-    end
-end
-
-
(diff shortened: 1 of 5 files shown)


src/Compat.jl
```diff
@@ -741,6 +741,31 @@ if VERSION < v"1.10.0-DEV.1404"
...
+macro public(arg)
+    symbols = _get_symbols(arg)
+    if VERSION >= v"1.11.0-DEV.469"
+        esc(Expr(:public, symbols...))
+    end
+end
+
+_get_symbols(symbol::Symbol) = (symbol,)
+function _get_symbols(symbols)
+    if Base.isexpr(symbols, :tuple) && all(x -> x isa Symbol, symbols.args)
```
LilithHafner, line 761:
```suggestion
    if symbols isa Expr && symbols.head == :tuple && all(x -> x isa Symbol, symbols.args)
```
1.6 doesn't have `isexpr`

fredrikekre:
Can I propose `@compat public foo, bar`?

LilithHafner:
Is `@compat` already used or would this be the only non-deprecated use of the `@compat` macro?

fredrikekre:
It used to be used for all these kind of things in the past, and it is still there: https://github.com/JuliaLang/Compat.jl/blob/c59d116e7039db494ac61548a52651f3d8ba7b30/src/compatmacro.jl#L26

LilithHafner:
IIUC `@compat` was used for syntax changes in the pre-1.0 erra, and was rendered a no-op when Compat.jl dropped support for pre-1.0 code? If that history is correct, then `@compat public foo, bar` seems appropriate.

fredrikekre:
But that was because Compat.jl dropped support for pre Julia 1.0, no? Then there was no point in having all the old syntax transformations.

LilithHafner:
@fredrikekre, I think we are in agreement. I also prefer your proposed syntax. We should probably wait at least a day or so to let other folks chime in, though.

mcabbott:
This does not seem to allow macros:
```
julia> @compat public fun, @mac
ERROR: LoadError: ArgumentError: cannot mark `(fun, #= REPL[15]:1 =# @mac())` as public. Try `@compat public foo, bar`.
```

src/Compat.jl
mcabbott: It's `Meta.isexpr`, btw.

LilithHafner:
Should this be backported to 3.x?

lgoettgens:
> Should this be backported to 3.x?

I would be in favor of this. There are many packages out there that support julia versions older than 1.6 and are thus unable to depend on Compat 4. 
Personally, I would really like to use `@public` via a Compat-compatible way in Aqua.

PR #779, opened 2022-08-21
Backport 3.x: `stack` (Julia PR 43334)

ararslan:
Backport of #777 to the `release-3` branch.

Closes #778

PR #806, opened 2023-10-02
Update version comment of keepat!

jonas-schulze:
Uses same wording as in #773


martinholters:
Same for `stack` (#779) then, too?

PR #807, opened 2023-10-05
pkgversion

palday:


src/Compat.jl
```diff
@@ -385,6 +385,58 @@ end
...
 
+# this function is available as of Julia 1.9
+# https://github.com/JuliaLang/julia/pull/45607
+@static if !isdefined(Base, :pkgversion)
+    using TOML: parsefile
+    export pkgversion
+
+    const project_names = ("JuliaProject.toml", "Project.toml")
+    const manifest_names = ("JuliaManifest.toml", "Manifest.toml")
+    const preferences_names = ("JuliaLocalPreferences.toml", "LocalPreferences.toml")
```
martinholters, line 396: Are these used anywhere?
```suggestion
```

martinholters (commented):
In terms of docs/tests/version bump, this is all fine, but I have no idea whether the logic implemented here is ok. The tests pass, so it's probably good, but if anyone knowing about this stuff could take a brief look, that would be welcome.

src/Compat.jl
```diff
@@ -385,6 +385,58 @@ end
...
+
+    function locate_project_file(env::String)
+        for proj in project_names
+            project_file = joinpath(env, proj)
+            if Base.isfile_casesensitive(project_file)
+                return project_file
+            end
+        end
+        return true
```
martinholters, line 405:
```suggestion
        return nothing
```
Would be more idiomatic, or is there a good reason for return `true` if the project file could not be found?

src/Compat.jl
palday: Ah, no, this is copy-pasta from the implementation in Base

src/Compat.jl
palday: IIRC I grabbed this from the Base implementation, but this isn't part of the public API, so I'm fine with `nothing`

src/Compat.jl
```diff
@@ -385,6 +385,56 @@ end
     end
 end
 
+# this function is available as of Julia 1.9
+# https://github.com/JuliaLang/julia/pull/45607
```
omus, line 389:
```suggestion
# https://github.com/JuliaLang/julia/pull/45607
# https://github.com/JuliaLang/julia/pull/45695
```

src/Compat.jl
```diff
@@ -385,6 +385,56 @@ end
...
+    load.
+
+    To get the version of the package that imported the current module
+    the form `pkgversion(@__MODULE__)` can be used.
+    """
+    function pkgversion(m::Module)
+        path = pkgdir(m)
+        isnothing(path) && return nothing
+        return get_pkgversion_from_path(path)
+    end
```
omus, line 435: Should probably incorporate the lock changes from https://github.com/JuliaLang/julia/pull/45861 which call out https://github.com/JuliaLang/julia/pull/45607 in that it added memory corruption.

omus: Possibly also the changes from: https://github.com/JuliaLang/julia/pull/46738

src/Compat.jl
```diff
@@ -385,6 +385,56 @@ end
     end
 end
 
+# this function is available as of Julia 1.9
+# https://github.com/JuliaLang/julia/pull/45607
+@static if !isdefined(Base, :pkgversion)
+    using TOML: parsefile
+    export pkgversion
+
+    const project_names = ("JuliaProject.toml", "Project.toml")
```
omus, line 394:
```suggestion
    const require_lock = isdefined(Base, :require_lock) ? Base.require_lock : ReentrantLock()
    const project_names = ("JuliaProject.toml", "Project.toml")
```

src/Compat.jl
```diff
@@ -385,6 +385,56 @@ end
...
+    The version is read from the package's Project.toml during package
+    load.
+
+    To get the version of the package that imported the current module
+    the form `pkgversion(@__MODULE__)` can be used.
+    """
+    function pkgversion(m::Module)
+        path = pkgdir(m)
+        isnothing(path) && return nothing
+        return get_pkgversion_from_path(path)
```
omus, line 434:
```suggestion
        path === nothing && return nothing
        Base.@lock require_lock begin
            v = get_pkgversion_from_path(path)
            pkgorigin = get(Base.pkgorigins, Base.PkgId(Base.moduleroot(m)), nothing)
            # Cache the version
            if pkgorigin !== nothing && pkgorigin.version === nothing
                pkgorigin.version = v
            end
            return v
        end
```

omus: Just noticed that Julia 1.6 `Base.PkgOrigin` doesn't have the `version` field. Will update

omus: Suggestion has been updated to include the `hasfield(Base.PkgOrigin, :version)` check

src/Compat.jl
omus: Reference to Base returning `true`: https://github.com/JuliaLang/julia/blob/0f7b598ad2a690cae4fe67b3f834e476a77399f4/base/loading.jl#L584. I also think it's fine either way.

omus (approved):

README.md
```diff
@@ -118,6 +118,8 @@ changes in `julia`.
 
 * `@something` and `@coalesce` as short-circuiting versions of `something` and `coalesce` ([#40729]) (since Compat 3.29)
 
+* `pkgversion` Return the version of the package that imported module ([#45607]) (since Compat 4.11)
```
omus, line 121: Tweaking the README before merging.
```suggestion
* `pkgversion(m::Module)` returns the version of the package that loaded a given module ([#45607]) (since Compat 4.11)
```

omus:
Will merge after CI completes again

omus:
Not going to bother waiting on the Windows Julia 1.6 job as it passed on the previous commit and this last change was just documentation.

PR #808, opened 2023-11-17
Backport `public` support to release-3

LilithHafner:
cc @lgoettgens

### DIFF
--- a/src/compatmacro.jl
+++ b/src/compatmacro.jl
@@ -118,6 +118,7 @@ function _destructure_named_tuple(ex::Expr)
 end
 
 # https://github.com/JuliaLang/julia/pull/50105
+# No ambiguity with import because this requires the first argument to be a symbol.
 macro compat(public::Symbol, symbols_expr::Union{Expr, Symbol})
     public == :public || throw(ArgumentError("Invalid Syntax: `@compat $public $symbols_expr`"))
     symbols = _get_symbols(symbols_expr)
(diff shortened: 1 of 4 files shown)


lgoettgens:
The failure in 1.2 and older seems to be genuine (see https://github.com/JuliaLang/Compat.jl/actions/runs/6909455727/job/18800838907?pr=808#step:6:172).

src/compatmacro.jl
```diff
@@ -116,3 +116,38 @@ function _destructure_named_tuple(ex::Expr)
     push!(ex.args, values)
     return ex
 end
+
+# https://github.com/JuliaLang/julia/pull/50105
+macro compat(public::Symbol, symbols_expr::Union{Expr, Symbol})
+    public == :public || throw(ArgumentError("Invalid Syntax: `@compat $public $symbols_expr`"))
+    symbols = _get_symbols(symbols_expr)
+    if VERSION >= v"1.11.0-DEV.469"
```
lgoettgens, line 124:
```suggestion
    if VERSION >= v"1.11.0-DEV.469"
        symbols = _get_symbols(symbols_expr)
```
This should fix it, as the old julia versions no longer traverse the faulty path via `_get_symbols`.

LilithHafner: I want to keep the parsing outside the version check so that folks get errors consistently on all versions.
With that change, `@compat 1+1 == 2` would expand to `nothing` on old versions of Julia and throw on new versions of Julia.

I've tried to implement a fix that actually supports `@var_str` while still throwing on invalid inputs on all versions.

lgoettgens: Great! That sounds like a good idea :)

lgoettgens:
Gentle bump

PR #809, opened 2023-11-17
Mention the availability of `public` in compat 3.47

LilithHafner:
Depends on #808

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ changes in `julia`.
 
 ## Supported features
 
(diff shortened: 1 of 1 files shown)


lgoettgens:
The mentioned release somehow doesn't show up in the list of releases.

martinholters:
From the TagBot job:
```
git@github.com: Permission denied (publickey).
```
Maybe some repo admin could check whether the `DOCUMENTER_KEY` secret is set up correctly and matches a key configured for this repo? (The name is a bit misleading, maybe we want to change it to `TAGBOT_KEY` while at it and adjust `TagBot.yml` accordingly?)

KristofferC:
There is a `DOCUMENTER_KEY` key setup at least. Reading through https://github.com/JuliaRegistries/TagBot#setup it seems we should have everything that is needed..

PR #810, opened 2023-11-27
Fix `current_exceptions` test on nightly

martinholters:
The printing of the `UndefVarError` now includes the originating module.


martinholters:
Tests on nightly still fail, but now due to JuliaLang/julia#52010, ref https://github.com/JuliaLang/Compat.jl/pull/804#issuecomment-1823613510.

PR #811, opened 2023-11-27
Deprecate `sort` for `NTuple` and other iterables

martinholters:
As this functionality was removed from Julia in https://github.com/JuliaLang/julia/pull/52010.

CC @ericphanson

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -15,3 +15,48 @@ Base.@deprecate_binding parseatom Meta.parseatom false
 Base.@deprecate_binding parseall Meta.parseall false
 import UUIDs
 Base.@deprecate_binding uuid5 UUIDs.uuid5 true
+
+# https://github.com/JuliaLang/julia/pull/46104
+# reverted in https://github.com/JuliaLang/julia/pull/52010
+if VERSION < v"1.10.0-DEV.1404" ||
+    (VERSION >= v"1.10-rc2" && VERSION < v"1.11.0-") ||
+    VERSION >= v"1.11.0-DEV.924"
+    using Base: Ordering, Forward, ord, lt, tail, copymutable, DEFAULT_STABLE, IteratorSize, HasShape, IsInfinite
+    function Base.sort(v; kws...)
+        Base.depwarn("sorting arbitrary iterables is deprecated", :sort)
+        if v isa AbstractString
+            throw(ArgumentError("sort(::AbstractString) is not supported"))
+        end
+        if v isa NTuple
+            return _sort(v; kws...)
+        end
+        if v isa Tuple
+            throw(ArgumentError("sort(::Tuple) is only supported for NTuples"))
+        end
+        size = IteratorSize(v)
+        size == HasShape{0}() && throw(ArgumentError("$v cannot be sorted"))
+        size == IsInfinite() && throw(ArgumentError("infinite iterator $v cannot be sorted"))
+        sort!(copymutable(v); kws...)
+    end
+
+    function _sort(x::NTuple{N}; lt::Function=isless, by::Function=identity,
+                  rev::Union{Bool,Nothing}=nothing, order::Ordering=Forward) where N
+        o = ord(lt,by,rev,order)
+        if N > 9
+            v = sort!(copymutable(x), DEFAULT_STABLE, o)
+            tuple((v[i] for i in 1:N)...)
+        else
+            _sort(x, o)
+        end
+    end
+    _sort(x::Union{NTuple{0}, NTuple{1}}, o::Ordering) = x
+    function _sort(x::NTuple, o::Ordering)
+        a, b = Base.IteratorsMD.split(x, Val(length(x)>>1))
+        merge(_sort(a, o), _sort(b, o), o)
+    end
+    merge(x::NTuple, y::NTuple{0}, o::Ordering) = x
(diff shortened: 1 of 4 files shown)


martinholters:
This is on top of #810 to see whether tests now pass on nightly.

src/deprecated.jl
```diff
@@ -15,3 +15,48 @@ Base.@deprecate_binding parseatom Meta.parseatom false
 Base.@deprecate_binding parseall Meta.parseall false
 import UUIDs
 Base.@deprecate_binding uuid5 UUIDs.uuid5 true
+
+# https://github.com/JuliaLang/julia/pull/46104
+# reverted in https://github.com/JuliaLang/julia/pull/52010
+if VERSION < v"1.10.0-DEV.1404" ||
+    (VERSION >= v"1.10-rc2" && VERSION < v"1.11.0-") ||
```
martinholters, line 22: This assumes that JuliaLang/julia#52010 will actually be in v1.10-rc2.

src/deprecated.jl
```diff
@@ -15,3 +15,48 @@ Base.@deprecate_binding parseatom Meta.parseatom false
...
+        Base.depwarn("sorting arbitrary iterables is deprecated", :sort)
+        if v isa AbstractString
+            throw(ArgumentError("sort(::AbstractString) is not supported"))
+        end
+        if v isa NTuple
+            return _sort(v; kws...)
+        end
+        if v isa Tuple
+            throw(ArgumentError("sort(::Tuple) is only supported for NTuples"))
+        end
```
martinholters, line 35: I've changed the dispatch to only add this single method to `Base.sort`. If part of this functionality is re-added to`Base` in the future, it will likely be for more specific argument types and we then won't overwrite these new methods. Adding this method `Base` for any future Julia versions still makes me a bit uncomfortable.

martinholters:
The `sort` tests still pass (except for the `@allocated` ones), but of course are quite noise with many depwarns now. Should we remove them?

KristofferC:
No, this has to be reverted. We cannot have this level of type piracy in `Compat` just because it was added on master for a short while.

martinholters:
> No, this has to be reverted. We cannot have this level of type piracy in `Compat` just because it existed on Julia master for a while.
 
Yeah, that is super unfortunate, but if we just revert this, we have to do a new major version if we take SemVer seriously (which I like to do). And new major versions of Compat are not welcome by everyone, ref. https://github.com/JuliaLang/Compat.jl/pull/767#issuecomment-1127423282. A middle ground might be to keep this (deprecated) in Compat but only for Julia < v1.10. That would mean that for `sort` on general iterables, you'd need Julia < 1.10. Not sure that's much better, but at least, Compat doesn't pollute Base forever then.

> I mean, Compat should not even overload things in Base. You are supposed to have to call `Compat.sort` if you want to opt in to the "future" behavior of sort.

Hm, we've traditionally added methods to functions in Base whenever possible, reasoning that we're only doing it for past Julia versions where we know it's safe. And it's convenient from a user perspective. But I agree that always requiring the `Compat.` prefix would be much safer. We might want to adopt that in the future. But it doesn't help us with the mess we're in here now.

src/deprecated.jl
```diff
@@ -15,3 +15,48 @@ Base.@deprecate_binding parseatom Meta.parseatom false
 Base.@deprecate_binding parseall Meta.parseall false
 import UUIDs
 Base.@deprecate_binding uuid5 UUIDs.uuid5 true
+
+# https://github.com/JuliaLang/julia/pull/46104
+# reverted in https://github.com/JuliaLang/julia/pull/52010
+if VERSION < v"1.10.0-DEV.1404" ||
+    (VERSION >= v"1.10-rc2" && VERSION < v"1.11.0-") ||
+    VERSION >= v"1.11.0-DEV.924"
```
martinholters, line 23:
```suggestion
if VERSION < v"1.10.0-DEV.1404"
```
To stop adding methods to `Base.sort` in future Julia versions while keeping `sort` working on general iterables on Julia < 1.10 with Compat (where people might already be using it).

Ref. https://github.com/JuliaLang/Compat.jl/pull/811#issuecomment-1831397828

KristofferC:
Can we at least run PkgEval on this change and see if it breaks something in practice and if not we just remove it? Having this deprecated (but still there) in all of < 1.10 is awful. It shouldn't even be there in the first place since this is a feature that was introduced in 1.10 and should not automatically be available on pre 1.10. We don't backport features for a reason. 

The type of development where things that are tested on Julia master quickly get put into Compat in a type piracy way and then "cannot be removed due to semver" is completely unworkable. 

FWIW, I also think this change broke a package (https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/46617e5_vs_0ba6ec2/ITensorVisualizationBase.primary.log) where, ironically, their own type piracy (https://github.com/ITensor/NDTensors.jl/blob/19c079c554b546c57b3f5bd72db15c39112ff5d0/src/tupletools.jl#L162) stopped working.

So at least we have some evidence that the added type piracy here broke a package so based on semver (which we take seriously) it should be reverted.

martinholters:
Yeah, these are good arguments for reverting.

> Can we at least run PkgEval on this change 

Can nanosoldier to this for us or would someone need to run it on their own box?

Meanwhile, searching for `Compat = "4.9"` only found ` NDTensors`, so that's a likely candidate for being broken -- again.

Issue #814, opened 2023-11-29
Avoid type piracy by defining distinct functions


martinholters:
As argued by @KristofferC ion https://github.com/JuliaLang/Compat.jl/pull/811#issuecomment-1829627924 (and I tend to agree), Compat should not commit type piracy by adding methods to Base functions for Base types, but rather define distinct functions of the same name, so that usage needs to be explicit opt-in (by qualification or explicitly named `using`).

This is different from how we used to do it, by I think it's worthwhile. 

If we agree to do it this way, we should start by applying this rule to any new features. I'm not sure we should also go to the trouble of deprecating the old style in favor of the new style for existing features. Anyway, this likely warrants some discussion before setting anything in motion.

martinholters:
One issue here is that getting rid of old stuff becomes slightly more annoying. Let me explain with the very simple
https://github.com/JuliaLang/Compat.jl/blob/8819abe252579b610cb48c5ad83d51b45a90ddba/src/Compat.jl#L801-L803
Wtih the style proposed here, this would become
```julia
if VERSION < v"1.9.0-DEV.461"
    VersionNumber(args...) = Base.VersionNumber(args...)
    VersionNumber(v::VersionNumber) = v
else
    const VersionNumber = Base.VersionNumber
end
```
And users then do `Compat.VersionNumber(x)` whenever `x` is (or at least could be) a `VersionNumber`. All good. But then, eventually, we drop support for Julia prior to 1.10 in Compat v5.0.0, say. Then what? Keep `const VersionNumber = Base.VersionNumber` indefinitely? Deprecate it to `Base.VersionNumber` and remove it in v6? Keep it for v5, deprecate in v6, remove in v7?

We would like people to declare compatibility with v5 asap, as it will be much more light-weight after getting rid of the old stuff, but for packages still supporting older Julia, they would also keep compatibility with v4 and keep using `Compat.VersionNumber`, because they have to.

For those functions which required the `Compat.` before, I think we went with the deprecate in v5, remove in v6 scheme. But those were only few. If this now concerns every piece of Compat, the keep in v5 as no-op, deprecate in v6, remove in v7 might be in order.

Opinions?

PR #818, opened 2024-02-12
`allequal(f, itr)` and `allunique(f, itr)`

mcabbott:
Adds https://github.com/JuliaLang/julia/pull/47679

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -739,7 +739,7 @@ end
     @test allunique(abs2, Vector{Any}(1:100))
     # These cases don't call the function at all:
     @test allunique(error, [])
-    @test_skip allunique(error, [1])
+    @test_skip allunique(error, [1]) # depends on updated code in Base to work
 end
 @testset "allequal(f, xs)" begin
     @test allequal(abs2, [3, -3])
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -729,6 +729,34 @@ end
...
+@testset "allunique(f, xs)" begin
+    @test allunique(sin, 1:3)
+    @test !allunique(sin, [1,2,3,1])
+    @test allunique(sin, (1, 2, pi, im))  # eltype Any
+    @test allunique(abs2, 1:100)
+    @test !allunique(abs, -10:10)
+    @test allunique(abs2, Vector{Any}(1:100))
+    # These cases don't call the function at all:
+    @test allunique(error, [])
+    @test_skip allunique(error, [1])
```
martinholters, line 742:
```suggestion
    @test_skip allunique(error, [1]) # depends on updated code in Base to work
```

martinholters (approved):
The `@test_skip`s are a bit unfortunate, but the documentation in Base allows this ("The precise number of calls is regarded as an implementation detail."), so ok.

But this reminds me of #814. We would be better off if we had defined unexported `Compat.allequal` and `Compat.allunique`. Then we could just keep rolling our own here, regardless of Julia version.

mcabbott:
Yea maybe. I'd need another coffee to think about what happens on the Julia versions where this package stops exporting `allequal` but Base doesn't yet provide `allequal(f, itr)`. If this package had never exported it, that would indeed be simpler.

PR #821, opened 2024-02-17
logrange

mcabbott:
This adds https://github.com/JuliaLang/julia/pull/39071 

Note that:
* ~~The version check may need to change once backported to 1.11 branch~~
* Before Julia 1.9, it won't use TwicePrecision, as the necessary log/exp functions aren't defined. The result is slightly worse floating point accuracy. 
* It doesn't print the same way, since there's a big messy `print_range` function (for LinRange) whose signature the PR widened. 
* ~~The docstrings don't show up in REPL help, I don't know why. Maybe that's true of all functions here. It means you can't see the notes about the above points.~~

<details>

```
julia> logrange(1, 4, length=5)
5-element Compat.LogRange{Float64, Float64}:
 1.0
 1.414213562373095
 2.0
 2.82842712474619
 4.0

help?> logrange
search: logrange

  No documentation found.

  Compat.logrange is a Function.

  # 2 methods for generic function "logrange":
  [1] logrange(start::Real, stop::Real; length) in Compat at /Users/me/.julia/dev/Compat/src/Compat.jl:832
  [2] logrange(start::Real, stop::Real, length::Integer) in Compat at /Users/me/.julia/dev/Compat/src/Compat.jl:831

help?> Compat.LogRange
  No documentation found.

  Compat.LogRange is of type UnionAll.

  Summary
  ≡≡≡≡≡≡≡≡≡

  struct UnionAll <: Type{T}

julia> VERSION
v"1.6.0"
```

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ changes in `julia`.
 
 * `allequal(f, itr)` and `allunique(f, itr)` methods. ([#47679]) (since Compat 4.13.0)
 
-* `logrange(lo, hi; length)` is like `range` but for `*` not `+`. ([#39071]) (since Compat 4.14.0)
+* `logrange(lo, hi; length)` is like `range` but with a constant ratio, not difference. ([#39071]) (since Compat 4.14.0) Note that on Julia 1.8 and earlier, the version from Compat will has slightly lower floating-point accuracy than the one in Base (Julia 1.11 and later).
 
 * `Iterators.cycle(itr, n)` is the lazy version of `repeat(vector, n)`. ([#47354]) (since Compat 4.13.0)
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ changes in `julia`.
 
 * `allequal(f, itr)` and `allunique(f, itr)` methods. ([#47679]) (since Compat 4.13.0)
 
-* `logrange(lo, hi; length)` is like `range` but for `*` not `+`. ([#39071]) (since Compat 4.13.0)
+* `logrange(lo, hi; length)` is like `range` but for `*` not `+`. ([#39071]) (since Compat 4.14.0)
 
 * `Iterators.cycle(itr, n)` is the lazy version of `repeat(vector, n)`. ([#47354]) (since Compat 4.13.0)
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ changes in `julia`.
 
 * `allequal(f, itr)` and `allunique(f, itr)` methods. ([#47679]) (since Compat 4.13.0)
 
-* `logrange(lo, hi; length)` is like `range` but with a constant ratio, not difference. ([#39071]) (since Compat 4.14.0) Note that on Julia 1.8 and earlier, the version from Compat will has slightly lower floating-point accuracy than the one in Base (Julia 1.11 and later).
+* `logrange(lo, hi; length)` is like `range` but with a constant ratio, not difference. ([#39071]) (since Compat 4.14.0) Note that on Julia 1.8 and earlier, the version from Compat has slightly lower floating-point accuracy than the one in Base (Julia 1.11 and later).
 
 * `Iterators.cycle(itr, n)` is the lazy version of `repeat(vector, n)`. ([#47354]) (since Compat 4.13.0)
 
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@ changes in `julia`.
 
 ## Supported features
 
+* `logrange(lo, hi; length)` is like `range` but for `*` not `+`. ([#39071]) (since Compat 4.13.0)
+
 * `Iterators.cycle(itr, n)` is the lazy version of `repeat(vector, n)`. ([#47354]) (since Compat 4.13.0)
 
 * `@compat public foo, bar` marks `foo` and `bar` as public in Julia 1.11+ and is a no-op in Julia 1.10 and earlier. ([#50105]) (since Compat 3.47.0, 4.10.0)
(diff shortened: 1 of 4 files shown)


LilithHafner:
I don't have the energy/enthusiasm to review this PR. Feel free to proceed without me.

src/Compat.jl
```diff
@@ -803,6 +803,224 @@ if VERSION < v"1.11.0-DEV.1579"
     Iterators.cycle(xs, n::Integer) = Iterators.flatten(Iterators.repeated(xs, n))
 end
 
+# https://github.com/JuliaLang/julia/pull/39071
+if VERSION < v"1.12.0-DEV.2"  # might become wrong! PR will be backported to 1.11
```
ararslan, line 807: Seems safer in this case to use `isdefined`:
```suggestion
if !isdefined(Base, :logrange)  # VERSION < v"1.12.0-DEV.2" or appropriate 1.11.x after backporting
```

src/Compat.jl
```diff
@@ -803,6 +803,224 @@ if VERSION < v"1.11.0-DEV.1579"
...
+    # function Base.show(io::IO, ::MIME"text/plain", r::LogRange)
+    #     isempty(r) && return show(io, r)
+    #     summary(io, r)
+    #     println(io, ":")
+    #     print_range(io, r, " ", ", ", "", " \u2026 ")
+    # end
+
+    _exp_allowing_twice64(x::Number) = exp(x)
+
+    if VERSION >= v"1.9"  # then Base.Math._log_ext exists, allowing this high-precision path:
```
ararslan, line 997: Could use `isdefined` here as well
```suggestion
    if isdefined(Base.Math, :_log_ext)  # allows this high-precision path:
```

src/Compat.jl
mcabbott: Oh right, that's better.

README.md
```diff
@@ -72,6 +72,8 @@ changes in `julia`.
 
 * `allequal(f, itr)` and `allunique(f, itr)` methods. ([#47679]) (since Compat 4.13.0)
 
+* `logrange(lo, hi; length)` is like `range` but for `*` not `+`. ([#39071]) (since Compat 4.13.0)
```
mcabbott, line 75: Version:
```suggestion
* `logrange(lo, hi; length)` is like `range` but for `*` not `+`. ([#39071]) (since Compat 4.14.0)
```

src/Compat.jl
```diff
@@ -803,6 +803,224 @@ if VERSION < v"1.11.0-DEV.1579"
...
+    # function Base.show(io::IO, ::MIME"text/plain", r::LogRange)
+    #     isempty(r) && return show(io, r)
+    #     summary(io, r)
+    #     println(io, ":")
+    #     print_range(io, r, " ", ", ", "", " \u2026 ")
+    # end
+
+    _exp_allowing_twice64(x::Number) = exp(x)
+
+    if isdefined(Base.Math, :_log_ext)  # allows this high-precision path:
```
mcabbott, line 997: At present tests fail on 1.8, as the function exists but with different methods:
```suggestion
    if isdefined(Base.Math, :_log_ext) && hasmethod(Base.Math._log_ext, Tuple{UInt64})  # allows this high-precision path, Julia 1.9 & 1.10:
```
Alternatively, could go back to a version check:
```suggestion
    if VERSION >= v"1.9"  # allows this high-precision path:
```

ararslan: FYI it looks like the precise version that made the change is 1.9.0-DEV.318

src/Compat.jl
```diff
@@ -803,6 +803,224 @@ if VERSION < v"1.11.0-DEV.1579"
...
+    # function Base.show(io::IO, ::MIME"text/plain", r::LogRange)
+    #     isempty(r) && return show(io, r)
+    #     summary(io, r)
+    #     println(io, ":")
+    #     print_range(io, r, " ", ", ", "", " \u2026 ")
+    # end
+
+    _exp_allowing_twice64(x::Number) = exp(x)
+
+    if VERSION >= v"1.9"  # allows this high-precision path:
```
ararslan, line 997: For maximal pedantry
```suggestion
    if VERSION >= v"1.9.0-DEV.318"  # Julia PR #44717 allows this high-precision path:
```

mcabbott: Sure, why not! Thanks for checking.

martinholters:
> The docstrings don't show up in REPL help, I don't know why. Maybe that's true of all functions here. It means you can't see the notes about the above points.

I think that's because they are inside the `if`, not at top level, so need an eplicit `@doc`.

src/Compat.jl
```diff
@@ -803,6 +803,224 @@ if VERSION < v"1.11.0-DEV.1579"
...
+
+    julia> logrange(1, 1000, length=4) ≈ 10 .^ (0:3)
+    true
+    ```
+
+    See the [`Compat.LogRange`](@ref Compat.LogRange) type for further details.
+
+    !!! compat "Julia 1.9"
+        The version of this struct in Compat.jl does not use `Base.TwicePrecision{Float64}`
+        before Julia 1.9, so it sometimes has larger floating-point errors on intermediate points.
```
martinholters, line 848: Should this be mentioned in the README item? It's easy to miss here...

README.md
```diff
@@ -72,7 +72,7 @@ changes in `julia`.
 
 * `allequal(f, itr)` and `allunique(f, itr)` methods. ([#47679]) (since Compat 4.13.0)
 
-* `logrange(lo, hi; length)` is like `range` but for `*` not `+`. ([#39071]) (since Compat 4.14.0)
+* `logrange(lo, hi; length)` is like `range` but with a constant ratio, not difference. ([#39071]) (since Compat 4.14.0) Note that on Julia 1.8 and earlier, the version from Compat will has slightly lower floating-point accuracy than the one in Base (Julia 1.11 and later).
```
martinholters, line 75:
```suggestion
* `logrange(lo, hi; length)` is like `range` but with a constant ratio, not difference. ([#39071]) (since Compat 4.14.0) Note that on Julia 1.8 and earlier, the version from Compat has slightly lower floating-point accuracy than the one in Base (Julia 1.11 and later).
```

martinholters:
If you include a version bump in Project.toml, we can tag a new version immediately after merge. Otherwise seems good to go.

PR #823, opened 2024-02-21
Update TagBot.yml

ViralBShah:


ViralBShah:
I updated the Documenter keys as well. Can you try again?

martinholters:
I've manually triggered TagBot to create the missing releases and fixed up their descriptions to mention the correct pull requests.

PR #827, opened 2024-05-04
Explain the compat workflow for functions

gdalle:
It may be silly but I spent quite some time trying to make `@compat stack([x, y])` work whereas all I needed was `Compat.stack([x, y])`

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ and then as needed add
 @compat ...compat syntax...
 ```
 
+or 
+
+```julia
+Compat.function_available_in_later_julia_version(args..., kwargs...)
+```
+
 wherever you want to use syntax that differs in the latest Julia
 `master` (the development version of Julia). The `compat syntax` is usually
 the syntax on Julia `master`. However, in a few cases where this is not possible,
--- a/README.md
+++ b/README.md
@@ -54,17 +54,13 @@ and then as needed add
 @compat ...compat syntax...
 ```
 
-or 
-
-```julia
-Compat.function_available_in_later_julia_version(args..., kwargs...)
-```
-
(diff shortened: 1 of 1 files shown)


martinholters:
I'm a bit confused because after `using Compat`, all of the following should work:
```julia
stack([x, y])
Compat.stack([x, y])
@compat stack([x, y])
```
That said, it may make sense to clarify that `@compat` is mostly for _syntax_ changes and new functions/methods are just made available by `Compat`. However, your addition is inserted right before "wherever you want to use syntax that differs", but it's actually not about syntax. Maybe I'm too picky? Would appreciate others to chime in.

gdalle:
> I'm a bit confused because after using Compat, all of the following should work:

Maybe but the only thing that is mentioned on the README is the _macro_ `@compat`, and since I qualify all my exports, I only did

```julia
using Compat: @compat
```

> That said, it may make sense to clarify that @compat is mostly for syntax changes and new functions/methods are just made available by Compat.

Exactly, that's the part I had not understood.

> However, your addition is inserted right before "wherever you want to use syntax that differs", but it's actually not about syntax.

I'll modify the phrasing

martinholters (approved):
Ok with me, but I'd really appreciate others' opinions.
(And sorry for the delay.)

gdalle:
No worries about the delay. As for the PR, I always welcome more opinions but I feel like it is a relatively low-stakes addition to the docs, so perhaps they are not necessary for merging?

martinholters:
Sorry this has fallen through the cracks. No reason not to merge - even if the current phrasing might not be perfect, this PR certainly does not make things any worse.

PR #829, opened 2024-08-03
Add `Fix{N}` for fixing a single positional argument at any position

### LINKED ISSUE #828: Add `Fix{N}` utility for fixing an argument
I'm eager to use the `Fix{N}` syntax from https://github.com/JuliaLang/julia/pull/54653 in older Julia versions. It should be backwards compatible except for the multiple dispatch component.

MilesCranmer:
Adds https://github.com/JuliaLang/julia/pull/54653.

Fixes #828.

```julia
julia> using Compat: Fix

julia> map(Fix{3}(muladd, 1.0), 1:3, 4:6)
3-element Vector{Float64}:
  5.0
 11.0
 19.0
```

---

**Multiple dispatch**: This is a bit of an edge case, and presumably many functions in Compat are like this, but since `Fix1` is broadly used:

Note that even with this PR, `Compat.Fix{1}` is not the same as `Base.Fix1` on Julia versions earlier than `1.12.0-DEV.966`. Therefore, if you wish to use this as a way to dispatch on `Fix{N}` for libraries compatible with Julia earlier than v1.12, you should declare a function both for `Compat.Fix{1}` as well as `Base.Fix1`. You can do this with a

```julia
@static if !isdefined(Base, :Fix)
    #= Declare additional methods for Compat.Fix =#
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ changes in `julia`.
 
 ## Supported features
 
+* `Fix{N}` which fixes an argument at the `N`th position ([#54653]) (since Compat 4.16.0)
+
 * `chopprefix(s, prefix)` and `chopsuffix(s, suffix)` ([#40995]) (since Compat 4.15.0)
 
 * `logrange(lo, hi; length)` is like `range` but with a constant ratio, not difference. ([#39071]) (since Compat 4.14.0) Note that on Julia 1.8 and earlier, the version from Compat has slightly lower floating-point accuracy than the one in Base (Julia 1.11 and later).
@@ -192,3 +194,4 @@ Note that you should specify the correct minimum version for `Compat` in the
 [#47679]: https://github.com/JuliaLang/julia/pull/47679
 [#48038]: https://github.com/JuliaLang/julia/issues/48038
 [#50105]: https://github.com/JuliaLang/julia/issues/50105
+[#54653]: https://github.com/JuliaLang/julia/issues/54653
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ changes in `julia`.
 
 ## Supported features
 
-* `Fix{N}` which fixes an argument at the `N`th position ([#54653]) (since Compat 4.16.0)
+* `Compat.Fix{N}` which fixes an argument at the `N`th position ([#54653]) (since Compat 4.16.0)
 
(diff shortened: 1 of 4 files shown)


mcabbott (approved):
LGTM, but I don't have merge rights.

MilesCranmer:
@martinholters would you be willing to take a look

src/Compat.jl
```diff
@@ -1122,6 +1122,67 @@ if VERSION < v"1.8.0-DEV.1016"
     export chopprefix, chopsuffix
 end
 
+# https://github.com/JuliaLang/julia/pull/54653: add Fix
+@static if !isdefined(Base, :Fix)
```
martinholters, line 1126:
```suggestion
@static if !isdefined(Base, :Fix) # VERSION < v"1.12.0-DEV.981"
```
(Can leave the condition as is, but having the Julia version noted comes handy once we set about weeding out old stuff.)

martinholters (commented):
>Note that even with this PR, `Compat.Fix{1}` is not the same as `Base.Fix1` on Julia versions earlier than `1.12.0-DEV.966`. Therefore, if you wish to use this as a way to dispatch on `Fix{N}` for libraries compatible with Julia earlier than v1.12, you should declare a function both for `Compat.Fix{1}` as well as `Base.Fix1`.

Should this be documented somewhere more findable than the PR description?

Apart from that LGTM.

README.md
```diff
@@ -72,6 +72,8 @@ changes in `julia`.
 
 ## Supported features
 
+* `Fix{N}` which fixes an argument at the `N`th position ([#54653]) (since Compat 4.16.0)
```
martinholters, line 75:
```suggestion
* `Compat.Fix{N}` which fixes an argument at the `N`th position ([#54653]) (since Compat 4.16.0)
```

MilesCranmer:
> Should this be documented somewhere more findable than the PR description?

I’ll add it to the docs for the function

MilesCranmer:
Done.

martinholters (approved):
LGTM. Barring objections, I'll merge this and #830 tomorrow and tag a new version.

MilesCranmer:
Great, thanks!

PR #835, opened 2025-03-18
Add `@gcsafe ccall`

kpamnany:
From [GPUToolbox.jl](https://github.com/JuliaGPU/GPUToolbox.jl/blob/4cdaedb0f0650e20008c5c767e73cfe1ecb59cda/src/ccalls.jl#L101).

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1221,6 +1221,77 @@ else
     using Base: Fix, Fix1, Fix2
 end
 
+## version of ccall that calls jl_gc_safe_enter|leave around the inner ccall
+# note that this is generally only safe with functions that do not call back into Julia.
+# when callbacks occur, the code should ensure the GC is not running by wrapping the code
+# in the `@gcunsafe` macro
+
+const HAS_CCALL_GCSAFE = VERSION >= v"1.13.0-DEV.70" || v"1.12-DEV.2029" <= VERSION < v"1.13-"
+
+"""
+    @gcsafe_ccall ...
+
+Call a foreign function just like [`@ccall`](https://docs.julialang.org/en/v1/base/c/#Base.@ccall), but marking it safe for the GC to run. This is
+useful for functions that may block, so that the GC isn't blocked from running, but may also
+be required to prevent deadlocks (see JuliaGPU/CUDA.jl#2261).
+
+Note that this is generally only safe with non-Julia C functions that do not call back into
+the Julia directly.
+"""
+macro gcsafe_ccall end
+export @gcsafe_ccall
+
+if HAS_CCALL_GCSAFE
+    macro gcsafe_ccall(expr)
+        exprs = Any[:(gc_safe = true), expr]
+        return Base.ccall_macro_lower((:ccall), Base.ccall_macro_parse(exprs)...)
+    end
+else
+    function ccall_macro_lower(func, rettype, types, args, nreq)
+        # instead of re-using ccall or Expr(:foreigncall) to perform argument conversion,
+        # we need to do so ourselves in order to insert a jl_gc_safe_enter|leave
+        # just around the inner ccall
+
+        cconvert_exprs = []
+        cconvert_args = []
+        for (typ, arg) in zip(types, args)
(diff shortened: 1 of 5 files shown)


kpamnany:
@martinholters: looks like an upstream change has broken a test you added? Can you take a look please?

martinholters:
Huh? AFAICT, the problem originates in https://github.com/JuliaLang/Compat.jl/blob/8819abe252579b610cb48c5ad83d51b45a90ddba/test/runtests.jl#L158-L174 which was added in #746 by @c42f and then modified by @Sacha0 in #758. But I can take a look non-the-less, of course. Presumable the error caused by accessing `__not_a_binding__` has changed by the binding partitioning work? Maybe explicitly throwing an exception would be less brittle?

kpamnany:
So sorry for the mistaken tag @martinholters! I didn't look too closely... maybe @Sacha0 can take a look at it?

martinholters:
No worries. And it's just the "Suggestion: check for spelling errors or missing imports." that has been added to the exception printing at some point that causes the test failure. Fix at #836.

test/Project.toml
```diff
@@ -0,0 +1,5 @@
+[deps]
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
martinholters, line 5: I think we should rather add `InteractiveUtils` to the `test` target in to root `Project.toml`. Or remove that target configuration there and fully switch to a separate `test/Project.toml`, but that might best be done in its own PR. (And I'm not aware of the proc/cons of both approaches.)

martinholters:
Any chance to make this work on v1.6 and v1.7?

test/Project.toml
kpamnany: I don't know what's best either. Will wait for input from other reviewers.

kpamnany:
> Any chance to make this work on v1.6 and v1.7?

@vchuravy would know... when he gets back and reviews this.

test/Project.toml
kpamnany: Did as you suggested.

src/Compat.jl
```diff
@@ -1221,6 +1221,81 @@ else
     using Base: Fix, Fix1, Fix2
 end
 
+## version of ccall that calls jl_gc_safe_enter|leave around the inner ccall
+# note that this is generally only safe with functions that do not call back into Julia.
+# when callbacks occur, the code should ensure the GC is not running by wrapping the code
+# in the `@gcunsafe` macro
+
+if VERSION >= v"1.8"
```
vchuravy, line 1229: I don't see why this code should be lower bounded to 1.8?

kpamnany:
```
gcsafe_ccall: Error During Test at /Users/kpamnany/Work/Compat.jl/test/runtests.jl:1086
  Test threw exception
  Expression: gc_safe_ccall() isa UInt64
  could not load symbol "jl_rand":
  dlsym(RTLD_DEFAULT, jl_rand): symbol not found
<snip>
```

kpamnany: Calling any Julia C function marked `JL_NOTSAFEPOINT` will serve, yeah?

kpamnany: Switched to calling `jl_errno`.

KristofferC:
I got asked to review here so I give my personal opinion. I don't think Compat with a bunch of internals that tries to backport features from later Julia versions is a good idea in general. I don't think adding a Compat dependency (which transitively gives you dependencies on very large stdlibs) is very nice, especially when it is done on small packages that themselves have many other packages depending on them. I personally rather see some copy-pasting if something like this is really needed.

kpamnany:
@KristofferC's points make sense to me. I can copy-paste this macro instead; that was also an alternative @vchuravy had suggested.

PR #837, opened 2025-03-20
Require Julia 1.10 and deprecate type-pirating methods

### LINKED ISSUE #814: Avoid type piracy by defining distinct functions

martinholters:
Compat dropped support for Julia versions prior to v1.6 about 14 months after it was released, 6 months after it became LTS (by the release of 1.7). As v1.10 has been released about 13 months ago and became LTS about 5 months ago, we can start thinking about dropping support for versions prior to v1.10. Which this PR does.

Seizing the opportunity of a major version bump, this does another change that might be worth more discussion: It resolves #814 by applying it to the few remaining offenders. Maybe this PR - threatening to actually have an impact - receives more feedback than #814... If there are objections, I happily remove the last two commits to reduce this to the minimum-required-version-bump only.


lgoettgens:
I am really in favor of getting rid of the type piracy like your are doing here.

One minor comment: In the feature overview at https://github.com/JuliaLang/Compat.jl/tree/mh/deprecate-type-piracy?tab=readme-ov-file#supported-features, it looks a bit weird that `insertdims` and `logrange` don't need a `Compat.` qualification. Would it make sense to make sense to try to unify how to use `Compat.jl`'s different features by not exporting these two (but instead making them `@compat public`)? No strong opinion either way, but I wanted to point it out anyway.

PR #841, opened 2025-07-20
add `@__FUNCTION__` macro

MilesCranmer:
https://github.com/JuliaLang/julia/pull/58940

The macro implemented here differs slightly because it needs to refer to the internal variable `var"#self#"`, rather than the new `Expr(:thisfunction)` directive, as the later required parser changes to expand compatibility with callable structs.

However, the only change this means in practice is that the Compat version will throw an error in cases where the Julia 1.13.0-DEV.880 version would work. So there would not be any silent differences; only very obvious ones. I have mentioned these cases on the Compat version docstring.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Compat"
 uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
-version = "4.17.0"
+version = "4.18.0"
 
 [deps]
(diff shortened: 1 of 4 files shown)


MilesCranmer:
@martinholters let me know if you have any comments. I'm eager to get this merged as I need it in a package

MilesCranmer:
@martinholters sorry for pinging but just wanted to check-in on this (usually I only have time to code on the weekend these days)

PR #842, opened 2025-09-24
test: `Fix`: adjust "structs not in `Base`" test

nsajko:
* Do not test the exact return type of `Fix{1}(MyStruct, 1)`. That does not seem necessary, given that the intention behind the test is merely ensuring that `Fix` works correctly for a newly-defined `MyStruct`.

* Test that `Fix{1}(MyStruct, 1)()` calls `MyStruct(1)`.

The motivation is making PkgEval pass for PR JuliaLang/julia#59623.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1066,10 +1066,6 @@ end
             construction_happened::Bool = false
             struct MyStruct
                 x::Int
-                function MyStruct(x::Int)
-                    construction_happened = true
-                    new(x)
-                end
             end
             f = Fix{1}(MyStruct, 1)
             @test f isa Fix{1}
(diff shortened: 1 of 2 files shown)


martinholters (commented):
I'm not convinced https://github.com/JuliaLang/julia/pull/59623 is the way to go, but this test shouldn't stand in the way; ok with me to weaken it.

martinholters:
You can also bump the version, so that we can do a release immediately after merging.

test/runtests.jl
```diff
@@ -1063,11 +1063,17 @@ end
             @test_throws ArgumentError("expected type parameter in `Fix` to be `Int`, but got `1::UInt64`") Fix{UInt64(1)}(>, 1)
         end
         @testset "Specialize to structs not in `Base`" begin
+            construction_happened::Bool = false
             struct MyStruct
                 x::Int
+                function MyStruct(x::Int)
+                    construction_happened = true
+                    new(x)
+                end
```
nsajko, line 1072:
```suggestion
```

martinholters (approved):

PR #843, opened 2025-09-29
Update CI workflow to use latest action versions

ViralBShah:


.github/workflows/CI.yml
```diff
@@ -18,33 +18,45 @@ jobs:
...
+            version: pre
+          - os: macOS-latest
+            arch: aarch64
+            version: nightly
+
     steps:
-      - uses: actions/checkout@v3
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v5
+      - uses: julia-actions/setup-julia@v2
```
martinholters, line 51: Shouldn't dependabot take care of these? 🤔
EDIT: dependabot was just enabled in #844, missed that, sorry.

PR #847, opened 2025-11-10
Add `chopprefix`/`chopsuffix` methods for `AbstractChar` from PR 59425

ararslan:
These were added in [\#59425](https://github.com/JuliaLang/julia/pull/59425), during the 1.13 development cycle.

### DIFF
--- a/src/Compat.jl
+++ b/src/Compat.jl
@@ -1132,6 +1132,10 @@ end
 
 # https://github.com/JuliaLang/julia/pull/59425
 if VERSION < v"1.13.0-DEV.1078"
+    if VERSION >= v"1.8.0-DEV.1016"
+        # Extend from Base without qualification (avoids duplicating the definitions)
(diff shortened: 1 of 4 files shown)


src/Compat.jl
```diff
@@ -1130,6 +1130,24 @@ if VERSION < v"1.8.0-DEV.1016"
     export chopprefix, chopsuffix
 end
 
+# https://github.com/JuliaLang/julia/pull/59425
+if VERSION < v"1.13.0-DEV.1078"
```
ararslan, line 1134: Ah, probably need this
```suggestion
if VERSION < v"1.13.0-DEV.1078"
    if VERSION >= v"1.8.0-DEV.1016"
        # Extend from Base without qualification (avoids duplicating the definitions)
        import Base: chopprefix, chopsuffix
    end
```

